PR #20, opened 2026-06-28
ci: change arm runner name

matteoacrossi:
This pull request updates the runner configuration for ARM-based jobs in the GitHub Actions workflows. The main change is replacing the generic `aarch64` runner label with the more specific `ubuntu-24.04-arm` label for both deployment and testing workflows. This ensures that ARM jobs run on the correct runners for public repos.

See https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories.


robertodr (approved):
Nice

PR #21, opened 2026-06-28
ci(workflows): :gear: split SonarQube analysis from cpp checks

matteoacrossi:
## Summary
- decouple cpp-checks from python-checks by removing SonarQube/coverage aggregation from the C++ job
- add a dedicated sonarqube-analysis job that depends on both python-checks and cpp-checks
- upload C++ Sonar inputs from cpp-checks and consume them in sonarqube-analysis
- move stable Sonar settings into sonar-project.properties and keep runtime compile-commands path in workflow args
- make publish-test-results depend on sonarqube-analysis

## Notes
Runtime-generated paths (like build-wrapper compile commands) remain in workflow arguments because they are produced during CI execution.


robertodr (dismissed):
Looks legit

robertodr (approved):

PR #25, opened 2026-06-29
refactor: fix the interface and tests

ludmilaasb:


SimoneGasperini (commented):
Thank you @ludmilaasb! Looks good to me but I would wait the final approval by @diagonal-hamiltonian before merging 🙏🏽

src/monoprop/monomial_propagator.py
```diff
@@ -346,46 +345,58 @@ def schrodinger(self) -> bool:
...
         self,
-        majoranas: list[tuple[int, ...]] | None = None,
+        quantum_circuit: IQuantumCircuit | MonomialCircuit,
         operator_coeffs: None | list[float] | np.ndarray = None,
-        parameter_mapping: list[int] | np.ndarray | None = None,
-        gen_coeffs: list[float] | np.ndarray | None = None,
-        parameters: list[float] | np.ndarray | None = None,
         *,
-        evolve_with_coeffs: bool = False,
+        ignore_circuit_parameters: bool = True,
```
adamglos92, line 379: This is more direct name, especially since it is connected to contract_partially
```suggestion
        contract_graph: bool = True,
```

src/monoprop/monomial_propagator.py
```diff
@@ -346,46 +345,58 @@ def schrodinger(self) -> bool:
...
-           and ``operator_coeffs``. Operator coefficients can be obtained from a
+        1. Build only the propagation graph by ignoring circuit parameters.
+        2. Build the propagation graph with coefficient information by using the circuit
+           parameters and ``operator_coeffs``. Operator coefficients can be obtained from a
            prior call to ``contract_partially(inplace=False)`` if you want to
            preserve the graph.
         3. Propagate and contract immediately without building a graph by providing
-           ``majoranas``, ``parameter_mapping``, ``gen_coeffs``, and
-           ``parameters`` only (do not provide ``operator_coeffs``). This mode is
+           the only the circuit (do not provide ``operator_coeffs``). This mode is
```
adamglos92, line 392: For this mode it is required that the graph is "empty" (everything that could be contracted was already contracted) 
```suggestion
           only the circuit. Assumes MonomialPropagator is parameter-free (either no propagation was invoked, or the graph was already contracted). This mode is
```

src/monoprop/monomial_propagator.py
```diff
@@ -396,14 +407,20 @@ def propagate(
...
+            if isinstance(quantum_circuit, MonomialCircuit)
+            else quantum_circuit.get_monomial_circuit()
         )
+        majoranas = quantum_circuit.majoranas
 
-        if evolve_with_coeffs:
-            parameter_mapping = self.quantum_circuit.param_inds
-            gen_coeffs = self.quantum_circuit.gen_coeffs
-            parameters = self.quantum_circuit.parameters
+        if ignore_circuit_parameters or operator_coeffs:
```
adamglos92, line 417: `ignore_circuit_parameters` is too flexible in here. Whether you need to store coeffciients depends on the following:
* if atol are non-trivial and graph is not contracted - store
* if atol are non-trivial and graph is contracted - don't store (as they will be immediately consumed)
* if atol are trivial (None) - no need to store, but no harm
Actually because we can change the atol, perhaps we should always store unless contracted

src/monoprop/monomial_propagator.py
```diff
@@ -416,10 +433,8 @@ def propagate(
 
     def expectation_value_functional(
         self,
-        parameter_mapping: list[int] | np.ndarray | None = None,
-        gen_coeffs: list[float] | np.ndarray | None = None,
         *,
-        use_coeffs: bool = False,
+        ignore_coeffs: bool = False,
```
adamglos92, line 437: I'm confused what this argument is really doing. For me it only makes sense when it is False

adamglos92: same in all the other functions

src/monoprop/monomial_propagator.py
```diff
@@ -546,8 +547,9 @@ def gradient_functional(
 
     def expectation_value(
         self,
+        parameters: list[float] | np.ndarray | None = None,
```
adamglos92, line 550: Being honest, I don't know what parameters = None would mean in here. Perhaps we should just expect explicit empty list?

src/monoprop/monomial_propagator.py
```diff
@@ -699,10 +681,12 @@ def contract_partially(
                 if parameter length doesn't match mapping requirements, or if mapping
                 length exceeds the number of evolved Majoranas.
         """
+        parameter_mapping = None
+        gen_coeffs = None
         if not ignore_coeffs:
```
adamglos92, line 686: I have no idea what is the meaning of ignore_coeffs. @diagonal-hamiltonian? @ludmilaasb?

src/monoprop/qiskit_conversion.py
```diff
@@ -87,7 +87,6 @@ def to_qiskit_operator(
 
 def from_qiskit_circuit(
```
adamglos92, line 88: Question: initial X gates are not welcome?

src/monoprop/qiskit_conversion.py
adamglos92, line 93: "each being a linear combination of commuting Pauli strings" (currently might be interpreted as if the gates should commute)

adamglos92 (commented):
Looks largely OK, a few comments

src/monoprop/qiskit_conversion.py
ludmilaasb: No, this is converting just the "ansatz" part. And this is out of the scope of this PR.

src/monoprop/monomial_propagator.py
ludmilaasb: Parameters = None is understand as an empty list.

src/monoprop/monomial_propagator.py
ludmilaasb: Ignore coeffs = empty lists

diagonal-hamiltonian:
Maybe I don't see it but how is the initial state extracted from the circuit

diagonal-hamiltonian:
I think the initial state should still be part of the circuit, or else then it is not really a circuit. What do you guys think? @adamglos92 @ludmilaasb

src/monoprop/monomial_propagator.py
adamglos92: I don't know what would it mean. @diagonal-hamiltonian does it make sense to you?

src/monoprop/qiskit_conversion.py
adamglos92: OK, just wanted to understand it

adamglos92:
> I think the initial state should still be part of the circuit, or else then it is not really a circuit. What do you guys think? @adamglos92 @ludmilaasb

I agree, requesting from the user to extract it every time sounds a little cumbersome

src/monoprop/monomial_propagator.py
diagonal-hamiltonian: t computes the energy with no graph contraction

src/monoprop/monomial_propagator.py
ludmilaasb: So the only thing that matters is if the graph is contracted, right? And what determinate if the graph is contracted is operator_coeff been passed or if you want to use the circuit parameters.

src/monoprop/monomial_propagator.py
ludmilaasb: What this argument does is to ignore the circuit coefficients values. This basically is equivalent of not passing gen_coeffs, param_inds, parameters... etc as you guys were doing before.

ludmilaasb:
> > I think the initial state should still be part of the circuit, or else then it is not really a circuit. What do you guys think? @adamglos92 @ludmilaasb
> 
> 
> 
> I agree, requesting from the user to extract it every time sounds a little cumbersome

What do you mean the user extract all the time? The initial state is only passed on the instantiation of MP.

ludmilaasb:
Besides, removing initial state from the circuit was your suggestion, @adamglos92

src/monoprop/qiskit_conversion.py
ludmilaasb: If I remember correctly, you suggested separating the initial state from the rest of the circuit. If you want to have the initial state accepted back into the Qiskit circuit, please open an issue. Before we were parsing it.

src/monoprop/monomial_propagator.py
adamglos92: operator_coeff could be made optional and computed internally with the stored parameters. But if we don;t want to do that providing operator_coeff (or not) might identify the desire enough, although it is quite confusing + not needed if the mbs is empty. Having such keyword is far more explicit and clear

src/monoprop/monomial_propagator.py
adamglos92: and I don't know what would mean they were not provided. The MBS graph was parameter-free?
question to @diagonal-hamiltonian  I suppose

src/monoprop/monomial_propagator.py
adamglos92: I don't remember what was the function before. but from what I remember the function itself was never doing contraction.

adamglos92:
> Besides, removing initial state from the circuit was your suggestion, @adamglos92

to be clear: I never said it cannot be grouped under the same object. But it should be easily accessible to user because it is provided in different moments than the circuit. + it was more argument about the output of conversion functions, not input

PR #27, opened 2026-06-29
refactor!: 🔥 remove MPData in favor of minimal test-data loaders

matteoacrossi:
## Summary

Removes the `MPData` data class (Python **and** C++) and replaces it with minimal, test-only loaders that read fixtures directly into the real API. `MPData` existed only to (de)serialize msgpack test fixtures; it was exported from the public `monoprop` package and shipped test-loading code (plus an `msgpack-cxx` dependency) inside the core library, despite having no non-test users.

> [!WARNING]
> **Breaking change:** `MPData` is no longer exported from the `monoprop` package.

## What changed

**Fixtures** (`tests/data/*.msgpack`)
- Repacked into a flat, minimal msgpack schema keeping only the 9 fields the suite consumes (drops `ansatz_*`, `evolved_hamiltonian`, `tag`, version metadata, `fermionic_pool`, `commutator_gradient`).
- New `tests/data/README.md` documents the schema (previously undocumented).
- Files shrink substantially, e.g. `random_exact` 214 KB → 930 B, `lih_fermionic_spin_exact` 73 KB → 16 KB.

**Python**
- Delete `src/monoprop/mp_data.py` and the `MPData` export from `__init__.py`.
- Delete `tests/test_mp_data.py` (only tested `MPData` itself).
- `tests/cases.py`: add `load_problem()` building `MonomialCircuit` + `MonomialOperator` directly; drop the dead `MPFermData` NamedTuple and unused `FermionicProblem` fields.
- `tests/test_monoprop_mpi.py`: use `load_problem`; `_make_mp` reuses the loaded `operator`/`monomial_circuit`.

**C++**
- Delete `src/monoprop/DataGenerators.{h,cpp}` — removes test-only loading code and the `msgpack-cxx` dependency from the shipped library.
- Add test-only `tests/cpp/TestData.{h,cpp}` with a minimal `CaseData` struct + `load_case()`.
- Move the `msgpack-cxx` CPM declaration to the top-level `CMakeLists.txt` (so the imported target is visible in the `tests/cpp` scope) and link it into `monoprop_unit_tests.x`.

## Verification

- C++ unit suite: **31/31 ctest tests pass** (`-L unit`) on the repacked fixtures.
- Python: **383 passed** (non-MPI); `test_monoprop_mpi.py` **7 passed, 1 skipped** (2-rank-only) under `--with-mpi` single-rank.
- `ruff check`, `ruff format`, `clang-format` clean; no remaining `MPData`/`DataGenerators` references.

## Notes for reviewers

- The full multi-rank MPI matrix (`just test-py-mpi` at 2+ ranks) requires an MPI-enabled rebuild and was **not** run locally — worth confirming in CI.
- The one-off repack migration script was run locally and is **not** committed (it reads the now-replaced old format). Happy to add it under `tools/` for provenance if preferred.


tests/cpp/TestData.h
```diff
@@ -0,0 +1,46 @@
...
+///
+/// Mirrors the flat msgpack schema documented in `tests/data/README.md`; only
+/// the fields exercised by the C++ test suite are kept.
+struct CaseData {
+    double actual_expval{0.0};
+    monoprop::VecZ hartree_fock;
+    monoprop::VecZ param_inds;
+    monoprop::VecD gen_coeffs;
+    monoprop::VecD parameters;
+    std::vector<monoprop::VecZ> majoranas;
```
diagonal-hamiltonian, line 34: Do you think this could be a C++ version of the python MonomialCircuit object and we can unify the inputs?

diagonal-hamiltonian (commented):
Looks great cheers. If we want to just use this for the tests thats okay, but we could unify the MonomialCircuit python and C++ code, but this is good for now.

tests/cpp/TestData.h
matteoacrossi: Definitely. Is this something that is cooking in #25?

matteoacrossi:
Yeah my plan was to quickly get rid of the old tests and `MPData`. Should we do this in a new PR?

tests/cpp/TestData.h
diagonal-hamiltonian: No this is just changing the python interface which is currently confusing

diagonal-hamiltonian:
yeah new PR. This is good

CMakeLists.txt
```diff
@@ -123,6 +123,17 @@ if(NOT DEFINED SKBUILD)
...
+  # msgpack data fixtures. Declared here (rather than in src/) so the imported
+  # target is visible in the tests/cpp directory scope.
+  cpmaddpackage(
+    NAME "msgpack-cxx"
+    VERSION "7.0.0"
+    GIT_REPOSITORY "https://github.com/msgpack/msgpack-c"
+    GIT_TAG "cpp-7.0.0"
+    OPTIONS "MSGPACK_USE_BOOST OFF" "MSGPACK_CXX20 ON"
+  )
+
```
ludmilaasb: Maybe this could be directly in the CMakeList of the test folder?

ludmilaasb (dismissed):
LGTM!

CMakeLists.txt
robertodr: Agreed.

robertodr (approved):

robertodr:
Thank you @matteoacrossi!

PR #32, opened 2026-06-29
feat: memory and time benchmarks with pytest

diagonal-hamiltonian:
## Summary

Adds a pytest-based benchmark suite under `benches/` measuring the **wall-clock time** and **peak physical memory (PSS)** of monoprop's core operations in a single sweep, across both the Heisenberg and Schrödinger pictures, and MPI- and thread-aware. Each run regenerates `benches/results/REPORT.md` with one column per label, so serial, multi-rank, and multi-thread variants sit side by side. Not part of the normal `pytest` run — invoked with `just bench`.

## What's included

- **Random benchmarks** (`bench_random.py`; configurable, both pictures) — `build_graph`, `pare`, `energy`, `gradient` (graph-based) and `inplace`.
- **Fixed-model benchmarks** (`bench_models.py`; heavy, Heisenberg-only, in-place) — a 120-qubit Fermi-Hubbard Trotter trajectory and a 127-qubit Pauli-basis kicked-Ising circuit (IBM Eagle heavy-hex). Every config field is overridable via `--<model>-<field>`.
- **MPI-aware** — operations are barrier-wrapped so the timed cost is the makespan across ranks; only rank 0 writes/prints. `_mpi_check.py` aborts early if a non-MPI build is loaded (which would replicate the full operator per rank and OOM).
- **Report** (`report.py`) — merges `<label>.json` (metadata, sizes, configs) and `time-<label>.json` (pytest-benchmark) into `REPORT.md`.

## Memory profiling

A background sampler reads **PSS** (`/proc/self/smaps_rollup`) while each operation runs. This captures the C++/TBB allocations `tracemalloc` misses at no measurable cost (compute releases the GIL, so timing and memory share one sweep), and sums correctly across ranks — PSS bills shared pages once, so the per-rank sum is the job's true physical RAM. Under MPI the per-rank timelines merge into the **peak-of-sum** (the largest summed footprint that actually coexisted), not the sum of per-rank peaks. Linux-only; degrades to a no-op elsewhere.

## How to run

```bash
just bench serial      # serial run -> benches/results/REPORT.md
just bench-smoke       # quick sanity check (tiny sizes)

just bench-build-mpi   # build MPI once; bench recipes use --no-sync so it survives
monoprop_NUM_THREADS=2 just bench-mpi r5t2 5 --map-by slot:PE=2 --bind-to core
```

## Testing

- `tests/test_bench_builders.py`, `tests/test_bench_memory.py`, `tests/test_bench_report.py` cover the model builders, PSS/peak-of-sum logic, and report builder (part of the normal suite).
- `just bench-smoke` runs the full sweep end-to-end; `ruff` and all pre-commit hooks pass.

---
_Re-created from #24, which GitHub auto-closed after `feat/python-benches` was rebased/force-pushed. Same branch, commits replayed._


pyproject.toml
```diff
@@ -77,6 +77,10 @@ test = [
   "pytest-datadir",
   "pytest-mpi>=0.6",
 ]
+bench = [
+  { include-group = "test" },
+  "pytest-benchmark>=4",
```
robertodr: let's go for latest and greatest
```suggestion
  "pytest-benchmark>=5",
```

pyproject.toml
```diff
@@ -157,11 +161,15 @@ addopts = [
...
-pythonpath = ["."]
+# Collect both the test suite (test_*.py) and the benchmark suite (bench_*.py);
+# benches/ is excluded from testpaths so it only runs when invoked explicitly.
+python_files = ["test_*.py", "bench_*.py"]
+pythonpath = [".", "benches"]
 markers = [
   "slow: mark test as slow (>=1 minute runtime)",
   "mpi: mark test as requiring --with-mpi (pytest-mpi)",
   "qiskit: mark test as requiring qiskit (optional dependency)",
+  "bench: mark as a benchmark (collected only from the benches/ suite)",
```
robertodr: `pytest-benchmark` has its own marker: https://pytest-benchmark.readthedocs.io/en/latest/usage.html#markers isn't it good enough?

benches/_builders.py
```diff
@@ -0,0 +1,535 @@
...
+
+
+def make_random_problem(
+    *,
+    gen_length: int = 4,
+    obs_terms: int = 10000,
+    num_generators: int = 100,
+    num_modes: int = 128,
+    cutoff: int = 6,
+    seed: int = 0,
```
robertodr: I think we can leave it up to the user to fix the seed
```suggestion
    seed: int | np.random.Generator | None = None,
```

benches/_mpi_check.py
```diff
@@ -0,0 +1,124 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Preflight check that the MPI bench actually distributes work across ranks.
+
+monoprop splits operator terms across ranks by hashing each term to a rank
```
robertodr: @diagonal-hamiltonian I really don't understand what you're trying to do... If you want to run with MPI and the extension was not built with MPI, we should have a much nicer, simpler way to discover that, _e.g._ with a module-level attribute `mp.has_mpi` or something.

benches/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+| Knob | How | Example |
+|---|---|---|
+| MPI ranks | `--ranks N` | `--ranks 4` |
+| mpiexec args (incl. pinning) | `--mpiexec-args` | `--mpiexec-args="--bind-to core"` |
+| threads & pinning | `--env KEY=VAL` (repeatable) | `--env monoprop_NUM_THREADS=2` |
+| run name / report column | `--label` | `--label r4t2` |
+
+`monoprop` reads `monoprop_NUM_THREADS` at import; add `OMP_NUM_THREADS` /
+`OMP_PROC_BIND` / `OMP_PLACES` for OpenMP pinning. Use a custom `--label` when
```
robertodr: those env-vars won't be used though.
```suggestion
`monoprop` reads `monoprop_NUM_THREADS` at import. Use a custom `--label` when
```

benches/run.py
```diff
@@ -0,0 +1,240 @@
...
+# Environment variables recorded in each run's metadata (threading + pinning).
+TRACKED_ENV = (
+    "monoprop_NUM_THREADS",
+    "OMP_NUM_THREADS",
+    "OMP_PROC_BIND",
+    "OMP_PLACES",
+    "OPENBLAS_NUM_THREADS",
+    "MKL_NUM_THREADS",
+    "NUMEXPR_NUM_THREADS",
+    "TBB_NUM_THREADS",
```
robertodr: I don't think we need to track those
```suggestion
    "TBB_NUM_THREADS",
```

src/monoprop/bindings/binder.h
```diff
@@ -213,5 +213,14 @@ auto bind_monomial_propagator(nb::module_ &mod) -> void {
     cls.def("graph_data", &MonomialPropagator<NumModes>::graph_data);
 
     cls.def("graph_layers", &MonomialPropagator<NumModes>::graph_layers);
+
+    // Read-only structural memory accounting (capacity-based byte totals), so the
+    // benchmark report can attribute footprint to the operator vs the graph.
```
robertodr: these slop comments always feel slightly out of place 🤷 
```suggestion
```

benches/results/.gitignore
robertodr, line 1: I suggest to put a README.md (that we track) in this folder, with this information:
```
This directory holds locally generated benchmark artifacts (REPORT.md, time-*.json, meta-*.json, params-*.json, opsize-*.json, configs-*.json, mem-*.json). See benches/README.md for more information.

These files should **NOT** be added to the repo.
```

and then ignore everything in this folder (except `README.md`) in the top-level `.gitignore`. Basically what we do for the `sandbox` folder.

pyproject.toml
```diff
@@ -300,6 +307,11 @@ flake8-annotations.allow-star-arg-any = true
 "tests/**/test_*.py" = ["ANN", "I002"]
 # ignore I002 (isort required import) for all `__init__.py` files
 "src/monoprop/**/__init__.py" = ["I002"]
+# benchmark suite: asserts are fine (S101), it is an intentional non-package
+# directory loaded via pythonpath (INP001), the MPI communicator type is
+# unavoidably dynamic (ANN401), and the run.py driver intentionally prints
+# progress (T201) and shells out to pytest/mpiexec (S603/S607).
+"benches/*" = ["S101", "INP001", "ANN401", "T201", "S603", "S607"]
```
robertodr: you can deactivate (and remove) type annotations for all the `bench_*.py` files. they only add visual noise.

benches/bench_monoprop.py
robertodr, line 1: can you split random and static into separate files? Can you come up with a better name than "static" 😅 ?

benches/bench_monoprop.py
```diff
@@ -0,0 +1,216 @@
...
+)
+
+if TYPE_CHECKING:
+    from _builders import RandomProblem
+
+    from monoprop import MonomialPropagator
+
+PARE_THRESHOLD = 1e-10
+# In-place random bench keeps its own truncation constant (see design non-goals).
+INPLACE_LOWER_ATOL = 1e-5
```
robertodr: wrap in fixtures plz

benches/conftest.py
robertodr, line 1: the memory tracking is Linux-only. Are we fine with that? I don't have better ideas, just so you know. `pytest-memray` might do it, but maybe not in a format amenable to performance tracking.

benches/_builders.py
diagonal-hamiltonian: this is handy for benchmarking consistently myself

benches/_mpi_check.py
diagonal-hamiltonian: thats actually not a bad shout

benches/conftest.py
matteoacrossi: Fine with Linux only.

diagonal-hamiltonian: i tried memray first but it regreesed timings

pyproject.toml
diagonal-hamiltonian: Removed the custom `bench` marker and the `@pytest.mark.bench` decorators — it was never used for selection (testpaths scopes normal `pytest` runs to `tests/`, and the bench recipes target `benches/` directly).

benches/_mpi_check.py
diagonal-hamiltonian: Done — `_mpi_check.py` now just imports monoprop and checks `monoprop.has_mpi`; `just bench-mpi` runs it as a preflight before the heavy passes.

benches/README.md
diagonal-hamiltonian: Reworded — the README no longer documents the `--env` passthrough at all; the thread count is set via the `monoprop_NUM_THREADS` env var directly (`monoprop_NUM_THREADS=10 just bench serial-t10`).

benches/run.py
diagonal-hamiltonian: `run.py` is gone — launching moved to `just bench` / `just bench-mpi` recipes, so there is no env-tracking list anymore.

src/monoprop/bindings/binder.h
diagonal-hamiltonian: Removed in the `deslop` commit (6007b6e).

benches/results/.gitignore
diagonal-hamiltonian: Done — replaced `benches/results/.gitignore` with a tracked `README.md` (with that note), and added `benches/results/**` + `!benches/results/README.md` to the top-level `.gitignore`.

pyproject.toml
diagonal-hamiltonian: Done — stripped the annotations from the bench test functions and added `"benches/bench_*.py" = ["ANN"]` to per-file-ignores.

benches/bench_monoprop.py
diagonal-hamiltonian: Split into `bench_random.py` and `bench_models.py`. Went with "model" for the fixed real-model suite: `MODELS` registry, `test_model[hubbard|pauli]`, `model_configs`.

benches/bench_monoprop.py
diagonal-hamiltonian: Done — added a `make_random_propagator` factory fixture; the build-graph/inplace `setup` callbacks call it instead of rebuilding inline.

benches/conftest.py
```diff
@@ -0,0 +1,415 @@
...
+
+    Under MPI the per-rank PSS values are summed via a collective reduce (every rank
+    runs every test), giving the job's true physical RAM. Written env-gated and
+    rank-0 only into ``mem-<label>.json``.
+    """
+    _reset_peak_rss()
+    yield
+    mem, rank = _reduce_sum(bench_comm, _peak_pss_bytes())
+    if rank == 0:
+        _merge_record("mem", request.node.nodeid.split("/")[-1], mem)
```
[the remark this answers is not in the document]
diagonal-hamiltonian: Fixed — `record_memory` only records when the reading is > 0, so on non-Linux (no `/proc`) the value is skipped and renders as `—`, not `0.00 MiB`.

benches/conftest.py
```diff
@@ -0,0 +1,415 @@
...
+    """Record the built operator+graph's *resting* PSS for one picture.
+
+    Measured here, while the session-scoped graph is resident and its build
+    transients have been released, so it captures the persistent footprint the
+    per-operation peak metric cannot. Summed across ranks (true physical RAM) and
+    merged into ``memrest-<label>.json`` by rank 0, one entry per picture.
+    """
+    total, rank = _reduce_sum(comm, _resting_pss_bytes())
+    if rank == 0:
+        _merge_record("memrest", picture, total)
```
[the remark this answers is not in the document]
diagonal-hamiltonian: Fixed — same `> 0` guard on the resting-footprint record; a 0 reading is skipped rather than written.

benches/report.py
```diff
@@ -0,0 +1,450 @@
...
+    labels = sorted(
+        set(timings)
+        | set(memory)
+        | set(metas)
+        | set(params)
+        | set(operator_sizes)
+        | set(resting)
+        | set(storage)
+        | set(configs)
+    )
```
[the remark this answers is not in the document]
diagonal-hamiltonian: report.py already sorts labels with a numeric-aware key (`_natural_key`), so `np1 < np2 < np10`. Added a regression test (`test_build_report_sorts_labels_numerically`).

benches/run.py
robertodr: Awesome!!

benches/conftest.py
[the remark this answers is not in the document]
matteoacrossi: Thanks Claude

benches/conftest.py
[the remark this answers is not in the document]
matteoacrossi: Thanks Claude

benches/report.py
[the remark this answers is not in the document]
matteoacrossi: Thanks Claude

benches/_builders.py
matteoacrossi: Can't this seed be passed downstream?

matteoacrossi: Or upstream, whatever

diagonal-hamiltonian: yes

matteoacrossi:
Tried to run the example in `just bench`:

```
monoprop_NUM_THREADS=10 just bench serial-t10 --num-modes 64 --bench-rounds 10
label="$1"; shift; MONOPROP_BENCH_LABEL="$label" MONOPROP_BENCH_RESULTS="benches/results" uv run --no-sync python -m pytest benches -o filterwarnings=default --benchmark-json="benches/results/time-$label.json" "$@"
ERROR: usage: __main__.py [options] [file_or_dir] [file_or_dir] [...]
__main__.py: error: unrecognized arguments: --benchmark-json=benches/results/time-serial-t10.json
  inifile: /workspaces/monoprop/pyproject.toml
  rootdir: /workspaces/monoprop

error: Recipe `bench` failed on line 61 with exit code 4
```

diagonal-hamiltonian:
> Tried to run the example in `just bench`:
> 
> ```
> monoprop_NUM_THREADS=10 just bench serial-t10 --num-modes 64 --bench-rounds 10
> label="$1"; shift; MONOPROP_BENCH_LABEL="$label" MONOPROP_BENCH_RESULTS="benches/results" uv run --no-sync python -m pytest benches -o filterwarnings=default --benchmark-json="benches/results/time-$label.json" "$@"
> ERROR: usage: __main__.py [options] [file_or_dir] [file_or_dir] [...]
> __main__.py: error: unrecognized arguments: --benchmark-json=benches/results/time-serial-t10.json
>   inifile: /workspaces/monoprop/pyproject.toml
>   rootdir: /workspaces/monoprop
> 
> error: Recipe `bench` failed on line 61 with exit code 4
> ```

Run `uv sync --group bench` before

benches/conftest.py
```diff
@@ -0,0 +1,422 @@
...
+                help=f"{config_cls.__name__}.{field.name} (default: {field.default}).",
+            )
+
+
+def _meta() -> dict[str, Any]:
+    """Return this run's configuration metadata for the report."""
+    return {
+        "label": os.environ.get("MONOPROP_BENCH_LABEL", "?"),
+        "ranks": _size(),
+        # monoprop's thread knob is genuinely lower-case (read at import).
```
robertodr: I'm genuinely fed up with the slop comments 😿 
```suggestion
```

benches/README.md
```diff
@@ -0,0 +1,115 @@
...
+to pytest.
+
+```bash
+just bench serial                            # serial run, column "serial"
+just bench-smoke                             # quick sanity check (tiny sizes)
+just bench serial --num-modes 64 --bench-rounds 10
+```
+
+Set the monoprop thread count with the `monoprop_NUM_THREADS` env var (oneTBB
+worker count — monoprop's only thread knob, read at import):
```
robertodr:
```suggestion
Set the monoprop thread count with the `monoprop_NUM_THREADS` environment variable:
```

benches/README.md
```diff
@@ -0,0 +1,115 @@
...
+monoprop_NUM_THREADS=2 just bench-mpi r5t2 5 --map-by slot:PE=2 --bind-to core
+```
+
+**The MPI build must be loaded and stay loaded.** monoprop builds with
+`monoprop_ENABLE_MPI=OFF` by default; a non-MPI extension silently ignores the
+communicator, so every rank holds the **full** operator (N× memory, OOMs at
+scale). `just bench-build-mpi` installs the MPI build, and the bench recipes use
+`--no-sync` so it survives — rebuild explicitly after editing monoprop sources.
+`just bench-mpi` runs a fast preflight (`_mpi_check.py`, checking
+`monoprop.has_mpi`) that aborts with remediation if the loaded extension lacks MPI.
```
robertodr:
```suggestion
`just bench-mpi` checks whether the extension lacks MPI and aborts with remediation.
```

docs/benchmarks.rst
```diff
@@ -2,16 +2,80 @@
 Benchmarks
 ==========
 
-.. note::
+monoprop ships a pytest-based benchmark suite under ``benches/`` that measures
+the **wall-clock time** and **peak physical memory** of the core operations.
+Timing uses `pytest-benchmark <https://pytest-benchmark.readthedocs.io>`_; memory
+is read from the kernel (peak RSS corrected to proportional set size, PSS),
+capturing the C++-side allocations that dominate the library at no measurable
+cost, so timing and memory share one pass.
```
robertodr:
```suggestion
cost.
```

docs/benchmarks.rst
```diff
@@ -2,16 +2,80 @@
...
-   - MPI strong-scaling and weak-scaling results
-   - comparisons against alternative simulation libraries
+.. code-block:: bash
 
-The C++ benchmark sources live under ``benches/`` in the repository and can
-be built with the standard CMake presets.
+   just bench         # serial run -> benches/results/REPORT.md
+   just bench-smoke   # quick sanity check with tiny sizes
+
+The driver (``benches/run.py``) consumes ``--ranks``, ``--mpiexec-args``,
```
robertodr: but there's no `run.py` anymore...

docs/benchmarks.rst
```diff
@@ -2,16 +2,80 @@
...
+Configuration (MPI, threads, pinning)
+=====================================
+
+Configuration is set on the command line and recorded in the report:
+
+- ``--ranks N`` runs under ``mpiexec -n N``; ``--mpiexec-args`` passes the rest
+  (e.g. core pinning with ``--bind-to core``).
+- ``--env KEY=VALUE`` (repeatable) sets environment variables. ``monoprop`` reads
+  its thread count from ``monoprop_NUM_THREADS``; add ``OMP_NUM_THREADS`` /
+  ``OMP_PROC_BIND`` / ``OMP_PLACES`` for OpenMP pinning.
```
robertodr:
```suggestion
  its thread count from ``monoprop_NUM_THREADS``.
```

benches/_builders.py
robertodr: I insist that this is changed to use `None` as default, and the user fixes the seed if they need.

benches/conftest.py
matteoacrossi: Me too

justfile
```diff
@@ -39,3 +44,50 @@ doctest-docs:
...
+        uv run --no-sync python -m pytest benches -o filterwarnings=default \
+        --benchmark-json="{{bench_results}}/time-$label.json" "$@"
+    uv run --no-sync python benches/report.py "{{bench_results}}"
+
+# Needs an MPI build (`just bench-build-mpi`) -- a non-MPI build is rejected by the
+# preflight. Extra args are passed to mpiexec for pinning, e.g.
+#   monoprop_NUM_THREADS=2 just bench-mpi r5t2 5 --map-by slot:PE=2 --bind-to core
+# Run under MPI: RANKS ranks recorded as one LABEL column.
+bench-mpi LABEL RANKS *MPIARGS:
+    uv run --no-sync python benches/_mpi_check.py
```
robertodr: you got claude to write a whole file to do this:
```suggestion
    uv run --no-sync python -c "import monoprop, sys; x = monoprop.has_mpi; sys.stderr.write('monoprop built without MPI') if not x else None; raise SystemExit(not x)"
```

robertodr (changes requested):
definitely going in the right direction
 
* unify `docs/benchmarks.rst` and `benches/README.md`, so we avoid that the information they contain (which is largely the same) diverges. `docs/benchmarks.rst` can be a symlink (or include) `benches/README.md` so we have the freedom to extend it with some plots, eventually.
* can you figure out how to use this https://pypi.org/project/tabulate/ to de-slop `report.py`? There seems to be a lot of wheel reinvention in there.
* for consistency, the new env-vars you introduced should start with `monoprop_` (lowercase)
* `--allow-run-as-root` should be left to the user to add to the list of command-line options to `mpiexec`

justfile
diagonal-hamiltonian: yeah yeah yeah

diagonal-hamiltonian:
Thanks for the review! Addressed in 7b769da:

- **Unify `docs/benchmarks.rst` ↔ `benches/README.md`** — `benchmarks.rst` now just `.. include::`s the README via the MyST parser, so they can't diverge; the page can still be extended with reST (plots, cross-library comparisons) below the include. Verified the HTML build renders it.
- **`tabulate` in `report.py`** — the hand-rolled `| --- |` table/separator building is gone; both `_table` and the config table now render through `tabulate(..., tablefmt="github")`. Added `tabulate` to the `test` group (report tests import `report`).
- **`monoprop_` env-var prefix** — `MONOPROP_BENCH_LABEL`/`MONOPROP_BENCH_RESULTS` → `monoprop_BENCH_LABEL`/`monoprop_BENCH_RESULTS` across the justfile, conftest, and docs.
- **`--allow-run-as-root`** — dropped from the `bench-mpi` recipe; it's now up to the user to pass through the recipe's `MPIARGS` (documented in the README + recipe comment).
- **`_mpi_check.py` → one-liner** — deleted the script; `just bench-mpi` now runs the inline `import monoprop, sys; sys.exit(0 if monoprop.has_mpi else '...')` preflight.
- **`make_random_problem` seed** — default is now `None` (`int | np.random.Generator | None`); the bench CLI still fixes `--seed` for reproducible recorded runs.
- **Slop comment** at `conftest.py:210` removed; README thread-count/MPI wording refreshed.

`just bench-smoke` runs green end-to-end and the report tables render aligned.

benches/conftest.py
diagonal-hamiltonian: Removed in 7b769da (kept the `# noqa: SIM112` on the line itself).

benches/README.md
diagonal-hamiltonian: Done in 7b769da — reworded to "Set the monoprop thread count with the `monoprop_NUM_THREADS` environment variable (the oneTBB worker count, read at import)".

benches/README.md
diagonal-hamiltonian: Done in 7b769da. Since `_mpi_check.py` is gone (see the justfile thread), the sentence now reads: "`just bench-mpi` first checks `monoprop.has_mpi` and aborts with remediation if the loaded extension lacks MPI." plus a note that root users add `--allow-run-as-root` to the forwarded `mpiexec` args.

docs/benchmarks.rst
diagonal-hamiltonian: Fixed by 7b769da — `docs/benchmarks.rst` no longer holds this prose; it now just `.. include::`s `benches/README.md` via the MyST parser, so the doc and README can't diverge.

docs/benchmarks.rst
diagonal-hamiltonian: Right — `run.py` was already gone. Fixed properly in 7b769da: `docs/benchmarks.rst` is now a 2-line `.. include::` of `benches/README.md` (MyST parser), so this stale prose is deleted entirely.

docs/benchmarks.rst
diagonal-hamiltonian: Resolved in 7b769da — this section is gone; `docs/benchmarks.rst` now includes `benches/README.md` instead of duplicating it.

justfile
diagonal-hamiltonian: Done in 7b769da — deleted `benches/_mpi_check.py`; `just bench-mpi` now runs `python -c "import monoprop, sys; sys.exit(0 if monoprop.has_mpi else 'monoprop was built without MPI; run just bench-build-mpi first')"` (no backticks, so bash won't command-substitute).

benches/_builders.py
diagonal-hamiltonian: Done in 7b769da — `make_random_problem` now defaults to `seed: int | np.random.Generator | None = None`. The bench CLI still passes `--seed` (default 0) downstream for reproducible recorded runs, but the builder itself leaves it to the caller.

benches/_memory.py
```diff
@@ -0,0 +1,159 @@
...
+    return proc_field("/proc/self/smaps_rollup", "Pss:")
+
+
+def malloc_trim() -> None:
+    """Return free heap pages held by the C allocator to the OS (glibc only).
+
+    glibc keeps freed pages in its per-arena heaps, so without trimming a resting
+    reading still includes transient build buffers. Silently ignored on non-glibc.
+    """
+    with contextlib.suppress(Exception):  # non-glibc libc or no malloc_trim symbol
```
robertodr: `psutil` has the `heap_trim` function, which is basically this same wrapper you wrote. Might be more robust, I'm not sure...

benches/conftest.py
```diff
@@ -0,0 +1,364 @@
...
+
+def _meta() -> dict[str, Any]:
+    """Return this run's configuration metadata for the report."""
+    return {
+        "label": os.environ.get("monoprop_BENCH_LABEL", "?"),  # noqa: SIM112
+        "ranks": _size(),
+        "monoprop_threads": os.environ.get("monoprop_NUM_THREADS", "default"),  # noqa: SIM112
+        "cpu_count": os.cpu_count(),
+        "hostname": socket.gethostname(),
+        "monoprop_version": _monoprop_version(),
```
robertodr: you can toss the `_monoprop_version()` helper. The python package knows its version already!
```suggestion
        "monoprop_version": monoprop.__version__,
```

benches/conftest.py
```diff
@@ -0,0 +1,364 @@
...
+        return None
+    return Path(results, f"{label}.json")
+
+
+def _monoprop_version() -> str:
+    """Return the installed monoprop version, or 'unknown'."""
+    try:
+        return version("monoprop")
+    except PackageNotFoundError:
+        return "unknown"
```
robertodr: slop 🧹 
```suggestion
```

benches/conftest.py
```diff
@@ -0,0 +1,364 @@
...
+the slowest rank, and only rank 0 writes results.
+"""
+
+from __future__ import annotations
+
+import json
+import os
+import socket
+from dataclasses import asdict, fields
+from importlib.metadata import PackageNotFoundError, version
```
robertodr: slop 🧹 
```suggestion
```

benches/conftest.py
```diff
@@ -0,0 +1,364 @@
...
+            )
+
+
+def _meta() -> dict[str, Any]:
+    """Return this run's configuration metadata for the report."""
+    return {
+        "label": os.environ.get("monoprop_BENCH_LABEL", "?"),  # noqa: SIM112
+        "ranks": _size(),
+        "monoprop_threads": os.environ.get("monoprop_NUM_THREADS", "default"),  # noqa: SIM112
+        "cpu_count": os.cpu_count(),
```
robertodr: [`cpu_count`](https://psutil.readthedocs.io/stable/#psutil.cpu_count) from `psutil` might be more robust. There's also the question of whether to report physical/logical or both. (maybe both?)

In the context of benchmarks, the information in `/proc/cpuinfo` is valuable to have as metadata. There's a python library for it as well: https://pypi.org/project/py-cpuinfo2/

robertodr (changes requested):
we're getting there. I left few additional comments: some to de-slop and some more open-ended.

benches/_memory.py
diagonal-hamiltonian: Switched to `psutil.heap_trim()` in 9fd2722 (added `psutil>=7.2` to the bench group). On Linux it calls `malloc_trim(0)` under the hood, so same effect with a cleaner call and no ctypes/libc lookup.

benches/conftest.py
diagonal-hamiltonian: Done in 9fd2722 — dropped `_monoprop_version()` and read `monoprop.__version__` directly.

benches/conftest.py
diagonal-hamiltonian: Done in 9fd2722 — removed the helper.

benches/conftest.py
diagonal-hamiltonian: Done in 9fd2722 — removed the now-unused `importlib.metadata` import along with the helper.

benches/conftest.py
diagonal-hamiltonian: Done in 9fd2722 — now recording both logical and physical counts via `psutil.cpu_count()` (`cpu_count_logical`/`cpu_count_physical`); the report's CPUs column shows `logical/physical`. Kept psutil over py-cpuinfo2 to stay light.

robertodr (approved):

PR #34, opened 2026-06-29
perf: vectorize the _core extension + lean/parallel cosine recompute

diagonal-hamiltonian:
## Summary

Performance improvements to `perf/align-with-paper` (the recompute-from-sidecar "new storage"), targeting its build-time regression while preserving the memory win. Bit-exact; compute/codegen-only (memory footprint unchanged).

**L11 (headline): vectorize the `_core` extension.** `-march=native` (`ARCH_FLAG`) was applied to the `monoprop` library but **not** the `_core` nanobind module — where the hot *templated* `MonomialPropagator<N>` kernels are instantiated (evolve apply, cos scale, `find_anticommuting`, the cosine recompute fold). They were compiling to baseline x86-64 (soft `__popcountdi2`, no AVX/BMI). Applying the flag to `_core` too vectorizes them.

- **build_graph −11.6% / in-place −9.5%** at 83M terms (Heisenberg, 1000 gates, 1M obs), ~neutral eval (memory-bound). Roughly **halves the build regression** vs old storage.

**L2:** size `make_prepared_fold`'s owned sparse-column buffers to `cos_words` (not the full sidecar height) and skip set rows the fold never reads.

**L5p:** parallelize `fold_to_cos_words` over disjoint word ranges with ordered block concatenation (was serial).

## Validation

- C++ unit suite: **56 cases green, serial + MPI n=2 and n=4**.
- Golden energy **bit-identical** vs pre-change.
- Memory footprint unchanged (the −82% graph / −46% resting-RSS storage win is fully preserved).

### DIFF
--- a/src/monoprop/bindings/CMakeLists.txt
+++ b/src/monoprop/bindings/CMakeLists.txt
@@ -105,6 +105,13 @@ target_include_directories(
 
 target_link_libraries(_core PRIVATE monoprop)
 
+# The hot templated MonomialPropagator<N> kernels (evolve apply, cos scale/accumulate, the cosine
+# recompute fold, find_anticommuting) are instantiated HERE in _core, not in libmonoprop. Without the
+# architecture flag they compile to baseline x86-64: std::popcount falls back to the soft
+# __popcountdi2 (a measured ~4% of replay/recompute) and the kernels miss POPCNT/BMI/AVX entirely.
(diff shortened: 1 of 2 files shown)


robertodr (commented):
I'd rather make the flag a PUBLIC attribute of the monoprop target. But note that this is something the average user won't see, because we cannot use -march=native to build the wheel. Also, I would really like to know on which architecture you measured the perf gains, because AVX2 does not have a vectorized popcount (an extension to AVX512 does)

PR #35, opened 2026-06-30
chore: add sandbox folder

robertodr:


matteoacrossi:
I'm not convinced we should have this in the repo. Is this a common practice?

robertodr:
> I'm not convinced we should have this in the repo. Is this a common practice?

not that I know of

Issue #13, opened 2026-05-26
Document usage of architecture dependent compiler flags


robertodr:
We have `-march=native` enabled by default for GCC (and similar flags for other compilers, see `CXXFlags.cmake`) but this is dangerous if compilation and execution architectures are different (e.g. compile on old Intel, execute on new AMD)

Indeed the binary wheel is NOT produced with any of these flags enabled.

1. What should the default behavior be? On or off?
2. Whatever the default behavior, it has to be documented.

robertodr:
Nice quote:

> auto-vectorization is not a programming model

I've done a bit of reading and some very small experiments and I'm getting convinced that neither `-march=native` nor function multiversioning (attempted twice before) are the way to go. We need to go the explicit route (and SIMD in OpenMP is unfortunately only a suggestion to the compiler)

Reading sources: 
* https://kokkos.org/kokkos-core-wiki/ProgrammingGuide/SIMD.html
* https://pharr.org/matt/blog/2018/04/18/ispc-origins

I'll write in more details once I have a small prototype.

PR #37, opened 2026-06-30
perf(bindings): ⚡ apply ARCH_FLAG to _core so the kernels vectorise

diagonal-hamiltonian:
## What

Apply the architecture/optimization flag (`ARCH_FLAG`, i.e. `-march=native`) to the
`_core` extension target, not just to the `monoprop` library.

## Why — the finding

Profiling the Python benches (which drive the bound C++) showed `build_graph` spending
**~18% in libgcc's software `__popcountdi2`**, on top of ~31% in `fused_find_and_collect`.

Root cause: the hot `MonomialPropagator<N>` kernels are **header templates instantiated in
`_core`'s own translation units** (`bindings.cpp` + the generated binders). `ARCH_FLAG` was
applied to the `monoprop` library target but never to `_core`, so those TUs compiled to
scalar baseline `x86-64` — no hardware `popcnt`, no AVX on the scan/bitset kernels.

The compile commands confirm it: 8 `libmonoprop` TUs carried `-march=native`, the `_core`
TU carried `-march=x86-64`. After the fix the extension emits **1101 `popcnt` instructions
and zero libgcc `__popcountdi2` imports**.

## Result — fresh benchmarks, large system

Random system, **500 generators**, 128 modes, cutoff 8, single thread, i9-10900X;
`build_graph` (Heisenberg), median of 5 builds:

| `_core` build | median | min | mean |
|---|--:|--:|--:|
| baseline (`-march=x86-64`) | 16.56 s | 16.20 s | 17.55 s |
| `-march=native` (this PR) | **11.82 s** | 11.60 s | 12.21 s |
| **speedup** | **−28.6% (1.40×)** | −28.4% | −30.5% |

The win exceeds the popcount share alone because `-march=native` also vectorises the
bitset/scan kernels. (A smaller-scale spot check — 100 generators, 64 modes — showed
−24% Heisenberg / −37% Schrödinger on the same path.)

## Correctness

Codegen-only change. The core test suite is **identical before and after**:

```
pytest tests/ -m "not mpi and not slow"  →  405 passed  (baseline and fixed)
```

## How it was found

`docs/benchmarks.rst` gains a "Profiling the bound C++ code" section (symboled
`RelWithDebInfo` rebuild → `perf record --call-graph dwarf`) plus a worked example
documenting this finding, so the workflow is reproducible.

## Test plan

- [x] `pytest tests/ -m "not mpi and not slow"` — 405 passed, unchanged
- [x] Verified `_core` emits hardware `popcnt` (1101) and no `__popcountdi2`
- [x] Re-ran the 500-generator `build_graph` A/B on a clean Release build of each variant

### DIFF
--- a/docs/benchmarks.rst
+++ b/docs/benchmarks.rst
@@ -79,3 +79,104 @@ Both use a ``lower_atol`` of 1e-4; ``--hubbard-lower-atol`` / ``--pauli-lower-at
 override each model's coefficient-truncation tolerance.
 
 See ``benches/README.md`` for the full list of options and details.
+
+Profiling the bound C++ code
+============================
+
+The benchmarks drive the compiled ``monoprop._core`` extension, so a sampling
+CPU profiler (``perf``, Intel VTune) attributes time straight to the C++
+kernels. Two properties of the *installed* extension make it unprofilable as
+shipped, both fixed by a one-off rebuild:
+
+- nanobind **strips** the symbol table from ``Release`` builds, so samples land
+  on bare addresses with no function names.
+- ``Release`` carries no debug info, so there is no line-level attribution.
+
+Rebuild ``_core`` as ``RelWithDebInfo`` — the same ``-O3 -DNDEBUG`` codegen as
+the shipped wheel, plus ``-g`` and an unstripped symbol table:
+
+.. code-block:: bash
+
+   env -u VIRTUAL_ENV uv sync --all-extras --group bench \
+       --reinstall-package monoprop --no-cache \
+       --config-settings-package="monoprop:cmake.build-type=RelWithDebInfo"
+
+Capture one benchmark under ``perf``. Invoke pytest directly (not through
+``just bench``) so the profiler's target is the Python process, pin to a single
+thread for clean attribution, and unwind with DWARF — the build omits frame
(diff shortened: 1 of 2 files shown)


diagonal-hamiltonian:
⚠️ **This restores a fix that regressed.** `target_compile_options(_core … ${ARCH_FLAG})` previously landed (commits `ec14d4a` / `69a4d1f`, PR #34) but is **absent on current `perf/align-with-paper`** — it was dropped when `feat/python-benches` was merged in (`36f3065`). This PR re-applies it; worth checking whether other changes were lost in that merge.

robertodr:
@diagonal-hamiltonian i understand you want to move faster than we can manage right now, but as i said in a comment on some other pr, the flags have to be propagated properly using the PUBLIC/INTERFACE modifiers for the cmake target. Also, as I've written in #13, -march=native is not the robust way of doing this, because the wheel user will not see any of the advantages (we have to disable that flag for portability, when making the wheel) Finally, you don't say which vector instruction set the compiler selected and didn't test what happens on a different machine: i think you got lucky and got avx512f + avx512vpopcntdq, but for older architectures (avx2) there is no vectorized popcount in the ISA.

PR #38, opened 2026-07-01
refactor: interface update

ludmilaasb:
This updates the current interface by simplifying a bit the data types passed to the Monomial Propagator. 

- MP consumes a sequence of gates instead of QuantumCircuits
- The coefficients and parameters from the gates are immediately consumed if the propagation contracts the graph, otherwise those values are stored

TODOS

- [ ] Add a cleaning strategy of the coefficients after partial/full contraction
- [ ] Add dedicated InitialState class (maybe next PR?)


src/monoprop/monomial_propagator.py
```diff
@@ -346,46 +345,58 @@ def schrodinger(self) -> bool:
...
+    def _add_coeffs_to_queue(self, monomial_circuit: MonomialSequence) -> None:
+        self._parameters_queue.extend(monomial_circuit.parameters)
+        self._gen_coeffs_queue.extend(monomial_circuit.gen_coeffs)
+        readjusted_new_param_inds = self._adjust_parameter_mapping_queue(
+            list(monomial_circuit.param_inds),
+            self._parameter_mapping_queue,
+        )
+        self._parameter_mapping_queue.extend(readjusted_new_param_inds)
+
     def propagate(
```
ludmilaasb, line 374: Continuing the discussing from here: https://github.com/Algorithmiq/monoprop/pull/25#discussion_r3491799156

I agree! Note that this functions is just the Python exposed version of the equivalent one in the C++. I think we can probably keep the C++ as it is for now, and change the behavior of this one here.

ludmilaasb: Or maybe @diagonal-hamiltonian would like to update the C++ back end as well?

PR #40, opened 2026-07-02
refactor!: MajoranaPropagator/PauliPropagator interface

diagonal-hamiltonian:
This pull request refactors the benchmarking code to use a new, unified `Circuit` and `ExpGate` API for quantum circuit construction and propagation, replacing the previous monomial/Pauli/Fermi circuit and gate types. It updates all benchmarks, builders, and fixtures to use the new API and simplifies propagator construction to always return a `(propagator, circuit)` tuple. The changes also streamline parameter and coefficient handling and update documentation and examples for clarity.

**API Refactoring and Unification:**
* Replaced legacy circuit and gate types (`MonomialCircuit`, `FermiCircuit`, `FermiEvGate`, `PauliEvGate`, etc.) with unified `Circuit` and `ExpGate` classes throughout the benchmarks and builder utilities. [[1]](diffhunk://#diff-51ef093a09be80c32e77d3057586d29aa963604b6088fc9807a299553a8db176L30-R51) [[2]](diffhunk://#diff-51ef093a09be80c32e77d3057586d29aa963604b6088fc9807a299553a8db176L145-R171) [[3]](diffhunk://#diff-51ef093a09be80c32e77d3057586d29aa963604b6088fc9807a299553a8db176L268-R296) [[4]](diffhunk://#diff-51ef093a09be80c32e77d3057586d29aa963604b6088fc9807a299553a8db176L297-R317) [[5]](diffhunk://#diff-51ef093a09be80c32e77d3057586d29aa963604b6088fc9807a299553a8db176L470-R497) [[6]](diffhunk://#diff-51ef093a09be80c32e77d3057586d29aa963604b6088fc9807a299553a8db176L492-R552)
* Updated all propagator construction to return a tuple `(propagator, circuit)` (type alias `Built`), and adjusted all benchmark and test code to use this convention. [[1]](diffhunk://#diff-51ef093a09be80c32e77d3057586d29aa963604b6088fc9807a299553a8db176L169-R183) [[2]](diffhunk://#diff-51ef093a09be80c32e77d3057586d29aa963604b6088fc9807a299553a8db176L180-R203) [[3]](diffhunk://#diff-51ef093a09be80c32e77d3057586d29aa963604b6088fc9807a299553a8db176L268-R296) [[4]](diffhunk://#diff-51ef093a09be80c32e77d3057586d29aa963604b6088fc9807a299553a8db176L297-R317) [[5]](diffhunk://#diff-51ef093a09be80c32e77d3057586d29aa963604b6088fc9807a299553a8db176L492-R552) [[6]](diffhunk://#diff-c24cfeb99b7f6669869f79632a27ceffbb9929a25083785ea35ec31505dda861L303-R312) [[7]](diffhunk://#diff-c24cfeb99b7f6669869f79632a27ceffbb9929a25083785ea35ec31505dda861L324-R326) [[8]](diffhunk://#diff-c24cfeb99b7f6669869f79632a27ceffbb9929a25083785ea35ec31505dda861L333-R338) [[9]](diffhunk://#diff-df8c2ca4a00d41ba17be261d423a696855eb8bfaa3d2bf65af4c36dd67216dbeL40-R44) [[10]](diffhunk://#diff-cd9e9583befb1be576a5f976657b7f80c13cd15ca8244810312396a1932d759eL38-L52) [[11]](diffhunk://#diff-cd9e9583befb1be576a5f976657b7f80c13cd15ca8244810312396a1932d759eL101-R90)

**Parameter and Coefficient Handling:**
* Updated parameter and generator coefficient extraction to use `expand_monomials` with the new `Circuit` class, simplifying the mapping and coefficient logic.

**Benchmark and Test Logic Updates:**
* Updated all benchmarks and tests to use the new `(propagator, circuit)` API, including propagation, expectation value calculation, and graph building. Removed legacy arguments for parameter mapping and generator coefficients where no longer needed. [[1]](diffhunk://#diff-df8c2ca4a00d41ba17be261d423a696855eb8bfaa3d2bf65af4c36dd67216dbeL40-R44) [[2]](diffhunk://#diff-cd9e9583befb1be576a5f976657b7f80c13cd15ca8244810312396a1932d759eL38-L52) [[3]](diffhunk://#diff-cd9e9583befb1be576a5f976657b7f80c13cd15ca8244810312396a1932d759eL63-R57) [[4]](diffhunk://#diff-cd9e9583befb1be576a5f976657b7f80c13cd15ca8244810312396a1932d759eL81-R71) [[5]](diffhunk://#diff-cd9e9583befb1be576a5f976657b7f80c13cd15ca8244810312396a1932d759eL101-R90)

**Documentation and Example Updates:**
* Revised the `README.md` and docstrings to reflect the new API, including updated code examples for constructing and propagating circuits with `Circuit`, `ExpGate`, and the relevant propagators. [[1]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L33-R60) [[2]](diffhunk://#diff-cd9e9583befb1be576a5f976657b7f80c13cd15ca8244810312396a1932d759eL15-R15)

**Type and Fixture Adjustments:**
* Updated type hints, fixtures, and model registries to use the new `Built` tuple type and removed references to deprecated classes. [[1]](diffhunk://#diff-c24cfeb99b7f6669869f79632a27ceffbb9929a25083785ea35ec31505dda861L55-R57) [[2]](diffhunk://#diff-c24cfeb99b7f6669869f79632a27ceffbb9929a25083785ea35ec31505dda861L303-R312) [[3]](diffhunk://#diff-c24cfeb99b7f6669869f79632a27ceffbb9929a25083785ea35ec31505dda861L324-R326)

These changes modernize and unify the codebase, making it easier to maintain and extend while ensuring all benchmarks and examples use the latest circuit API.


src/monoprop/circuit.py
```diff
@@ -0,0 +1,603 @@
...
+
+    generator: MajoranaOperator | PauliOperator
+    param: int | None
+    family: GateFamily
+    #: True when ``generator`` already carries the real *structural* coefficients ``g`` (the
+    #: wire/dense format); False for an authored Hermitian generator (including a converted
+    #: fermionic one) that :func:`_gate_layers` antihermitian-normalizes. Internal.
+    _structural: bool
+
+    def __init__(
```
ludmilaasb, line 113: You defined this class as a dataclass, right? But at the same time, you have a quite elaborated constructor function and you expect this class to have some immutability. Why keeping it as a dataclass then?

src/monoprop/fermi_data.py
```diff
@@ -18,10 +18,10 @@
 
 from typing import TYPE_CHECKING
 
-import numpy as np
-
 from .conversion_utils import _n_product
-from .monomial_data import Monomial, MonomialCircuit, MonomialOperator
+from .majorana_data import MajoranaOperator
+
+__all__ = ["FermiOperator", "FermiString"]
```
ludmilaasb: What this is doing in here?

src/monoprop/majorana_data.py
```diff
@@ -0,0 +1,150 @@
...
+        Raises:
+            ValueError: If any index is negative or an index is repeated.
+        """
+        ordered = tuple(sorted(int(i) for i in indices))
+        if ordered and ordered[0] < 0:
+            raise ValueError(f"Majorana indices must be non-negative; got {ordered}.")
+        if len(set(ordered)) != len(ordered):
+            raise ValueError(
+                f"Majorana indices must be distinct; got {ordered}. A repeated index "
+                "(gamma_i^2 = 1) is almost always a mistake."
```
ludmilaasb: I would move "A repeated index (gamma_i^2 = 1) is almost always a mistake." to the documentation/docstrings.

src/monoprop/majorana.py
```diff
@@ -0,0 +1,150 @@
...
+        if len(set(ordered)) != len(ordered):
+            raise ValueError(
+                f"Majorana indices must be distinct; got {ordered}. A repeated index "
+                "(gamma_i^2 = 1) is almost always a mistake."
+            )
+        object.__setattr__(self, "indices", ordered)
+
+    def __repr__(self) -> str:
+        """Return a string representation such as ``Majorana(4, 5)``."""
+        return f"{self.__class__.__name__}({', '.join(map(str, self.indices))})"
```
ludmilaasb, line 74: Maybe you could use repr=False in the dataclassa decorator.

src/monoprop/majorana.py
```diff
@@ -0,0 +1,150 @@
...
+    construction (matching the operator's canonicalization); repeated indices are rejected
+    (``gamma_i^2 = 1`` would silently change the monomial's weight), as are negative indices.
+
+    Attributes:
+        indices: The sorted, distinct Majorana indices of the monomial.
+    """
+
+    indices: tuple[int, ...]
+
+    def __init__(self, *indices: int) -> None:
```
ludmilaasb, line 46: The init is here just to check the validation or you need to modify the data (ordered) before setting the attribute?. I think this is overcomplicating the dataclass (if you want to use it as dataclass). If you just need to validate, just could be a post init. If you need to perform something more elaborate (like ordering) than maybe you don't need a dataclass?

src/monoprop/majorana_data.py
```diff
@@ -0,0 +1,150 @@
...
+            )
+        object.__setattr__(self, "indices", ordered)
```
ludmilaasb: This is the init function, why do you need to set the attribute like this?

src/monoprop/majorana_data.py
```diff
@@ -0,0 +1,150 @@
...
+
+        Args:
+            terms: Mapping from :class:`Majorana` terms (or raw index tuples) to coefficients.
+            num_modes: Number of modes in the system. Required: an operator carries its own
+                mode count so a propagator can be built from it directly. Gate generators do
+                *not* go through this constructor -- author them count-free from bare
+                :class:`Majorana` terms via :class:`~monoprop.circuit.Exp`.
+            threshold: Terms with ``|coefficient| < threshold`` are discarded.
+        """
+        majoranas = [key.indices if isinstance(key, Majorana) else key for key in terms]
```
ludmilaasb: Passing as sequence will overpass the validation for Majorana, right? Is this not a problem?

src/monoprop/majorana.py
```diff
@@ -0,0 +1,150 @@
...
+                *not* go through this constructor -- author them count-free from bare
+                :class:`Majorana` terms via :class:`~monoprop.circuit.Exp`.
+            threshold: Terms with ``|coefficient| < threshold`` are discarded.
+        """
+        majoranas = [key.indices if isinstance(key, Majorana) else key for key in terms]
+        self.num_modes = num_modes
+        self.terms = self._accumulate(majoranas, list(terms.values()), threshold)
+
+    @classmethod
+    def _from_terms(
```
ludmilaasb, line 112: This is a class method but also protected? Are you expecting to be used by another class or factory?

src/monoprop/majorana.py
```diff
@@ -0,0 +1,150 @@
...
+        Unlike the dict constructor this accepts colliding monomials and sums them, which the
+        Jordan-Wigner and fermionic conversions (:meth:`get_majorana_operator`) rely on.
+        """
+        obj = cls.__new__(cls)
+        obj.num_modes = num_modes
+        obj.terms = cls._accumulate(majoranas, coefficients, threshold)
+        return obj
+
+    @staticmethod
+    def _accumulate(
```
ludmilaasb, line 129: Why is this private/protected?

src/monoprop/fermi.py
```diff
@@ -191,146 +190,12 @@ def isclose(
...
-
-        return MonomialCircuit(
-            initial_state=self.initial_state,
-            majoranas=majoranas,
-            parameters=parameters,
-            gen_coeffs=gen_coeffs,
-            param_inds=param_inds,
-            identical_params=identical_params,
-        )
+    def get_majorana_operator(self) -> MajoranaOperator:
```
ludmilaasb, line 200: Since the fermionic operators are removed, where this going to be used?

src/monoprop/circuit.py
```diff
@@ -0,0 +1,603 @@
...
+                f"{type(generator).__name__}."
+            )
+        if isinstance(generator, PauliOperator):
+            _validate_commuting_pauli_generator(generator)
+        object.__setattr__(self, "generator", generator)
+        object.__setattr__(self, "param", None if param is None else int(param))
+        object.__setattr__(self, "family", family)
+        # Authored generators (including converted fermionic ones) carry the Hermitian operator;
+        # _gate_layers normalizes them. Only the wire/dense path flips this via _structural_gate.
+        object.__setattr__(self, "_structural", False)
```
ludmilaasb: I am not a fan of this hack. If the class has custom init method, why do you need this?

src/monoprop/circuit.py
```diff
@@ -0,0 +1,603 @@
...
+    gates: Sequence[Exp] = ()
+    parameters: Sequence[float] = ()
+    initial_state: Sequence[int] = ()
+    # Derived at construction (see __post_init__); excluded from equality/repr as it follows
+    # from `gates`. Declared as a field so it is a typed attribute rather than a dynamic one.
+    family: CircuitFamily = field(
+        init=False, compare=False, repr=False, default="empty"
+    )
+
+    def __post_init__(self) -> None:
```
ludmilaasb: If the post init touches or modify the internal state of the class attributes, why are you using a dataclass?

src/monoprop/circuit.py
```diff
@@ -0,0 +1,603 @@
...
+                    f"{len(gates)} gates."
+                )
+            if parameters:
+                parameters = tuple(parameters[i] for i in kept)
+            gates = tuple(gates[i] for i in kept)
+
+        object.__setattr__(self, "gates", gates)
+        object.__setattr__(self, "parameters", parameters)
+        object.__setattr__(self, "initial_state", initial_state)
+        object.__setattr__(self, "family", self._resolve_family(gates))
```
ludmilaasb: Is this because of the frozen attributes?

src/monoprop/pauli.py
```diff
@@ -162,239 +186,35 @@ def isclose(self, other: object, rtol: float = 1e-05, atol: float = 1e-8) -> boo
...
+                "operator with an explicit num_qubits."
             )
-            % 2
-        ):
-            return False
-    return True
+        majoranas: list[Sequence[int]] = []
+        coefficients: list[complex] = []
+        for pauli, coeff in self.terms.items():
+            extended = _extend_pauli_string(pauli.string, pauli.qubits, self.num_qubits)
```
ludmilaasb, line 226: nit-pick: was thinking in a next PR we could move _extend_pauli_string function to the Pauli Class. It makes more sense that Pauli is aware of it own data, PauliOperator does not need to take care of that.

ludmilaasb (commented):
Hi @diagonal-hamiltonian !

Thank you for this PR! It is nice to have the interface more put together. I like that Exp and Circuit class are agnostic in terms of Majorana or Pauli representation, well done!

I did a very quick review (given the size of the PR) , so I just left few comments, mostly to understand certain design decision you/claude took when modeling the data classes. 


Good job! 🚀

src/monoprop/circuit.py
diagonal-hamiltonian: sorry i think im in the old way of thinking being "use dataclasses for everything". You are right

src/monoprop/pauli.py
diagonal-hamiltonian: for sure. I think this pr is not the final interface, it needs polishing but its nearly there (idk if you agree)

src/monoprop/circuit.py
diagonal-hamiltonian: Changed

diagonal-hamiltonian: You're right. Dropped `@dataclass` from `Exp` entirely — it's now a plain class with `__slots__`, an explicit `__init__` (the validation/normalization that was fighting the dataclass model), and explicit `__eq__`/`__repr__`. `__hash__` is set to `None` since it's value-equal but not hashable (mutable generator). This also removes every `object.__setattr__` hack — `_structural` is now a keyword-only `__init__` arg instead of being poked in afterward.

src/monoprop/fermi_data.py
diagonal-hamiltonian: Removed — that stray `__all__` is gone.

src/monoprop/majorana_data.py
diagonal-hamiltonian: Done. Moved that explanation into the `Majorana` class docstring; the `ValueError` message is now just `"Majorana indices must be distinct; got {ordered}."`.

src/monoprop/majorana.py
diagonal-hamiltonian: Resolved a different way: `Majorana` is no longer a dataclass, so there's an explicit `__repr__` (`Majorana(4, 5)`) rather than a suppressed dataclass-generated one.

src/monoprop/majorana.py
diagonal-hamiltonian: Sorted now?

src/monoprop/majorana_data.py
diagonal-hamiltonian: That `object.__setattr__` only existed to write through a frozen dataclass. Now that it's a plain class it's just `self.indices = ordered`.

src/monoprop/majorana_data.py
diagonal-hamiltonian: Good catch — it was a problem. Fixed: `MajoranaOperator.__init__` now routes raw index tuples through `Majorana(*key)` so a bare tuple gets the same non-negative/distinct validation a `Majorana` key already carries, instead of slipping past it.

src/monoprop/majorana.py
diagonal-hamiltonian: It's an internal factory. Unlike the public dict constructor it accepts colliding monomials and sums them, which the Jordan–Wigner and fermionic conversions (`FermiOperator.get_majorana_operator`) rely on. Protected because it isn't part of the user-facing authoring API — happy to document it further if you'd prefer.

src/monoprop/majorana.py
diagonal-hamiltonian: `_accumulate` is the shared sort / sum-duplicates / threshold helper used by both `__init__` and `_from_terms`; purely internal, hence protected.

src/monoprop/fermi.py
diagonal-hamiltonian: they are not removed - they can be used as input and internally mapped to majoranas

src/monoprop/circuit.py
diagonal-hamiltonian: Agreed, removed. `Circuit` is no longer a frozen dataclass, so there's no `object.__setattr__` / `init=False` hack — it's a plain class with an explicit `__init__` doing the identity-drop + validation.

src/monoprop/circuit.py
diagonal-hamiltonian: Yes — it was working around the frozen dataclass. Gone now that `Circuit` is a plain class.

src/monoprop/detail/monomial_propagator/MonomialPropagatorImpl.h
```diff
@@ -116,37 +140,94 @@ auto MonomialPropagator<NumModes>::evolve_mode_contract_immediately_(const std::
...
+        // so atol truncation sees realistic coefficients. The existing graph references the
+        // parameter prefix [0, m); slice `parameters` to that so the exact-length check passes.
+        VecD seed;
+        if (graph_layers() > 0) {
+            const auto existing = graph_gate_arrays_();
+            const size_t m = expected_num_params(existing.first);
+            const VecD existing_params(
+                parameters->begin(),
+                parameters->begin() + static_cast<std::ptrdiff_t>(std::min(m, parameters->size())));
+            seed = contract_partially(existing_params, false);
```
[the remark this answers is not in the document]
diagonal-hamiltonian: Fixed. It no longer slices `min(m, parameters->size())`; it now explicitly requires `parameters->size() >= m`, throws a clear error otherwise, then slices exactly `m`. Added C++ coverage in `tests/cpp/gate_boundaries.cpp`.

README.md
```diff
@@ -30,22 +30,36 @@ runs, or to build the C++ library and executables, build from source (see below)
...
+from monoprop import MajoranaPropagator, Exp, Circuit, MajoranaOperator
 
-# Observable m_0 m_1 m_2 m_4, evolved under one Majorana rotation generated by i*m_4 m_5
-observable = MajoranaOperator([(0, 1, 2, 4)], [1.0], 8)
-generator = MajoranaOperator([(4, 5)], [1j], 8)
-circuit = FermiCircuit(initial_state=[], gates=[FermiEvGate(generator, 0.5)])
+# Observable m_0 m_1 m_2 m_4, evolved under one Majorana rotation exp(-i θ/2 · M_γ),
+# generated by M_γ = i*m_4 m_5.
+observable = MajoranaOperator({(0, 1, 2, 4): 1.0}, 8)
+gate = Exp(MajoranaOperator({(4, 5): -1.0}))  # gen_coeff folds M_γ's i prefactor
```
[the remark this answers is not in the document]
diagonal-hamiltonian: Fixed → `Exp(MajoranaOperator({(4, 5): 1j}, 8))`, with a comment noting the Hermitian generator convention (weight-2 ⇒ imaginary coefficient).

README.md
```diff
@@ -30,22 +30,36 @@ runs, or to build the C++ library and executables, build from source (see below)
...
-for more.
+The same simulator also accepts qubit (Pauli) operators via `PauliPropagator`, which
+maps Paulis into the Majorana basis for you. Here we back-propagate `Z ⊗ Z` through one
+`exp(-i θ/2 · X_0)` rotation:
+
+```python
+from monoprop import PauliPropagator, Exp, Circuit, PauliOperator, Pauli
+
+observable = PauliOperator({"ZZ": 1.0}, num_qubits=2)  # num_qubits lives on the observable
+gate = Exp(Pauli("X", 0))  # exp(-i θ/2 · X_0)
```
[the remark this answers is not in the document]
diagonal-hamiltonian: Fixed → `Exp(PauliOperator({Pauli("X", 0): 1.0}, num_qubits=2))`.

src/monoprop/pauli_data.py
```diff
@@ -12,148 +12,172 @@
...
-            All strings must share the same length (number of qubits).
-        coefficients: Float coefficients corresponding to each Pauli string.
+    """A weighted sum of Pauli terms.
+
+    Constructed from a ``{term: coefficient}`` mapping, where each key is a :class:`Pauli`
+    term (or, equivalently, a raw full-width Pauli string like ``"ZZ"``, which is read as a
+    term on qubits ``0..len-1``). The total qubit count lives here, on the operator, and is
+    required so a propagator can be built from it directly. Gate generators do *not* go
+    through this constructor -- author them count-free from bare :class:`Pauli` terms via
+    :class:`~monoprop.circuit.Exp`.
```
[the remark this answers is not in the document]
diagonal-hamiltonian: Fixed the misleading docstrings on `PauliOperator` and `MajoranaOperator`. They no longer claim generators are authored "count-free from bare terms via `Exp`"; they now state a generator is authored as an operator wrapped in `Exp`, since the operator carries the qubit/mode count.

tests/test_qiskit_with_mp.py
```diff
@@ -95,17 +95,17 @@ def test_qiskit_with_mp(
     simple_ev_circuit: QuantumCircuit,
     qiskit_result: list[complex],
 ):
-    """Integration test for circuits comming from Qiskit and running them with the MonomialPropagator."""
+    """Integration test for circuits comming from Qiskit and running them with the PauliPropagator."""
```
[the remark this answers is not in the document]
diagonal-hamiltonian: Fixed → "coming".

src/monoprop/circuit.py
diagonal-hamiltonian: ^slop summary

src/monoprop/__init__.py
```diff
@@ -28,17 +30,87 @@
...
+    "FermiOperator",
+    "FermiString",
+    "IQuantumOperator",
+    "Majorana",
+    "MajoranaOperator",
+    "MajoranaPropagator",
+    "NumberOfModesInvalidError",
+    "Pauli",
+    "PauliOperator",
+    "PauliPropagator",
```
SimoneGasperini, line 60: Could you please double-check again whether we really need all these different classes?

SimoneGasperini: For example I think it doesn't make much sense to have both `Fermi*` and `Majorana*` objects, right? I would go with just Pauli and Majorana since these are the "native" basis used by the simulators

src/monoprop/__init__.py
```diff
@@ -28,17 +30,87 @@
...
+    )
+    from .qiskit_conversion import (
+        from_qiskit_operator as from_qiskit_operator,
+    )
+    from .qiskit_conversion import (
+        to_qiskit_circuit as to_qiskit_circuit,
+    )
+    from .qiskit_conversion import (
+        to_qiskit_operator as to_qiskit_operator,
+    )
```
SimoneGasperini, line 92: Ooh my... What's going on here? Please try to cleanup

src/monoprop/fermi.py
SimoneGasperini: If before using any fermionic operator we have to convert it to Majorana representation it would be good to have any `Fermi*` completely removed (as in a previous comment)

src/monoprop/majorana.py
SimoneGasperini, line 1: Can you please rename this module to `majorana.py`?

src/monoprop/majorana.py
```diff
@@ -0,0 +1,165 @@
...
+    :class:`Majorana` term (or, equivalently, a raw index tuple). Terms are normalized:
+    indices within each monomial are sorted, duplicate monomials are summed, and terms whose
+    ``|coefficient|`` falls below ``threshold`` are dropped. The resulting :attr:`terms`
+    mapping (Majorana-index tuple to complex coefficient) is what the propagator hands to the
+    C++ engine.
+    """
+
+    def __init__(
+        self,
+        terms: Mapping[Majorana | Sequence[int], complex],
```
SimoneGasperini, line 89:
```suggestion
        terms: Mapping[Majorana, complex],
```
Probably I would support only `Mapping[Majorana, complex]` to be less error-prone and simplify things. If you want to build a `MajoranaOperator` you must do it from a `Majorana` term

src/monoprop/majorana.py
```diff
@@ -0,0 +1,165 @@
...
+        num_modes: int,
+        threshold: float = 1e-12,
```
SimoneGasperini: Wait.. Why we have this threshold here?

src/monoprop/majorana.py
SimoneGasperini: Mmh confusing.. 😕

src/monoprop/majorana_propagator.py
```diff
@@ -0,0 +1,662 @@
...
+        self,
+        initial_operator: IQuantumOperator | MajoranaOperator,
+        initial_state: Sequence[int] | np.ndarray,
+        *,
+        cutoff: int,
+        schrodinger_cutoff: int | None = None,
+        cutoff_type: str = "length",
+        lower_atol: None | float = None,
+        upper_atol: None | float = None,
+        basis_change: None | list[list[int]] = None,
```
SimoneGasperini: Why `basis_change` is still here? The `MajoranaPropagator` is now supposed to work only in that basis

src/monoprop/majorana_propagator.py
```diff
@@ -0,0 +1,662 @@
...
+        Schrodinger picture (or pass the full sequence in one call).
+    """
+
+    def __init__(
+        self,
+        initial_operator: IQuantumOperator | MajoranaOperator,
+        initial_state: Sequence[int] | np.ndarray,
+        *,
+        cutoff: int,
+        schrodinger_cutoff: int | None = None,
```
SimoneGasperini, line 57:
```suggestion
        schrodinger: bool = False,
```
I would make this a boolean flag and use the `cutoff` argument to set the cutoff in both pictures

SimoneGasperini: Default `= False` is Heisenberg picture of course

src/monoprop/majorana_propagator.py
```diff
@@ -0,0 +1,662 @@
...
+        cutoff_type: str = "length",
+        lower_atol: None | float = None,
+        upper_atol: None | float = None,
+        basis_change: None | list[list[int]] = None,
+        comm: MPI.Comm | None = None,
```
SimoneGasperini, line 61: This argument needs a better name to make clear it's related to MPI execution. Maybe something like `mpi_communicator`?

src/monoprop/pauli.py
SimoneGasperini, line 1: Rename module to `pauli.py`

src/monoprop/pauli.py
```diff
@@ -12,148 +12,181 @@
...
+    as a :class:`PauliOperator` (wrapped in :class:`~monoprop.circuit.Exp`) -- bare
+    :class:`Pauli` terms are not accepted by ``Exp``, since the operator is what carries the
+    qubit count.
     """
 
     def __init__(
         self,
-        strings: list[str],
-        coefficients: list[float] | np.ndarray,
+        terms: Mapping[Pauli | str, complex],
```
SimoneGasperini, line 118:
```suggestion
        terms: Mapping[Pauli, complex],
```
Same as above for Majorana

src/monoprop/pauli.py
```diff
@@ -12,148 +12,181 @@
...
+        num_qubits: int,
+        threshold: float = 1e-12,
```
SimoneGasperini: Same as above for Majorana

src/monoprop/pauli.py
SimoneGasperini: Definitely needs polishing in my opinion

src/monoprop/pauli_data.py
```diff
@@ -162,239 +195,35 @@ def isclose(self, other: object, rtol: float = 1e-05, atol: float = 1e-8) -> boo
...
             )
-            % 2
-        ):
-            return False
-    return True
+        majoranas: list[Sequence[int]] = []
+        coefficients: list[complex] = []
+        for pauli, coeff in self.terms.items():
+            extended = _extend_pauli_string(pauli.string, pauli.qubits, self.num_qubits)
+            majorana, jw_coeff = _pauli_to_fermi(extended)
```
SimoneGasperini: Why the function here is called `_pauli_to_fermi` but it returns Majorana? I'm confused 🤔

src/monoprop/pauli_propagator.py
```diff
@@ -0,0 +1,143 @@
...
+    from collections.abc import Sequence
+
+    import numpy as np
+    from mpi4py import MPI
+
+    from .circuit import Circuit, Exp
+    from .pauli_data import PauliOperator
+
+
+class PauliPropagator(MajoranaPropagator):
```
SimoneGasperini: Wait what? Why is `PauliPropagator` inheriting from `MajoranaPropagator`?? This doesn't make any sense to me

SimoneGasperini: At this point, I believe the best design choice would be to have a `MonomialPropagator` abstract class and both `PauliPropagator` and `MajoranaPropagator` inheriting from it and implementing their own custom stuff

src/monoprop/pauli_propagator.py
```diff
@@ -0,0 +1,143 @@
...
+    available on this class.
+    """
+
+    def __init__(
+        self,
+        initial_operator: PauliOperator,
+        initial_state: list[int] | np.ndarray,
+        *,
+        cutoff: int,
+        schrodinger_cutoff: int | None = None,
```
SimoneGasperini, line 56: Same as above

src/monoprop/pauli_propagator.py
```diff
@@ -0,0 +1,143 @@
...
+        lower_atol: None | float = None,
+        upper_atol: None | float = None,
+        comm: MPI.Comm | None = None,
```
SimoneGasperini, line 59: Same as above

src/monoprop/quantum_data.py
SimoneGasperini, line 1: Probably I'm missing something... @ludmilaasb Can you please explain to me once for all how and why `Protocol` is used here?

README.md
```diff
@@ -30,22 +30,36 @@ runs, or to build the C++ library and executables, build from source (see below)
...
-from monoprop.fermi_data import FermiCircuit, MajoranaOperator, FermiEvGate
+from monoprop import MajoranaPropagator, Exp, Circuit, MajoranaOperator
 
-# Observable m_0 m_1 m_2 m_4, evolved under one Majorana rotation generated by i*m_4 m_5
-observable = MajoranaOperator([(0, 1, 2, 4)], [1.0], 8)
-generator = MajoranaOperator([(4, 5)], [1j], 8)
-circuit = FermiCircuit(initial_state=[], gates=[FermiEvGate(generator, 0.5)])
+# Observable m_0 m_1 m_2 m_4, evolved under one Majorana rotation exp(-i θ/2 · M_γ),
+# generated by M_γ = i*m_4 m_5.
+observable = MajoranaOperator({(0, 1, 2, 4): 1.0}, 8)
```
SimoneGasperini: Please put all the arg names here: for example `num_modes=8`. It makes much easier to understand what's going on directly from the code

SimoneGasperini: Especially because this is likely the very first thing users will see since it's in the README

README.md
```diff
@@ -30,22 +30,36 @@ runs, or to build the C++ library and executables, build from source (see below)
...
+gate = Exp(MajoranaOperator({(4, 5): 1j}, 8))  # Hermitian generator: weight-2 => imaginary coeff
+circuit = Circuit(gates=[gate], parameters=[0.5])  # one angle value per gate
 
-mp = MonomialPropagator(observable, circuit, cutoff=16)
-mp.propagate(evolve_with_coeffs=True)
-print(mp.evolved_operator_dict())  # the gate splits the monomial into two terms
+mp = MajoranaPropagator.from_circuit(circuit, observable, cutoff=16)
```
SimoneGasperini, line 45: What about the `initial_state`? I probably got lost in the conversation... Is the final decision to completely remove it from both the `Circuit` and `Propagator` interface?

README.md
```diff
@@ -30,22 +30,36 @@ runs, or to build the C++ library and executables, build from source (see below)
...
+print(mp.evolved_operator())  # the gate splits the monomial into two terms
 ```
 
-The same simulator also accepts qubit (Pauli) and fermionic operators. See the
-[getting-started guide](https://docs.algorithmiq.fi/monoprop/getting-started.html)
-for more.
+The same simulator also accepts qubit (Pauli) operators via `PauliPropagator`, which
+maps Paulis into the Majorana basis for you. Here we back-propagate `Z ⊗ Z` through one
+`exp(-i θ/2 · X_0)` rotation:
```
SimoneGasperini: > which maps Paulis into the Majorana basis for you

This is not true anymore, right? The `PauliPropagator` will work natively in the Pauli basis

README.md
```diff
@@ -30,22 +30,36 @@ runs, or to build the C++ library and executables, build from source (see below)
...
+
+```python
+from monoprop import PauliPropagator, Exp, Circuit, PauliOperator, Pauli
+
+observable = PauliOperator({"ZZ": 1.0}, num_qubits=2)  # num_qubits lives on the observable
+gate = Exp(PauliOperator({Pauli("X", 0): 1.0}, num_qubits=2))  # exp(-i θ/2 · X_0)
```
SimoneGasperini: Why do you use the `Pauli` term here, whereas in the previous example you don't use the `Majorana` term at all? Perhaps this suggests that the `Pauli` and `Majorana` classes are superfluous

SimoneGasperini: Ideally we would like the two interfaces (for `Pauli` and `Majorana`) to be symmetric and the example should showcase this as well

SimoneGasperini (commented):
I focused exclusively on the Python interface, which I believe is the most important thing right now. I did my best but it's very likely I missed something here and there 😫

src/monoprop/majorana.py
ludmilaasb: Ok, if this is expected to be used by a factory, than it makes sense indeed. 

Simone, factory is a design pattern

src/monoprop/quantum_data.py
ludmilaasb: Instead of declaring an explicitly abstract class (ABC) with a bunch of abstract methods and  lots of code as a result, Python protocl works byu any class that has the methods declared in the speoce matching the methods declared in the protocol scope will be consider as the type defined in the protocol. It is great to interface design patterns. So you have the abstraction with less code, and any class matching the protocol can be used all right via Liskov substituition (they behave as the same). 

TLDR: Python protocols works as some sort of Inplicit inheritance, and we can use it to satisfy Liskov substitution principle (the "L" in SO**L**ID).

src/monoprop/majorana.py
SimoneGasperini: Or we just remove completely the `Majorana`/`Pauli` term (see below) and we build directly a `MajoranaOperator` with `terms: Mapping[Sequence[int], complex]` and `num_modes: int`

src/monoprop/circuit.py
```diff
@@ -0,0 +1,645 @@
...
+
+#: The family a gate's generator belongs to, inferred from its generator type. A
+#: :class:`~monoprop.fermi_data.FermiOperator` generator is converted to its Majorana form in
+#: :meth:`Exp.__init__`, so it becomes a ``"majorana"`` gate -- there is no ``"fermi"`` family.
+GateFamily = Literal["pauli", "majorana"]
+#: The family of a circuit; ``"empty"`` when it has no gates.
+CircuitFamily = Literal["pauli", "majorana", "empty"]
+
+
+class Exp:
```
SimoneGasperini: Why don't we call this `Gate`?

src/monoprop/circuit.py
```diff
@@ -0,0 +1,645 @@
...
+    """
+
+    __slots__ = ("_structural", "family", "generator", "param")
+
+    def __init__(
+        self,
+        generator: MajoranaOperator | PauliOperator | FermiOperator,
+        param: int | None = None,
+        *,
+        _structural: bool = False,
```
SimoneGasperini, line 118: Not sure what is the purpose of `_structural` here 🤔

src/monoprop/circuit.py
```diff
@@ -0,0 +1,645 @@
...
+        family: The generator family -- ``"pauli"`` or ``"majorana"`` -- inferred from the
+            generator type at construction (a fermionic generator becomes ``"majorana"``).
+    """
+
+    __slots__ = ("_structural", "family", "generator", "param")
+
+    def __init__(
+        self,
+        generator: MajoranaOperator | PauliOperator | FermiOperator,
+        param: int | None = None,
```
SimoneGasperini: Maybe rename `param` to something like `index`?

src/monoprop/__init__.py
diagonal-hamiltonian: I think the errors and iquantumoperator should go, but you need pauli and operator to input to the circuit

diagonal-hamiltonian: *pauli operator and majorana operator

src/monoprop/majorana.py
diagonal-hamiltonian: I think its nice to do this as shorthand as we check internally if the tuple is consitent

src/monoprop/majorana_propagator.py
diagonal-hamiltonian: Its needed as the schrodinger cutoff determines the order of expansion approximation for your initial state

src/monoprop/majorana_propagator.py
diagonal-hamiltonian: I think the type hint helps, I'm not against it but maybe its convention.. @robertodr ?

src/monoprop/pauli.py
diagonal-hamiltonian: same comment as above

src/monoprop/pauli_data.py
diagonal-hamiltonian: good point.. technically fermionic operators encompass majorana operators and creation and anihillation operators but i see the confusion

README.md
diagonal-hamiltonian: kk

README.md
diagonal-hamiltonian: its in the circuit

README.md
diagonal-hamiltonian: yeah the only way to do that in a coherent way is to pass binary strings instead of XYZ's for pauli but thats a can of worms

src/monoprop/majorana_propagator.py
robertodr: it's indeed conventional to call it `comm`.

robertodr: It's one of those things where I advise against spelling it out in full.

src/monoprop/circuit.py
diagonal-hamiltonian: because Gate[Pauli|Majorana] != Gate[Exp[Pauli|Majorana]] for example: Gate[XXX] could be interpreted as an X gate on 3 qubits. This is not a pauli gadget rotation .

src/monoprop/circuit.py
diagonal-hamiltonian: index is good. Thanks!

src/monoprop/pauli_propagator.py
diagonal-hamiltonian: Done — introduced an abstract `MonomialPropagator` base (`src/monoprop/monomial_propagator.py`) that owns all the shared graph-building, evaluation, and introspection. `MajoranaPropagator` and `PauliPropagator` now both inherit from it directly; `PauliPropagator` no longer subclasses `MajoranaPropagator`. Each concrete class only implements `__init__` + `_circuit_gates`.

src/monoprop/majorana_propagator.py
diagonal-hamiltonian: Good spot — fixed. `basis_change` is now fully internal and automatic (`None` for Majorana, Jordan–Wigner for Pauli). I removed the constructor argument, the setter, and the public property from both propagators, so a native Majorana propagator no longer exposes a basis knob at all.

src/monoprop/majorana_propagator.py
diagonal-hamiltonian: Keeping `comm` per @robertodr — it's the conventional name and the type hint documents that it's an MPI communicator.

src/monoprop/majorana.py
diagonal-hamiltonian: Done — renamed the module to `majorana.py`.

src/monoprop/pauli.py
diagonal-hamiltonian: Done — renamed to `pauli.py` (and `fermi_data.py` → `fermi.py` for symmetry), with all imports updated across the package, tests, docs, and benches.

src/monoprop/majorana.py
diagonal-hamiltonian: Kept the `Mapping[Majorana | Sequence[int], complex]` shorthand — a raw index tuple is routed through `Majorana(*key)`, so it gets exactly the same non-negative/distinct validation a `Majorana` key carries. Since it's validated internally I'd rather keep the convenience.

src/monoprop/pauli.py
diagonal-hamiltonian: Same as the `MajoranaOperator` above — keeping the `Pauli`-or-`str` shorthand; it's validated internally.

src/monoprop/majorana.py
diagonal-hamiltonian: It's the coefficient-drop threshold (default `1e-12`): terms with `|coefficient| < threshold` are discarded when the operator is built. Documented in the constructor docstring.

src/monoprop/pauli_data.py
diagonal-hamiltonian: Good point — renamed to `_pauli_to_majorana` (it returns Majorana indices + a phase, not a fermionic operator) and added a one-line docstring saying so.

src/monoprop/circuit.py
diagonal-hamiltonian: Done — renamed `Exp.param` → `Exp.index` (attribute, constructor arg, repr, and `_with_param` → `_with_index`).

src/monoprop/circuit.py
diagonal-hamiltonian: `_structural` is an internal flag set only by the wire/dense path (`Circuit.from_dense_arrays`): those generators already carry the real structural coefficients `g`, so `_gate_layers` passes them through instead of antihermitian-normalizing them. It's keyword-only and not part of the authoring API.

src/monoprop/__init__.py
diagonal-hamiltonian: Trimmed the public surface: `IQuantumOperator` and `NumberOfModesInvalidError` are no longer exported from `monoprop`. Kept `PauliOperator`/`MajoranaOperator` (the circuit inputs) and the `Fermi*` types — fermionic operators stay as an authoring/input format (and the output of `integrals_to_fermion`), mapped to Majorana internally via `get_majorana_operator()`.

src/monoprop/__init__.py
diagonal-hamiltonian: This is the lazy-export machinery that keeps qiskit an optional dependency: `import monoprop` never imports qiskit, but `from monoprop import to_qiskit_circuit` still works (via a PEP 562 module `__getattr__`), and `import *` stays qiskit-free for installs without the extra. Kept it; the module docstring explains the intent.

src/monoprop/fermi.py
diagonal-hamiltonian: Keeping `Fermi*` as a supported input format — it's the natural authoring form for chemistry/lattice models and the output of `integrals_to_fermion`, converted to Majorana internally. So it isn't redundant with `MajoranaOperator`; removing it would cost the fermionic entry point.

README.md
diagonal-hamiltonian: Done — the quick-example now passes `num_modes=8` / `num_qubits=2` as keywords consistently.

README.md
diagonal-hamiltonian: Removed — the README no longer describes how Paulis are handled internally (basis mapping). It just states that `PauliPropagator` takes qubit (Pauli) operators and circuits.

README.md
diagonal-hamiltonian: Agreed this is worth improving. Full Pauli/Majorana authoring symmetry (e.g. binary-string Pauli input) is a bigger change, so I've left it for a follow-up; for now the two examples at least use consistent keyword args.

src/monoprop/pauli.py
diagonal-hamiltonian: Agreed — leaving the move of `_extend_pauli_string` onto the `Pauli` class for a follow-up PR, as discussed.

src/monoprop/__init__.py
SimoneGasperini: I'm ok with lazy imports but why `from_qiskit_circuit as from_qiskit_circuit`?? Claude went crazy there

src/monoprop/majorana.py
SimoneGasperini: I'm not a fan because it's an object immediately discarding himself at construction. Is this required for any specific reason? In general I think only the Pauli/Majorana propagator should have the responsibility of discarding terms

SimoneGasperini: What happens when `lower_atol=None` at the propagator object level? Are terms with `|coefficient| < threshold` discarded anyway? That shouldn't happen!

diagonal-hamiltonian: Yeah maybe there should be no truncation in the majorana operator objects

SimoneGasperini: Yes that's the point, and same for Pauli operators

diagonal-hamiltonian: Great, could you remove the pruning from the Operator objects then please? Sorry not at home

SimoneGasperini: Yes will do, no worries

SimoneGasperini (dismissed):

matteoacrossi (approved):

PR #43, opened 2026-07-03
build: exclude gcc < 14 and clang < 18

### LINKED ISSUE #11: Document supported compilers
We work exclusively with GCC, but other modern compilers for Unix should be working as well. Which one works and which version is the minimal requested has to be documented.

I did some tests in the (distant) past with recent versions of icpx (Intel's newest c++ compiler) on Leonardo and things worked smoothly.

### LINKED ISSUE #12: C++ standard in use
We fixed the c++ standard to c++23 in a PR some time back. That was decided because:

1. We wanted to take advantage of the fully `constexpr` version of `std::bitset`, and
2. I had an idea of splitting Heisenberg/Schrodinger simulation modes using the curiously recurring template pattern, and in C++23 the syntax for it becomes much more compact.

However:

1. To work with MPI, @diagonal-hamiltonian had to implement a custom bitset. So we're not using `std::bitset` any longer.
2. The refactoring mentioned above hasn't happened.

So, we could revert to C++20 which would:

* Expand the list of supported compilers Algorithmiq/monoprop#11 
* Let us decrease the macOS deployment target, giving wider support for its binary wheel.

robertodr:
- [X] Test minimum supported GCC and Clang versions on Linux
- [x] Exclude less-than-minimum versions at configuration-time
- [x] close #11 
- [x] close #12

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -343,7 +343,7 @@ jobs:
             cmake_cxx_compiler: clang++
           - runner: macos-15
             compiler: apple-clang
-            cmake_cxx_compiler: g++
+            cmake_cxx_compiler: clang++
 
     steps:
       - uses: actions/checkout@v7
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -365,6 +365,10 @@ jobs:
       - name: Set clang-18/clang++-18 as default compilers
         if: ${{ matrix.compiler == 'clang-18' }}
         run: |
+          if ! command -v clang-18 >/dev/null; then
+            sudo apt-get update
+            sudo apt-get install -y clang-18
+          fi
           sudo update-alternatives --install /usr/bin/clang clang /usr/bin/clang-18 180
           sudo update-alternatives --set clang /usr/bin/clang-18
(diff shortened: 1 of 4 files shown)


matteoacrossi (dismissed):

robertodr:
@copilot 
* modify the cpp-checks-other job in the test.yml workflow to a matrix of jobs testing the oldest supported versions of the c++ compilers only for the ubuntu runners: GCC 14 and Clang 18.
* add ubuntu-latest as runner to cpp-checks-other only for Clang 18

robertodr:
This PR will have to be merged bypassing the requirements. I've changed the labels of the jobs, so the required ones are no longer being run. Once merged, the requirements need to be updated as well in the ruleset.

robertodr:
@matteoacrossi I'm not touching the documentation to report the minimum supported compiler versions, as it would create conflicts with #39. Once that's in, it's not difficult to do and then we could close #11 (and #12)

matteoacrossi (dismissed):

robertodr:
@copilot amend the documentation to:
* explicitly state that we require C++23 compliant compilers, and
* State which versions of GCC and clang are the minimum required on linux

matteoacrossi (approved):

PR #48, opened 2026-07-08
chore(prek): update and run pre-commit hooks

robertodr:
Update pre-commit hooks and run them on whole codebase.


robertodr:
https://github.com/scanaislop/aislop/blob/main/.pre-commit-hooks.yaml I don't think it can work in our lint workflow, since it hard codes the --staged flag 🫠

robertodr:
I'm disappointed to be honest. The precommit hook does not work when running in ci and the action does not add any message/comment to the PR. It seems of very limited usefulness

robertodr:
I'll keep the good parts (hooks updates, cmake reformatting) and toss aislop away.

diagonal-hamiltonian (approved):

PR #51, opened 2026-07-09
ci(deploy): fix definition of test group

robertodr:
psutil must be in the test, not the bench group


SimoneGasperini (dismissed):
Yes, thank you! 👍🏽

PR #54, opened 2026-07-10
ci(deploy): :rocket: migrate publish job to PyPI Trusted Publishing

matteoacrossi:
## Why
We want to publish directly to PyPI using the recommended, tokenless Trusted Publishing flow.

## What changed
- Updated `.github/workflows/deploy.yml` to publish to PyPI via `pypa/gh-action-pypi-publish@release/v1`
- Switched the publish job to GitHub OIDC with `environment: pypi` and `id-token: write`
- Added artifact provenance attestation generation via `actions/attest-build-provenance@v4`
- Removed AWS CodeArtifact credential/configuration steps and legacy Twine upload commands
- Removed the manual `force-upload` input and limited publishing to `release.published`
- Renamed workflow/job labels to reflect direct PyPI publishing

## Notes for rollout
- The `pypi` GitHub environment and PyPI Trusted Publisher mapping must be configured to allow this workflow to publish.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,13 +1,7 @@
-name: Publish monoprop Python sdist and wheels to Algorithmiq PyPI
+name: Publish monoprop Python sdist and wheels to PyPI
 
 on:
   workflow_dispatch:
-    inputs:
-      force-upload:
-        description: 'Upload artifacts without a corresponding release'
-        required: false
-        default: false
-        type: boolean
   release:
     types:
       - published
@@ -162,17 +156,18 @@ jobs:
           name: cibw-wheels-${{ matrix.os }}-${{ strategy.job-index }}
           path: wheelhouse/*.whl
 
-  upload_codeartifact:
-    name: Upload to AWS CodeArtifact
+  upload_pypi:
+    name: Upload to PyPI
     needs:
       - build-sdist
       - build-wheels
-    if: (github.event_name == 'release' && github.event.action == 'published') || github.event.inputs.force-upload
+    if: github.event_name == 'release' && github.event.action == 'published'
 
     runs-on: ubuntu-latest
-
+    environment: pypi
     permissions:
       id-token: write
+      attestations: write
       contents: read
 
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -162,17 +156,18 @@ jobs:
...
 
-  upload_codeartifact:
-    name: Upload to AWS CodeArtifact
+  upload_pypi:
+    name: Upload to PyPI
     needs:
       - build-sdist
       - build-wheels
-    if: (github.event_name == 'release' && github.event.action == 'published') || github.event.inputs.force-upload
+    if: github.event_name == 'release' && github.event.action == 'published'
```
matteoacrossi, line 164: @robertodr the bot removed force-upload, do we need it?

robertodr: it's not something that we can do on PyPI, so it's fine to remove.

This workflow has a `workflow_dispatch` trigger and I wanted to avoid that on every such trigger, we get a new artifact uploaded. Hence the option to the trigger to force an upload. We used it (only a few times) to upload pre-alpha versions, but since we're going open source, I think the proper way is to have alphas on the public PyPI anyway.

So basically, it's all good. We can keep the `workflow_dispatch` (without `force-upload` option) such that we can force a run of this workflow in case we want to test something.

.github/workflows/deploy.yml
```diff
@@ -162,17 +156,18 @@ jobs:
...
 
     runs-on: ubuntu-latest
-
+    environment: pypi
```
robertodr, line 167: both github and PyPA suggest to have this line, but I haven't understood what it does and if we need to do something extra GH-side https://docs.github.com/en/actions/how-tos/deploy/configure-and-manage-deployments/manage-environments

If we have the environment, the trusted publisher needs to be modified in the PyPI settings.

.github/workflows/deploy.yml
matteoacrossi: Perfetto

robertodr (approved):
🎡 📦 ⬆️

PR #79, opened 2026-07-15
chore: Majorana Propagation Benchmarks

ludmilaasb:
This PR adds benchmark scripts for 1D Hubbard model for MajoranaPropagation.jl and monoprop.


diagonal-hamiltonian:
Please add a number of terms comparison as well

diagonal-hamiltonian:
also please check the benchmarks as i already have a 1d hubbard model

ludmilaasb:
> Please add a number of terms comparison as well

It was included already, [here](https://github.com/Algorithmiq/monoprop/blob/4e7271259ded66ffa2c2fa5345c944b3238acba0/benches/third_party/majorana_prop/num_terms_vs_layers.png).

matteoacrossi:
I would like both benchmarks to be discussed in the documentation (under benchmarks). The plots (showing performance improvements wrt the other packages should be highlighted).

ludmilaasb:
> I would like both benchmarks to be discussed in the documentation (under benchmarks). The plots (showing performance improvements wrt the other packages should be highlighted).

Done, please review it.

benches/third_party/majorana_prop/julia_hubbard1d_benchmark_results.jsonl
matteoacrossi, line 1: Not sure if we should commit these files. @robertodr @fpietra opinions?

docs/content/docs/benchmarks.mdx
matteoacrossi, line 1: I think this page should start with the benchmarks against third party. Ideally, with the plots as high up as possible, with short explainer (this is a comaprison with a fermi hubbard model etc etc). Details on how to reproduce the plots and specs can come later, and the internal benches as third.

benches/third_party/majorana_prop/plot_results.py
matteoacrossi, line 1: I propose a single image with multiple panels, so it looks nicer and more compact in the docs.

matteoacrossi (changes requested):
This is looking good, but I propose some changes mostly on the presentation of the results.

matteoacrossi:
Also, I'd decrease the number of points in the plots to speed things up.

benches/third_party/majorana_prop/plot_results.py
ludmilaasb: Done

docs/content/docs/benchmarks.mdx
ludmilaasb: Done.

benches/third_party/majorana_prop/plot_results.py
ludmilaasb: I also took the liberty to change the script and picture for Pauli Propagation benchmark.

matteoacrossi:
Looking better now. I'd still decrease the points by running every other layer (the curves are quite clear anyway).

In the Benchmarks page, it's good that the plots are at the top, but there should be a couple of lines explaining briefly what they show (with links to the detailed sections below).

diagonal-hamiltonian (changes requested):
Good work! Could you refactor the models into some sort of /builder folder by any chance and look for places in docs/tutorials where these could be reused please?

ludmilaasb:
> Looking better now. I'd still decrease the points by running every other layer (the curves are quite clear anyway).
> 
> In the Benchmarks page, it's good that the plots are at the top, but there should be a couple of lines explaining briefly what they show (with links to the detailed sections below).

Done and done.

ludmilaasb:
>Good work! Could you refactor the models into some sort of /builder folder by any chance and look for places in docs/tutorials where these could be reused please?


I understand you want to avoid duplication regarding the hubbard models usage in scripts. However, concerning the benchmarks and tutorials I believe _they should be self contained_, meaning, I would prefer run any tutorial/benchmark scrips without the need of cloning the entire repository.

IMHO, centralizing the information is not really a blocker for having the benchmarks PRs merged, if centralizing the information is needed, could we address this in a next PR? 

@matteoacrossi , what do you think?

matteoacrossi (approved):
Please merge the conflicts with the base PR (I think @SimoneGasperini updated his results) then merge.

PR #96, opened 2026-07-17
feat: :sparkles: `to_qiskit_operator` returns now local gates

### LINKED ISSUE #71: nitpick: in `to_qiskit_circuit`
in `to_qiskit_circuit`, gates are written as PauliEvolutionGate (fine for me) but they are extended with identity operators. It would be great if the gate would local in qubits. again (especially that it seems that the code might be simpler)

Visual example:
```
from __future__ import annotations

from qiskit import QuantumCircuit
from qiskit.circuit.library import PauliEvolutionGate
from qiskit.quantum_info import Pauli

from monoprop.qiskit_conversion import from_qiskit_circuit, to_qiskit_circuit

qcirc = QuantumCircuit(3)
qcirc.rz(3.14, 0)
print("INPUT")
print(qcirc.draw())

circ = to_qiskit_circuit(from_qiskit_circuit(qcirc, [0, 2]), 3)
print("ROUNDTRIP RESULT")
print(circ.draw())

print("BETTER")
qcirc = QuantumCircuit(5)
qcirc.append(PauliEvolutionGate(Pauli("Z"), 3.14), [0])
print(qcirc.draw())
```
output:
```
INPUT
     ┌──────────┐
q_0: ┤ Rz(3.14) ├
     └──────────┘
q_1: ────────────
                 
q_2: ────────────
                 
ROUNDTRIP RESULT
     ┌─────────────────────┐
q_0: ┤0                    ├
     │                     │
q_1: ┤1 exp(-it IIZ)(1.57) ├
     │                     │
q_2: ┤2                    ├
     └─────────────────────┘
BETTER
     ┌──────────────────┐
q_0: ┤ exp(-it Z)(3.14) ├
     └──────────────────┘
q_1: ────────────────────
                         
q_2: ────────────────────
                         
```

adamglos92:
- [x] updated `to_qiskit_operator` so that the resulting gates are no longer full-system
- [x] added test that verifies the behaviour for `to_qiskit_operator`
- [x] added roundtrip test that verifies applying to_ and from_qiskit_operator in sequence results in the same circuit.

Closes #71


src/monoprop/qiskit_conversion.py
```diff
@@ -165,6 +164,21 @@ def from_qiskit_circuit(
     )
 
 
+def _extend_generator_minimally(
+    generator: PauliOperator,
+) -> tuple[dict[str, float], list[int]]:
+    qubits = sorted(set.union(*(set(p.qubits) for p in generator.terms)))
```
ludmilaasb: question: So for each term in the generator, they can have different qubits so that's why the union? And then sorted is inplicitly converting to list. Is that correct?

src/monoprop/qiskit_conversion.py
```diff
@@ -165,6 +164,21 @@ def from_qiskit_circuit(
...
+    localizing_qubit_map = {q: i for i, q in enumerate(qubits)}
+    result = {}
+    for pauli, coeff in generator.terms.items():
+        default_term = ["I"] * len(qubits)
+        for p, q in zip(pauli.string, pauli.qubits):
+            default_term[int(localizing_qubit_map[q])] = p
```
ludmilaasb: question: Do you need to cast the value to int, why? I though it was int by default by the way you constructed on line 171.

src/monoprop/qiskit_conversion.py
```diff
@@ -165,6 +164,21 @@ def from_qiskit_circuit(
...
+        result["".join(default_term)] = coeff
```
ludmilaasb: suggestion: This is basically duplicating `_extend_pauli_string` but for a smaller set of qubits. It is the same loop. Maybe you could just call it?

tests/test_qiskit_conversion.py
```diff
@@ -200,6 +200,28 @@ def test_single_gate(self):
...
+            initial_state=(),
+        )
+
+        result = to_qiskit_circuit(circuit, 5)
+
+        assert isinstance(result, QuantumCircuit)
+
+        qreg = result.qregs[0]
+        assert result.num_qubits == 5
+        assert len(result.data) == 1
```
ludmilaasb: What line 214 has to do with the following assertions? They seem uncorrelated.

tests/test_qiskit_conversion.py
```diff
@@ -200,6 +200,28 @@ def test_single_gate(self):
...
+        gate = result.data[0]
+        assert gate.operation.name == "PauliEvolution"
+        indices = tuple(qreg.index(qb) for qb in gate.qubits)
+        assert indices == (1, 2, 3)  # sorted order of qubits
+        assert gate.operation.time == 0.7
+        # reversed because of qiskit ordering
+        assert gate.operation.operator == SparsePauliOp.from_list([("ZYX", 1.5)])
```
ludmilaasb: question: Wouldn't be easier to just pass the expected qiskit gate and assert from that? I think the qiskit circuit may have some equality?

ludmilaasb (commented):
Thank you @adamglos92 for this PR!

I think it is ok, just left some nit-pick comments. 

Good job! 🚀

src/monoprop/qiskit_conversion.py
adamglos92: * they eventually need to be placed in a single PauliEvolutionGate. You can interpret it as putting identities to get `exp(it(XYI + ZZZ))` 
* Converting to list is fine as PauliEvolutionGate is expecting a sequence

src/monoprop/qiskit_conversion.py
adamglos92: you might be right actually. Seems like to eager AI autocompletion I missed 😓

tests/test_qiskit_conversion.py
adamglos92: interesting idea let me try

src/monoprop/qiskit_conversion.py
adamglos92: done

src/monoprop/qiskit_conversion.py
adamglos92: done

tests/test_qiskit_conversion.py
adamglos92: indeed! a leftover of when it made sense. done

tests/test_qiskit_conversion.py
adamglos92: done, ended up in a bit larger refactoring

src/monoprop/qiskit_conversion.py
ludmilaasb: I see. I though that for a PauliOperator they had to act in the same qubits by default, so that's why it felt redundant checking for all the terms.

tests/test_qiskit_conversion.py
ludmilaasb: Yes, but you add more tests cases, right? higher chance to catch a bug!

src/monoprop/qiskit_conversion.py
adamglos92: They had, but sometimes with identity :)

tests/test_qiskit_conversion.py
adamglos92: that was not my plan 😂 however I do think current cases are already decently catches the behaviour

adamglos92: Any more cases feels excessive for this PR

ludmilaasb (dismissed):

ludmilaasb (approved):

PR #97, opened 2026-07-17
refactor: :fire: is_identity functionality removed

### LINKED ISSUE #67: :bug: confusing `is_identity` name for Operators
Right now Majorana and Pauli operators have `is_identity`, which based on the name alone intuitively means I should expect np.eye operator (in the algebraic sense). However based on the implementation, it is verifying if the operator is 0 operator.
* either change the name to e.g. `is_zero` (probably what you want as I guess you are using it somehow)
* or change the implementation so that for all terms it is 0 except of the core term (empty tuple for Majorana and "I"*num_qubits for Pauli) where it should be 1

adamglos92:
- [x] removed `is_identity` methods from operators
- [x] removed the relevant test
- [x] removed usage of `is_identity` in `test_majorana_operator_normalizes_terms` (which was IMHO not testing antyhing substantial given assertion on `op.terms`

Closes #67

### DIFF
--- a/src/monoprop/fermi.py
+++ b/src/monoprop/fermi.py
@@ -155,10 +155,6 @@ class FermiOperator:
         out += ")"
         return out
 
-    def is_identity(self) -> bool:
-        """Check if the operator is the identity."""
-        return all(coef == 0 for coef in self.coefficients)
-
     def _as_dict(self) -> dict[tuple, complex]:
         """Return the operator as a dictionary."""
         result = {}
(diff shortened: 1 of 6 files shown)


ludmilaasb (dismissed):
LGTM!

adamglos92:
sorry @ludmilaasb wrong branch need to reverse

adamglos92:
As good as new

ludmilaasb (approved):

PR #106, opened 2026-07-20
build: only allow to build as python extension

robertodr:
This project is purely a Python extension but the build system, and configuration around it, still offered the (vestigial) option of building as a C++-only project. I now gate the compilation of the C++ unit tests behind a new CMake option (`monoprop_ENABLE_CXX_UNIT_TESTS`) and warn against compiling with direct invocation to CMake.


robertodr:
I mucked things up while using gitbutler to handle this PR... Closing and re-opening with the correct changeset.

PR #108, opened 2026-07-20
refactor: better schrodinger evolution interface

### LINKED ISSUE #104: Bug when using Schrodinger picture with PualiPropagation
There is potentially an issue with terms being truncated over-aggresively when in the Schrodinger picture and using Pauli operators.

In the following MWE, I set up an n= 6 qubit example with an empty circuit. 
The initial state is the all-zeros ket-bra (i.e. `initial_state=[]`). This state contains _all_ strings consisting of only I and Z.

The cutoff is chosen to be c = 4 < n, c= 6 = n, c = 12 = 2*n.

The issue is immediately visible without even propagating.

With c =4, only weight 0, 1 and 2 operators are kept. 
With c = 6, only weight <= 3 operators are kept.
With c = 12, we recover all operators.

I think this surfaces 2 issues:

(1) Strings consisting of only I/Z (equivalent to paired monomials) should never be truncated. This is resulting in immediate loss of signal :( .

(2) According to the documentation:
> Qubit operators: Pauli weight
PauliPropagator accepts Pauli operators and gates directly and measures each term by its Pauli weight — the number of qubits it touches. This is the only structural measure: cutoff bounds the Pauli weight.

In actual fact, it seems like cutoff is being directly applied at the Majorana level i.e. c = 6 leads to only terms with 3 Z-operators === 6 Majoranas. This can obviously be worked around by doubling c but potentially warrants a change in documentation.

```
import numpy as np
from monoprop import Circuit, ExpGate, Pauli, PauliOperator, PauliPropagator

num_qubits = 6

circuit = Circuit([])
parameters = [np.pi/8]

def majorana_to_pauli(indices, n):
    """(phase, pauli_string) such that the Majorana monomial = phase * pauli_string."""
    mult = {
        ("I", "I"): (1, "I"), ("I", "X"): (1, "X"), ("I", "Y"): (1, "Y"), ("I", "Z"): (1, "Z"),
        ("X", "I"): (1, "X"), ("X", "X"): (1, "I"), ("X", "Y"): (1j, "Z"), ("X", "Z"): (-1j, "Y"),
        ("Y", "I"): (1, "Y"), ("Y", "X"): (-1j, "Z"), ("Y", "Y"): (1, "I"), ("Y", "Z"): (1j, "X"),
        ("Z", "I"): (1, "Z"), ("Z", "X"): (1j, "Y"), ("Z", "Y"): (-1j, "X"), ("Z", "Z"): (1, "I"),
    }
    result = ["I"] * n
    if indices == ():
        return 1 + 0j, "".join(result)
    phase = 1 + 0j
    for idx in indices:
        j, letter = idx // 2, ("X" if idx % 2 == 0 else "Y")
        term = ["Z"] * j + [letter] + ["I"] * (n - j - 1)
        for q in range(n):
            p, c = mult[(result[q], term[q])]
            phase *= p
            result[q] = c
    return phase, "".join(result)


def print_weights(operator):
    weight_counts = {}
    example_term = {}
    for k, v in operator.items():
        phase, op = majorana_to_pauli(k, num_qubits)
        if not all(c in ("I", "Z") for c in op):
            continue  # discard the non-diagonal (unpaired) terms
        weight = sum(c == "Z" for c in op)
        weight_counts[weight] = weight_counts.get(weight, 0) + 1
        example_term.setdefault(weight, (op, np.real(phase * v)))

    print(f"cutoff={cutoff}: paired-term weight counts = {dict(sorted(weight_counts.items()))}")
    for weight in sorted(example_term):
        print(f"   e.g. weight {weight}: {example_term[weight]}")


for cutoff in (4, num_qubits, 2*num_qubits):
    propagator = PauliPropagator(
        PauliOperator({Pauli("I", 0): 1}, num_qubits),
        initial_state=[],
        cutoff=cutoff,
        schrodinger_cutoff=cutoff,
    )
    unevolved = propagator.evolved_operator(parameters)
    propagator.build_graph(circuit)
    evolved = propagator.evolved_operator(parameters)

    print('Unevolved')
    print_weights(unevolved)

    print()
```

adamglos92:
- [x] Schrodinger cutoff interpreted differently for PauliPropagator
- [x] docs and docstrings improved accordingly
- [x] test added to confirm the effect


JoshCudbyAlgo (dismissed):
lgtm 👍

src/monoprop/majorana_propagator.py
```diff
@@ -82,8 +82,10 @@ def __init__(
                 to an expectation value against a computational-basis state or Slater
                 determinant; discarding them would throw away signal.
             schrodinger_cutoff: Optional cutoff for Schrodinger-picture evolution. If
-                provided, enables the Schrodinger picture; if ``None``, the Heisenberg
-                picture is used.
+                provided, enables the Schrodinger picture, starting in a n initial state
```
diagonal-hamiltonian: The english here is a bit off. AI slop will fix

diagonal-hamiltonian (changes requested):
Looks good

src/monoprop/majorana_propagator.py
adamglos92: @diagonal-hamiltonian please check

ludmilaasb: Is this solved?

adamglos92: AI improved IMO, but @diagonal-hamiltonian ?

robertodr (approved):

diagonal-hamiltonian (approved):

PR #115, opened 2026-07-20
refactor: :recycle: improve error message when exceeding max number of modes

### LINKED ISSUE #84: Inform user what if (s)he wants more modes
My undersatnding is that the dispatch returns various numbers of modes so that you can have code optimized for a requested ones. 250 sounds like something large, but Monoprop is efficient in little-qubits-long-circuit and many-qubits-short-circuit.

Personally I'm fine with it being capped to the given number of qubits, but the least I would do is to inform user what (s)he can do if he needs more qubits/modes. Possibilities:
* for the package and extend
* write an issue
In particular I would expand the current error message in `dispatch`, currently
```
        case 250:
            cls = MonomialPropagator250
        case n if n > 250:
            errmsg = f"Number of Fermionic modes {n} invalid. num_modes must be <= 250."
            raise NumberOfModesInvalidError(errmsg)

    return cls
```

another thing: this codes looks like something that could be refactored with some e.g. on-the-fly metaprogramming. Would it be possible to do it so that up to 250 qubits the propagators are pre-compiled, and then there would be compiled on the fly?

adamglos92:
I updated the tools scripts, not sure if I need to do anything else

### DIFF
--- a/tools/generate-dispatch.py
+++ b/tools/generate-dispatch.py
@@ -192,7 +192,7 @@ def dispatch(num_modes: int) -> type[_SimulatorAdapter]:
 {cases}
         case n if n > {max_logical_num_modes}:
             errmsg = (
-                f"Number of Fermionic modes {{n}} invalid. num_modes must be <= 250."
+                f"Number of Fermionic modes {{n}} invalid. num_modes must be <= {max_logical_num_modes}."
                 "Contact monoprop developers if more than 250 Fermionic modes are required."
             )
             raise NumberOfModesInvalidError(errmsg)
--- a/tools/generate-dispatch.py
+++ b/tools/generate-dispatch.py
@@ -193,7 +193,7 @@ def dispatch(num_modes: int) -> type[_SimulatorAdapter]:
         case n if n > {max_logical_num_modes}:
             errmsg = (
                 f"Number of Fermionic modes {{n}} invalid. num_modes must be <= {max_logical_num_modes}."
-                "Contact monoprop developers if more than 250 Fermionic modes are required."
+                "Contact monoprop developers if more than {max_logical_num_modes} Fermionic modes are required."
             )
             raise NumberOfModesInvalidError(errmsg)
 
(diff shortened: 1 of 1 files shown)


tools/generate-dispatch.py
```diff
@@ -191,7 +191,10 @@ def dispatch(num_modes: int) -> type[_SimulatorAdapter]:
             raise NumberOfModesInvalidError(errmsg)
 {cases}
         case n if n > {max_logical_num_modes}:
-            errmsg = f"Number of Fermionic modes {{n}} invalid. num_modes must be <= {max_logical_num_modes}."
+            errmsg = (
+                f"Number of Fermionic modes {{n}} invalid. num_modes must be <= 250."
```
robertodr: 250 must be `{max_logical_num_modes}`

adamglos92: 🤦

adamglos92: please check. Again, do I have to run something? Or are tools scripts run automatically

adamglos92:
Closes #84 
I suppose?

tools/generate-dispatch.py
```diff
@@ -191,7 +191,10 @@ def dispatch(num_modes: int) -> type[_SimulatorAdapter]:
             raise NumberOfModesInvalidError(errmsg)
 {cases}
         case n if n > {max_logical_num_modes}:
-            errmsg = f"Number of Fermionic modes {{n}} invalid. num_modes must be <= {max_logical_num_modes}."
+            errmsg = (
+                f"Number of Fermionic modes {{n}} invalid. num_modes must be <= {max_logical_num_modes}."
+                "Contact monoprop developers if more than 250 Fermionic modes are required."
```
robertodr:
```suggestion
                "Contact monoprop developers if more than {max_logical_num_modes} Fermionic modes are required."
```

tools/generate-dispatch.py
robertodr: yes, they're run automatically when building the C++ code.

robertodr (approved):
Thanks @adamglos92

PR #116, opened 2026-07-20
fix: :bug: MajoranaOperator are now being checked for consisting of commuting terms

### LINKED ISSUE #80: non-commuting terms still allowed in the Fermionic space
Currently monoprop disallows having non-commuting Paulis in single ExpGate - which is good because we don't want to be responsible for Trotterization error. However, non-commuting terms are still allowed in Majorana basis.

MWE:
```
from monoprop import ExpGate, MajoranaOperator, Pauli, PauliOperator

# this does not throw error (but should!)
circuit = ExpGate(MajoranaOperator({(0,): 1.0, (1,): 1.0}, num_modes=4))

# this does throw error (correct!)
circuit = ExpGate(PauliOperator({Pauli("X"): 1.0, Pauli("Z"): 1.0}, num_qubits=4))
```

adamglos92:
- [x] Checks in MajoranaOperator if it consists of commuting terms; raises error otherwise
- [x] test with a few succeeding and failing cases
- [x] fix existing tests that were broken with a change (they change was unrelated) 

Closes #80 
Blocked by #114 as required FermiString to truncate negligible terms


src/monoprop/circuit.py
```diff
@@ -158,8 +158,10 @@ def __init__(
 
         self.generator = self._truncated_term(generator, atol)
 
-        if isinstance(generator, PauliOperator):
-            _validate_commuting_pauli_generator(generator)
+        if isinstance(self.generator, PauliOperator):
+            _validate_commuting_pauli_generator(self.generator)
+        elif isinstance(self.generator, MajoranaOperator):
+            _validate_commuting_majorana_generator(self.generator)
```
ludmilaasb, line 164: I think `_validate_commuting`  functions could be moved to their respective classes, so you don't need to check the type to apply to proper function.

ludmilaasb (commented):
Looking good! I have a small suggestion to keep the code a bit more clean but in overall is fine!

ludmilaasb (approved):
I think we can do the refactoring of the validation in a next PR. I am approving it for now.

PR #117, opened 2026-07-20
feat: ✨ Propagators use operators from now

### LINKED ISSUE #4: evolved_operator_dict should return a MajoranaOperator object
title

### LINKED ISSUE #5: Currently evolved_operator_dict returns majoranas even if your simulating paulis or arbitrary basis'
the easiest thing to do is flag it to the user. Decoding to general basis is a can of worms

adamglos92:
- [x] make PauliPropagator return PauliOperator (nitpicks after #33) 
- [x] make MajoranaPropagator return MajoranaOperator in a similar way
- [x] similarly update update_initial_operator
- [x] tests: for PauliPropagator.evolved_operator. Not great, but they verify the particular PauliPropagator responsibility
- [x] other tests updated
- [x] docs corrected

Closes #4
Closes #5


src/monoprop/conversion_utils.py
```diff
@@ -74,6 +74,63 @@ def _pauli_to_majorana(pauli: str) -> tuple[tuple[int, ...], complex]:
...
+    for i in range(n_qubits - 1, -1, -1):
+        has_even = (2 * i) in majorana_set
+        has_odd = (2 * i + 1) in majorana_set
+
+        if not flag_z:
+            if has_even and has_odd:
+                pauli[i] = "Z"
+                forward_coeff *= -1j
+            elif not has_even and not has_odd:
+                pauli[i] = "I"
```
ludmilaasb: Maybe this can be removed? The pauli is already initiated with "I". This check is not doing much

adamglos92: good point, let me think

adamglos92: it cannot be just removed, but it can be refactored

diagonal-hamiltonian (commented):
Amazing adam.

diagonal-hamiltonian (dismissed):

src/monoprop/conversion_utils.py
```diff
@@ -74,6 +74,63 @@ def _pauli_to_majorana(pauli: str) -> tuple[tuple[int, ...], complex]:
...
+        elif not has_even and not has_odd:
+            pauli[i] = "Z"
+        elif has_even:
+            pauli[i] = "Y"
+            flag_z = False
+            forward_coeff *= 1j
+        else:
+            pauli[i] = "X"
+            flag_z = False
+            forward_coeff *= -1j
```
ludmilaasb: Maybe instead of going through conditionals, we could just do some simpler math? Does it make sense if instead checking has_even and has_odd, what we are checking are kind of binary values, right? Something like:

00 -> I
01 -> X
10 -> Y
11 -> Z

I think this looks like we could use a hash table instead. This maybe also applies for flag_z:

00 -> No
01 -> Yes
10 -> Yes
11 -> No

And so on and so forth. Let's think about it in a next PR.

ludmilaasb (dismissed):
LGTM!

Sonarcloud is complaining about the cognitive complexity of one of the convertion functions but I think we can address this in a next PR. 

Overall, I like we are handling the proper types instead of reliying on dictionaries. 

Good job! 🚀

adamglos92:
Let's merge first #33

src/monoprop/conversion_utils.py
adamglos92: I reduced the cognitive complexity, not sure if enough to close your #131 issue

ludmilaasb: No no, I think it is fine!

adamglos92:
@ludmilaasb @robertodr in principle this is ready to go (up to some tidying) but it might conflict with #33

src/monoprop/pauli.py
```diff
@@ -111,10 +111,16 @@ def __repr__(self) -> str:
 class PauliOperator:
     """A weighted sum of Pauli terms.
```
diagonal-hamiltonian, line 113: docs

adamglos92: simplified

ludmilaasb (approved):

PR #120, opened 2026-07-21
docs: :memo: symbol unification

### LINKED ISSUE #78: inconsistent symbols between docstrings and documentation
In documentation creation/annihilation operators are noted with c, and Majorana operators with m. In docstrings by `a` and gamma. Personally I'm fine with both, but possibly should be unified

adamglos92:
- [x] replaced `gamma` as symbols for individual Majorana operators with `m` which was used far more in the text (and which is closer to Majorana monomial symbol `M`
- [x] replaced symbol `a` for annihilation/creation operators with `c` (as used in docs)

Closes #78


ozlemsalehi (approved):
Looks ok to me though I am not familiar with this one for instance: 

`terms = " ".join(f"c_{idx}^{op}" for idx, op in self.expression)`

adamglos92:
> Looks ok to me though I am not familiar with this one for instance:
> 
> `terms = " ".join(f"c_{idx}^{op}" for idx, op in self.expression)`

it is just trivial symbol substitution, and this string is just annihilation-creation formalism

ludmilaasb (approved):

PR #121, opened 2026-07-21
refactor: ♻️  improved user experience for only_rotate_len_k

adamglos92:
- [x] None as a default value
- [x] check against invalid values 
- [x] tests for the above

Note the changes are only done on the python side. 
Strongly related to #6, but not sure if enough to close it


ludmilaasb (approved):
LGTM!

ludmilaasb:
But I think we need more people to review it, cause this is related to user experience. @JoshCudbyAlgo ?

adamglos92:
@fpietra pointed it out

JoshCudbyAlgo:
Code changes look good - I haven't used this functionality so can't helpfully comment on user experience here, sorry!

JoshCudbyAlgo (approved):
Code changes look good - can't comment from UX perspective, sorry!

robertodr:
The change looks good to me, but I think it should be done at the c++ level (use std::optional<int> instead of int), since we already got bitten by the implicit cast of integers to booleans... Let's merge this as it is, but also open a new issue to not forget about lowering the logic to c++

adamglos92:
@robertodr the old issue maybe is still relevant, let's continue there

PR #123, opened 2026-07-21
docs: :memo: quantum chemistry tutorial improved

### LINKED ISSUE #90: Correct unexpected plot results in chemistry.ipynb
The energy of the circuit produced in chemistry.ipynb is not monotonic which is unexpected. This requires checking up, but according to @MGDelcey a possible reason might be a non-deterministic Hamiltonian generation

Possible solution would be: 
* hardcode the integrals as npz (with explanation why we are doing this, e.g. determinism). We can keep the pyscf code, but we should comment on why we are doing like this
* replace the current circuit with the accurate one

Alternative solution: Find different molecule (can still help with generating circuit)

adamglos92:
- [x] A particular integrals of Cr2 molecule saved
- [x] ansatz replaced with actually converging variant

Closes #90


matteoacrossi:
@copilot resolve the merge conflicts in this pull request (keep the title from `main`)

docs/notebooks/chemistry/chemistry.ipynb
```diff
@@ -402,6 +393,14 @@
...
     "- Try different properties ($S^2$, natural orbitals)"
    ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "643fbae0-87b0-4039-adf5-af6fee767b1c",
+   "metadata": {},
+   "outputs": [],
+   "source": []
```
matteoacrossi:
```suggestion
```

Remove the last empty cell.

matteoacrossi (approved):

PR #152, opened 2026-07-24
build(cmake): only allow to build as python extension

### LINKED ISSUE #136: Refine coverage collection for C++ files
We decided to leave the points raised by @robertodr's review of #135 for a later PR:
* Remove the narrative comments from the code, because they make me sick 
* Add -fcondition-coverage and -fpath-coverage to the compile flags for the coverage build (they give more information)
* Invoke gcovr with --conditions
* Figure out if you can use these options for gcov --prime-paths --prime-paths-lines --prime-paths-source when calling it through gcovr

_Originally posted by @robertodr in https://github.com/Algorithmiq/monoprop/pull/135#pullrequestreview-4754413234_

robertodr:
By opening this PR I confirm that I have read [CONTRIBUTING.md](https://github.com/Algorithmiq/monoprop/blob/ed30341adbb0a8aea3aabac02ce329be61b2d92e/CONTRIBUTING.md) and I agree to the terms of the [Contributor License Agreement](https://github.com/Algorithmiq/monoprop/blob/ed30341adbb0a8aea3aabac02ce329be61b2d92e/CLA.md).

## Summary

This project is purely a Python extension but the build system, and configuration around it, still offered the (vestigial) option of building as a C++-only project. 

Re-hash of #106, where I made a hideous mistake with the git history.

Fixes #136 

## Changes

- [X] Remove CMake code that allowed to build as a standalone C++ project, as that is no longer meaningful. A warning is issued against compiling with direct invocation to CMake.
- [X] C++ unit tests are gated behind a new CMake option `monoprop_ENABLE_CXX_UNIT_TESTS`, which is ON by default.
- [X] `CMakePresets.json` changed accordingly, as it's effectively only needed for test autodiscovery.
- [x] Edit `justfile` to account for loss of many presets for CMake.
- [X] Overhaul CMake code for installation of artifacts. The package is installed under `lib/python3.X/site-packages/monoprop` all other artifacts are installed in subfolders of it: 
    * `libmonoprop.so` is under `lib` (or `lib64`, depending on the machine)
    * all headers are under `include` with the proper hierarchical folder structure
    * CMake helper files are under `cmake`  The header files are now explicitly listed (instead of globbed) and we use `target_sources` with `FILE_SET` pento handle them sanely (instead of appending to a global list)

    I have dropped the distinction between private and public for now.
- [x] Simplified `test.yml` accordingly:
    * we no longer need separate jobs to run the C++ unit tests: they can be compiled within a wheel build and run, if need be.
    * Single coverage job running both Python and C++ tests.

## Checklist

- [ ] Documentation updated (docstrings, `docs/`, `CONTRIBUTING.md`) if needed

## AI/LLM disclosure

- [ ] I used the following tool to help write this PR description:
- [X] I used the following tool to generate or modify code: GitHub Copilot with auto-selection of model.

---
This is **part 1 of 3 in a stack** made with GitButler:
- <kbd>&nbsp;3&nbsp;</kbd> #180 
- <kbd>&nbsp;2&nbsp;</kbd> #173 
- <kbd>&nbsp;1&nbsp;</kbd> #152 👈


src/monoprop/bindings/CMakeLists.txt
```diff
@@ -134,21 +123,16 @@ target_include_directories(
...
-  RELATIVE_PATH
-  _rel
-  ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_BINDIR}
-  ${CMAKE_INSTALL_PREFIX}
-)
+# handle RPATH
+set(_plat_token "")
 if(APPLE)
-  set(_rpath "@loader_path/${_rel}")
+  set(_plat_token "@rpath")
```
robertodr:
```suggestion
  set(_plat_token "@loader_path")
```

.devcontainer/postStartCommand.sh
```diff
@@ -8,7 +8,7 @@ echo "Set up development environment in $WORKSPACE"
 # NOTE the docs dependency group is excluded by default (it requires Python 3.12
 # and we develop against 3.11); it is built separately via `just build-docs`.
 # dev/test/interactive are the default groups (see [tool.uv] default-groups).
-uv sync --all-extras -v
+uv sync --all-extras -v --config-settings-package="monoprop:cmake.define.monoprop_ENABLE_CXX_UNIT_TESTS=ON"
 
 # install prek hooks
-uv run prek install --install-hooks
+uv run prek install --install-hooks --overwrite
```
ludmilaasb, line 11: Why overwrite?

.github/workflows/test.yml
ludmilaasb: Does this doesn't need to be changed? We are not supposed to build with CMake anymore?

robertodr:
The PR is still in progress, it's not marked as draft as I need all CI jobs to run on each commit to check I'm not losing functionality. Sorry for the confusion!

.github/workflows/test.yml
robertodr: This particular step needs to stay (in some form) as it's the way that we are able to find `libmonoprop.so` in third-party packages. There's a lot of jobs in the `test.yml` workflows that I still need to change/remove.

ludmilaasb: Also, if we are going to use CMake anyway, I think the flag monoprop_ENABLE_CXX_UNIT_TESTS is off?

.devcontainer/postStartCommand.sh
robertodr: I was getting a warning from `prek` when starting the devcontainer. Whenever it sees a  `pre-commit` script under `.git/hooks` folder, it assumes you want to keep those pre-existing hooks, even though they're the same hooks

.github/workflows/test.yml
robertodr: I don't understand the relation of `monoprop_ENABLE_CXX_UNIT_TESTS` with the lines highlighted.

ludmilaasb: Ahh sorry, I wrote it in a extremely dislexic way :sweat_smile: 

I though that by default the flag for enabling the test is off, and there is some building  happening here

robertodr:
this will also need to be merged by bypassing requirements, since the job names changed. Plus a change in required checks in the ruleset.

robertodr:
This is now ready-ready to review

justfile
```diff
@@ -29,54 +29,90 @@ fumapy := "--with " + site + "/node_modules/fumadocs-python --with 'griffe<2' --
...
+test-mpi RANKS='':
+    export OMPI_MCA_rmaps_base_oversubscribe="1"
+    uv sync --all-extras --group test --reinstall-package monoprop --no-cache --config-settings-package="monoprop:cmake.define.monoprop_ENABLE_MPI=ON" -v
+    ranks="${1:-${monoprop_MPI_TEST_PROCS:-2}}"
     for r in ${ranks//;/ }; \
     do echo "Running full Python test suite with ${r} MPI rank(s)"; \
-    mpiexec --allow-run-as-root -n "$r" uv run --no-sync python -m pytest tests --with-mpi -v; \
+    mpiexec -n "$r" uv run --no-sync python -m pytest tests --with-mpi -v; \
+    echo "Running C++ unit tests with ${r} MPI rank(s)"; \
+    ctest --test-dir build/editable/Release --output-on-failure \
```
robertodr:
```suggestion
    ctest --test-dir build/editable/Release --output-on-failure; \
```

robertodr:
Comments to address from Copilot:

**justfile:49**
* The repo docs still reference the old Just recipes (`test-py`, `test-py-mpi`, `test-py-mpi-matrix`) that were removed/renamed here (see `docs/content/docs/testing.mdx:18,26-27`). This makes the documented test commands incorrect after this change.
```
test:
    uv run python -m pytest -m "not mpi"
    ctest --test-dir build/editable/Release --output-on-failure

# MPI is off by default in source builds, so build an MPI-enabled editable install
# first, then run the suite under mpiexec with --no-sync (avoids a per-rank resync).
# Pass RANKS as either a single integer or a semicolon-separated list (e.g. "1;2;4").

test-mpi RANKS='':
    export OMPI_MCA_rmaps_base_oversubscribe="1"
    uv sync --all-extras --group test --reinstall-package monoprop --no-cache --config-settings-package="monoprop:cmake.define.monoprop_ENABLE_MPI=ON" -v
    ranks="${1:-${monoprop_MPI_TEST_PROCS:-2}}"
    for r in ${ranks//;/ }; \
    do echo "Running full Python test suite with ${r} MPI rank(s)"; \
    mpiexec -n "$r" uv run --no-sync python -m pytest tests --with-mpi -v; \
    echo "Running C++ unit tests with ${r} MPI rank(s)"; \
    ctest --test-dir build/editable/Release --output-on-failure \
    done
```
**CMakeLists.txt:32**
* `include/monoprop/Info.h` is generated from `include/monoprop/Info.h.in`, but the placeholders `@PROJECT_VERSION_FULL@` and `@GIT_COMMIT@` are no longer defined anywhere in CMake (DynamicVersion was removed). As a result, `monoprop::version()` and `monoprop::git_commit()` will expand to empty strings in the built artifact.
```
endif()

if(APPLE)
```

PR #172, opened 2026-07-30
feat(benches): 🧪 add a lattice-size scaling benchmark for Pauli propagation

diagonal-hamiltonian:

## What

A second Pauli-propagation benchmark. The existing one walks Trotter steps at a fixed 6x6
lattice; this one holds the circuit depth and the truncation fixed and grows the lattice
instead, from 6x6 (36 qubits) to 18x18 (324 qubits), comparing monoprop against
`PauliPropagation.jl`, `QuEra ppvm`, Qiskit's `pauli-prop`, and `cuPauliProp` on a GPU.

Everything needed to reproduce the figures is in the repo. Either redraw them from the
committed ladder in seconds:

```bash
cd benches/third_party/pauli_prop
uv run python plot_scaling.py scaling_results.jsonl
uv run python plot_speedup.py scaling_results.jsonl
```

or measure your own:

```bash
uv run python run_scaling.py --output results/scaling.jsonl --timeout 300 \
  --threads monoprop=56 juliapp=28
```

## Result

How many times faster monoprop was at each size (a dash = that engine did not finish the size
within the 300 s per-point budget):

| qubits | PauliPropagation.jl | QuEra ppvm | Qiskit pauli-prop | cuPauliProp (GPU) |
| -----: | -----: | -----: | -----: | -----: |
| 36 | 16× | 107× | 165× | 4.4× |
| 64 | 21× | 209× | 213× | 5.7× |
| 100 | 33× | 190× | 287× | 8.5× |
| 144 | 40× | 182× | – | 10× |
| 196 | 55× | 184× | – | 11× |
| 256 | 81× | – | – | 11× |
| 324 | 105× | – | – | 15× |

At a fixed coefficient truncation the operator saturates — past ~100 qubits every engine
carries the same ~5.8M terms — so this measures cost per term kept rather than a growing
problem. monoprop's total runtime stays nearly flat there, which is what widens the gap.

Caveats, all written into the docs and the second figure's footnote:

- **Only two of the four CPU engines use more than one core.** ppvm and Qiskit's `pauli-prop`
  propagate on a single core (measured at 0.99 busy cores mid-propagation) and no environment
  variable changes that, so these ratios are wall-clock, not per-core.
- **`PauliPropagation.jl` runs its fastest documented path**, `Performance.propagate!` on a
  `VectorPauliSum`, which needs the dev branch (0.8.0+). That path leaves duplicate Paulis
  unmerged: ~38% more reported terms and an expectation value differing by ~1e-4.
- **The memory panel is not one quantity** — each engine reports what it exposes; ppvm and
  Qiskit expose nothing but a process-RSS proxy.

## Shape of the change

- `model.py` (the TFIM model, importing no propagation library) and `backends.py` (one function
  per engine, each importing its dependencies *inside* the function, so a missing GPU or
  extension only affects that engine) split out of `run_model.py`.
- `run_scaling.py` sweeps sizes with one subprocess per (size, engine) — `run_one.py` for
  Python, `run_scaling.jl` for Julia, same record schema. Isolation is what makes a long sweep
  survivable: a point that times out or gets OOM-killed costs only itself, is recorded with a
  non-`ok` status, and is skipped at larger sizes. Every record carries its own `host`,
  `threads` and `library_version`.
- `plot_scaling.py` and `plot_speedup.py` derive their titles from the records. Hardware is
  deliberately *not* in a title: the machine qualifies the numbers, so it goes in the markdown
  sidecar's provenance table beside them, and once in the docs prose.
- `pyproject.toml`: **removes** the `RUSTFLAGS = "-C target-feature=+aes,+neon"` override for
  ppvm rather than fixing its value. `neon` is an AArch64 feature name, so the build failed on
  x86_64 — and the override was never needed: ppvm ships a `.cargo/config.toml` that already
  selects the features its gxhash hasher wants per architecture (`+aes,+sse2` on x86_64), and an
  environment `RUSTFLAGS` *replaces* that choice. `uv.lock` is untouched (it records no build
  variables).
- A 2-line fix in the Majorana driver, which crashed writing its own result: `json.dumps`
  rejects the `np.int64` term count.

## Verification

- **`run_model.py` behaviour unchanged.** Per-step term counts, expectation values and operator
  memory are identical to the pre-refactor code on the same build, and the output schema is
  unchanged, so the committed `results.json` and `plot_results.py` are unaffected.
- **Sweep smoke test**: 6x6 with monoprop + Qiskit, both `ok`, Qiskit's term budget correctly
  taken from monoprop's count (5,601,452 for both).
- **ppvm without the override**: gxhash 3.5.0 and the whole ppvm native stack compile with
  `RUSTFLAGS` unset, relying only on the crate's own cargo config.
- **Figures** redrawn from the committed `scaling_results.jsonl` and inspected.
- `ruff` 0.15.20 (check + format) and the `prek` hooks pass on every touched file.

Not verified: a full clean `uv sync` of `benches/third_party` on the machine used here, which
fails for an unrelated reason — `pyqrack-cpu`, a transitive dependency of `bloqade-circuit`,
needs Boost headers to build from source.


diagonal-hamiltonian:

Three commits here extend the PR beyond the Pauli ladder, into the Majorana (1D Hubbard) benchmark next door. Same motivation as the ladder's per-backend thread caps: a wall-clock ratio is only meaningful once you know what each engine did with the cores it was given.

**`f163e87` — record what each run actually did with its cores.** The Hubbard benchmark fixed both sides at 8 threads and recorded no CPU time at all: the monoprop row echoed `$monoprop_NUM_THREADS` back as a string and the Julia row recorded nothing, so a serial run was indistinguishable from a threaded one. Both drivers now measure CPU time in-process across the timed loop (`getrusage` on the Python side, `/proc/self/stat` on the Julia side) and record `cpu_seconds` + `busy_cores` (= CPU/wall) per case, alongside integer `num_threads`, `peak_rss_MB`, `library_version` and `host`. The Julia rows also record `container` — only `VectorMajoranaSum` reaches MajoranaPropagation.jl's AcceleratedKernels path, the dict-backed `MajoranaSum` is serial regardless of `JULIA_NUM_THREADS` — and `gc_seconds`. Thread counts come from the environment rather than being fixed, since the two engines are not fastest at the same count.

That commit also drops the `2N` chemical-potential gates from the monoprop circuit. They are inert (`mu = 0`) and MajoranaPropagation.jl's circuit has no counterpart, but they still cost monoprop ~6% of its runtime; measured term-for-term identical with and without, and `--mu-gates` restores them for the A/B.

**`fafdfb5` — re-measure the committed results.** The old rows were taken at 8 threads on a 112-core node, which left most of the node idle for both engines. The JSONLs and `majorana_results.png` are re-measured with each engine at the count it is fastest at — monoprop 56, MajoranaPropagation.jl 28 (matching the ladder's default). monoprop comes out **22-142x** faster, up from 19-50x, the gap widening with depth and size: 141.6x at 60 sites / 18 layers, 4.2 s vs 592 s. Dividing out the cores each engine actually kept busy leaves **8.5-52x**, so most of the gap is algorithmic. Overlaps still agree to 1.8e-6 and the term-count ratio is unchanged at 1.44-1.50x, so it is the same computation.

The SPECS block is also corrected: this data comes from exclusive DCGP nodes (2x Xeon 8480+, 112 cores), not the 8-core Ice Lake login node it claimed.

**`50eed08` — the Hubbard runtime on its own log-scale figure.** On the combined figure's shared linear axis monoprop's curves are pressed flat against the axis; runtime is also the quantity cited on its own. `plot_results.py` now writes `majorana_runtime.png` beside `majorana_results.png`.

**`dfb2d49` — keep the docs to what the committed data shows.** The prose had grown into an argument about thread scaling — parallel-efficiency percentages, speed-up-over-serial figures, plateau and regression claims, on both the Majorana and the Pauli side. No thread sweep is committed here and no script in the repo measures one, so none of it was checkable from this PR. What each engine was *given* stays (one line per SPECS block, the per-side knobs, the per-row provenance above); the sweep results are gone.

One caveat a reviewer should weigh: the Hubbard headline is best-vs-best at different core counts (56 vs 28), which is why the per-busy-core figure is quoted beside it.

Note the PR title still describes only the Pauli ladder.

docs/content/docs/benchmarks.mdx
```suggestion
Majorana and Pauli propagation engines. The benchmark suite includes memory and time
```

docs/content/docs/benchmarks.mdx
```diff
@@ -28,20 +28,34 @@ Number of terms, final overlap, runtime, and memory vs. circuit layers for the 1
...
+591 s at 60 sites / 18 layers is drawn 140× the height of `monoprop`'s 4.2 s, and all three
+`monoprop` curves consequently lie along the bottom of the axis — annotated with the top of that
+band, since flat here means small, not zero. For `monoprop`'s own growth with depth, read the
+term-count and memory panels of the combined figure, or the numbers below.
+
 
 #### SPECS
 
-The benchmark is run on Leonardo (CINECA), in a single node with the following specifications:
+The benchmark is run on Leonardo (CINECA), one exclusive DCGP node per engine:
```
```suggestion
The benchmark is run on [Leonardo (CINECA)](https://docs.hpc.cineca.it/hpc/leonardo.html), one exclusive DCGP node per engine:
```

docs/content/docs/benchmarks.mdx
```diff
@@ -28,20 +28,34 @@ Number of terms, final overlap, runtime, and memory vs. circuit layers for the 1
...
+
+- CPU: 2× Intel Xeon Platinum 8480+ (112 cores total, Sapphire Rapids)
+- RAM: ~480 GB
 
-- Cores: 8 Intel Xeon Platinum 8358 CPU, 2.60GHz (Ice Lake)
-- RAM: 100 GB
+Each engine runs at the thread count it is fastest at — `monoprop` at 56 threads,
+`MajoranaPropagation.jl` at 28 — since they do not peak at the same count. Every record carries
+its own `num_threads`, `busy_cores`, `host` and `library_version`, so a redrawn figure never
+misreports what produced it.
```
```suggestion
`MajoranaPropagation.jl` at 28 — since they do not peak at the same count.
```

docs/content/docs/benchmarks.mdx
```diff
@@ -94,27 +108,46 @@ Each line of both JSONL files is one benchmark run (one system size / circuit de
...
 - `runtime_seconds`: wall-clock time to run all `n_layers` layers.
 - `memory_MB`: operator memory footprint at the final layer, in megabytes.
-
-The Python results additionally record `num_threads`, the value of the `monoprop_NUM_THREADS`
-environment variable used for the run.
+- `num_threads`: thread count the run was given (`monoprop_NUM_THREADS` / `JULIA_NUM_THREADS`).
+- `cpu_seconds`: CPU time consumed over the timed loop, summed over all threads.
+- `busy_cores`: `cpu_seconds / runtime_seconds` — how many cores the run actually kept busy.
+  A value near 1 means the run was serial whatever `num_threads` claimed, so this is what makes
+  a cross-library comparison interpretable rather than a core-count artifact.
```
```suggestion
  A value near 1 means the run was serial, irrespective of `num_threads`.
```

docs/content/docs/benchmarks.mdx
```diff
@@ -187,6 +220,11 @@ This installs `monoprop`, `qiskit`, `ppvm`, `pauli-prop`, and `cuquantum-python-
 `cuPauliProp (GPU)` engine requires an NVIDIA GPU with a working CUDA setup; the other three Python
 engines (`monoprop`, `QuEra ppvm`, `Qiskit pauli-prop`) run on CPU only.
 
+`ppvm` compiles a Rust extension whose hasher needs specific CPU features. Its own
+`.cargo/config.toml` already selects them per architecture, so leave `RUSTFLAGS` unset: an
+environment override replaces that file's choice, and a value valid on one architecture (say
+`+neon`) fails the build on another.
+
```
```suggestion
```

docs/content/docs/benchmarks.mdx
```diff
@@ -257,6 +295,92 @@ pauli-prop`. `monoprop` tracks the GPU-accelerated `cuPauliProp` closely while r
...
+  container driven by `Performance.propagate!`, which needs the **dev branch (0.8.0+)** for the
+  `Performance` submodule. That path leaves duplicate Pauli strings unmerged, so it reports ~38%
+  more terms than the exact `propagate` path and its expectation value differs by ~1e-4 — its
+  term count is a storage count, not an operator size.
+- **The memory figure is not one quantity.** `monoprop` and `cuPauliProp` report their own
+  operator accounting, `PauliPropagation.jl` its `Base.summarysize`, and ppvm and Qiskit only a
+  process-RSS proxy, because they expose nothing else.
+
+#### SPECS
+
```
robertodr, line 342:
```suggestion
The benchmark is run on [Leonardo (CINECA)](https://docs.hpc.cineca.it/hpc/leonardo.html):

```

docs/content/docs/benchmarks.mdx
```diff
@@ -257,6 +295,92 @@ pauli-prop`. `monoprop` tracks the GPU-accelerated `cuPauliProp` closely while r
...
+- CPU: 2× Intel Xeon Platinum 8480+ (112 cores total) — `monoprop` at 56 threads,
+  `PauliPropagation.jl` at 28 threads.
+- GPU: 1× NVIDIA A100 with 64 GB HBM2e, for `cuPauliProp`.
+
+Each record in the committed data carries its own `host`, `threads` and `library_version`, and
+both plotting scripts print that provenance beside the numbers, so a redrawn figure never
+misreports what produced it.
```
```suggestion
both plotting scripts print that provenance beside the numbers.
```

docs/content/docs/benchmarks.mdx
```diff
@@ -257,6 +295,92 @@ pauli-prop`. `monoprop` tracks the GPU-accelerated `cuPauliProp` closely while r
...
+uv run python run_scaling.py --output results/scaling.jsonl --timeout 300 \
+  --threads monoprop=56 juliapp=28
+```
+
+Isolating each point is what makes a long sweep survivable: an engine that exceeds its budget or
+gets OOM-killed costs only that point, is recorded with a non-`ok` status, and is skipped at
+every larger size (`--no-skip-after-fail` to keep trying). Its curve simply ends at the last
+size it completed. Restrict the engines with `--backends`, the sizes with `--sizes`, and sweep a
+strip rather than squares with `--ny`.
+
```
```suggestion
```

robertodr (approved):

PR #183, opened 2026-08-02
benches: add mlxQ exact-statevector baseline (Apple silicon) to the third-party Pauli benchmark

BoltzmannEntropy:
## What this adds

An **exact statevector baseline** for the third-party Pauli benchmark: [mlxQ](https://github.com/BoltzmannEntropy/Qupertino), an MLX-based simulator for Apple-silicon GPUs, wired in as a `mlxq` backend following the existing per-backend pattern in `benches/third_party/pauli_prop/backends.py` (lazy imports inside the function, one Trotter step + expectation value per timed point, same timer bracket).

Pauli propagation's natural competitor at small-to-medium sizes is exact simulation, and none of the current backends provides that reference. Wherever the 2^n statevector fits in memory its per-step cost is independent of operator growth, so this backend locates the lattice size below which propagation is not worth running — and, symmetrically, makes the size where propagation wins unarguable.

Changes:
- `backends.py`: `run_mlxq()` + registration (`MEMORY_METRICS`, `LABELS`, `THREAD_VARS`, new `APPLE_BACKENDS` group).
- `run_model.py` / `run_one.py`: dispatch and `--backends`/`--backend` choices.
- `run_one.py`: `_num_threads` now falls back to `os.cpu_count()` — `os.sched_getaffinity` does not exist on macOS, so `run_one.py`/`run_scaling.py` previously could not run there for *any* backend.
- `benches/third_party/README.md`: install/usage note.

## Measured (your committed TFIM scenario, resized)

Apple M1 Max (32 GB), macOS 26.3, monoprop 0.8.0 PyPI wheel, medians over ≥3 dedicated-process runs, per-point timing identical to `_run_steps`:

| Grid | Qubits | mlxQ (Metal) | monoprop | Faster |
|---|---|---|---|---|
| 4x4 | 16 | 0.027 s | 1.795 s | mlxQ ~66x |
| 4x5 | 20 | 0.062 s | 2.302 s | mlxQ ~37x |
| 5x5 | 25 | 0.807 s | 2.770 s | mlxQ ~3.4x |
| 4x7 | 28 | 49.8 s | 2.121 s | monoprop ~23x |
| 6x6 | 36 | ~0.5 TB, infeasible | 3.604 s | monoprop only |

Full report with raw per-run data, the kicked-Ising counterpart, and fairness caveats: <https://github.com/BoltzmannEntropy/Qupertino/tree/main/bench/monoprop_compare>.

## Correctness

Cross-validated against monoprop at 3x3/3x4 via `run_model.py`: expectation values agree to ~2.6e-5 (your `lower_atol=1e-6` truncation plus mlxQ's complex64). Independently, both engines match a dense-matrix `expm` reference (monoprop to ~1e-14, mlxQ to ~1e-6).

## Notes for maintainers

- **Dependencies**: `mlxq` is intentionally *not* added to the uv project — MLX wheels only exist for macOS/arm64 and the current `pyproject.toml` pins CUDA packages that can't resolve there anyway. Happy to add it behind a `sys_platform == "darwin"` marker instead if you prefer lockfile-managed installs.
- **Docs aside** (can file as a separate issue): while driving the native API we found `ExpGate`'s parameter p acts as exp(−i·p·P) — back-propagating Z through an X gate with parameter p returns cos(2p) — while the README quick-example comments read as exp(−i·θ/2·P). The library is self-consistent; a clarifying line in the docstring would save independent users a debugging cycle.
- Formatting checked with `ruff format`/`ruff check` (the one `I001` on `backends.py` imports fires on the untouched file as well outside the prek context).
- CLA: acknowledged, ready to sign the Individual CLA when the bot prompts.

### DIFF
--- a/benches/third_party/README.md
+++ b/benches/third_party/README.md
@@ -43,6 +43,26 @@ Any lattice size works: `--nx/--ny` override `settings.json`, and the observable
 lattice (the central horizontally-adjacent bond, which is the committed `[20, 21]` at 6x6)
 rather than staying pinned to indices that a resize would invalidate.
 
+### Optional: exact statevector baseline (Apple silicon)
+
+`--backends ... mlxq` adds [mlxQ](https://github.com/BoltzmannEntropy/Qupertino), an
+MLX-based statevector simulator, as an exact non-propagation baseline. Wherever the
+2^n statevector fits in memory, its per-step cost does not depend on the operator, so
+it locates the lattice size below which propagation is not worth running — on an
+M1 Max (32 GB) the crossover against monoprop sits at roughly 25–28 qubits, and a
+36-qubit statevector (~0.5 TB) is out of reach entirely. It is not part of this uv
+project's dependencies because its MLX dependency only exists on Apple silicon;
+install it there with
+
+```bash
+uv pip install "mlxq @ git+https://github.com/BoltzmannEntropy/Qupertino"
+```
+
+`MLXQ_METAL_KERNELS=1` selects its hand-written Metal kernels (its fastest tier; the
+first step of a process then carries one-off kernel compilation, which the fixed-size
+benchmark's step-0 drop already absorbs). Its memory column reports the exact
+statevector footprint, and its expectation values are exact up to complex64 (~1e-6).
+
 ### Scaling with lattice size
 
 ```bash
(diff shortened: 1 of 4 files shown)


robertodr:
Thanks for the contribution! I'm closing as it's out of scope for the moment.

Our third-party benchmarks are meant to compare against implementation of the _same_ simulation methods. While it's valuable to have an exact baseline, it should be a separate benchmark. Moreover, such a comparison would need to span other, more hardware-agnostic simulators.

PR #205, opened 2026-08-05
fix(benches): :bug: measure peak memory with the kernel high-water mark

Panadestein:

## Description

`benches/_memory.py` measured peak memory by sampling RSS from a background thread and keeping the maximum. monoprop's nanobind bindings never release the GIL, so that thread cannot run while the operation it is meant to measure is running.

## What's new

Peak memory now comes from the kernel's own high-water mark:

- `VmHWM` in `/proc/self/status` — the kernel updates it on every RSS increase, so it cannot miss a transient, and no thread is involved.
- Reset per measurement window by writing mode `5` (`CLEAR_REFS_MM_HIWATER_RSS`) to `/proc/self/clear_refs`. This is a counter reset, not a page-table walk, so it is cheap (27 µs per read, vs 41 µs for the RSS read it replaces).

`HighWaterMark` also records the settled floor (gc + heap trim) before opening the window, so callers can report growth rather than absolute footprint. It degrades gracefully: if `clear_refs` is not writable, `exact` is `False` and it falls back to current RSS.

## MPI

A sampler is still needed under MPI: no per-rank scalar can recover *which ranks peaked at the same moment*, which is what peak-of-sum needs. So `RssSampler` is kept, renamed `PssSampler`, and now samples PSS instead of RSS — summing RSS across ranks double-counted shared pages.

Neither metric is simultaneously exact and MPI-aware, so `conftest.py` records both (`memhwm` and `mem`) and `report.py` presents the exact one.


robertodr (approved):
LGTM! Thanks @Panadestein

robertodr (dismissed):

robertodr (dismissed):

benches/conftest.py
```diff
@@ -95,7 +100,8 @@ def _reduce_sum(comm: Any, value: int) -> int:
 _RESULTS: dict[str, Any] = {
     "meta": {},  # run configuration (ranks, threads, host, ...)
     "params": {},  # resolved random-problem hyperparameters
-    "mem": {},  # node id -> peak RSS bytes (per operation)
+    "mem": {},  # node id -> peak-of-sum PSS bytes (per operation; MPI lower bound)
+    "memhwm": {},  # node id -> summed per-rank peak RSS bytes (per operation; may fall back to exit RSS if VmHWM can't be reset)
```
robertodr:
```suggestion
    "memhwm": {},  # node id -> summed per-rank peak RSS bytes (per operation; may fall back to exit RSS if VmHWM can't be reset)  # noqa: E501
```

robertodr (approved):

PR #187, opened 2026-08-04
style(c++): :art: restructure source tree for C++ files

robertodr:
## Summary

Restructuring the location of the C++ files, since we have a lot more now and it's getting confusing to know where to add new things.

## Checklist

- [x] Tests added or updated to cover the changes
- [x] Documentation updated (docstrings, `docs/`, `CONTRIBUTING.md`) if needed
- [ ] `CHANGELOG` / release notes updated if applicable

## AI/LLM disclosure

- [ ] I did not use LLM tooling, or used it only privately for ideation
- [ ] I used the following tool to help write this PR description:
- [x] I used the following tool to generate or modify code: GPT-5.4

> [!IMPORTANT]
> By opening this PR I confirm that I have read [CONTRIBUTING.md](https://github.com/Algorithmiq/monoprop/blob/fa820c3e5a90f773417c367fedbf2bd50683b496/CONTRIBUTING.md) and I agree to the terms of the [Contributor License Agreement](https://github.com/Algorithmiq/monoprop/blob/fa820c3e5a90f773417c367fedbf2bd50683b496/CLA.md).

> [!WARNING]
> If you're contributing on behalf of your employer, contact [cla@algorithmiq.fi](mailto:cla@algorithmiq.fi) to arrange a Corporate CLA.

PR #209, opened 2026-08-05
chore(sonar)!: :broom: burn down the SonarQube backlog and unblock the quality gate

diagonal-hamiltonian:

Burns down the SonarQube backlog and repairs the coverage import that had been
aborting the entire analysis. **All 106 issues on the `main` baseline are now
accounted for: 73 fixed in code, 33 recorded as deliberate design with
file+rule-scoped suppressions, 0 left open.** Plus the clang-tidy backlog and the
one smell in the newly-in-scope Python package.

## Breaking changes

`refactor(cpp)!` — these `monoprop_EXPORT` signatures change:

| Function | Change |
| --- | --- |
| `ev` | 9 params → `(const EvalRequest &, const detail::CosCallbacks & = {})` |
| `ev_and_grad` | 10 params → `(const EvalRequest &, const detail::CosCallbacks & = {})` |
| `state_operator_derivative_local` | 8 → 7; the two adjacent doubles become `LayerAngle`; `(cos_acc, comm)` → `(comm, cos_acc)` |
| `evolve_step` | argument order normalised to `(comm, cos_scale)` |
| `evolve_operator` | argument order normalised to `(comm, cos_scale)` |

New public types: `EvalRequest` and `LayerAngle` (`MPFunctions.h` / `Evolution.h`),
`PropagatorConfigError` and `MultiPartitionUnsupported`
(`MonomialPropagator.h`).

**No Python-visible change.** The bindings reach this path only through
`MonomialPropagator`'s methods via `make_functional_`;
`src/monoprop/bindings/` and `tools/` are untouched.

16 new exception types are additive, and each derives from the **same** `std` base
its site threw before, so every existing `catch` still catches and every
Python-visible exception type is unchanged.

## Measured before / after

Both columns are read from the Sonar API, not estimated. "Before" is the `main`
analysis that was failing the gate; "after" is the PR analysis of `4cc8f3e`, the
first one that completed at all (see the two aborts below).

| Metric | `main` before | PR after |
| --- | --- | --- |
| **Quality gate** | **ERROR** | **OK** |
| **Issues** (`violations`) | **106** | **0** |
| **Code smells** | **104** | **0** |
| **Bugs** | **2** | **0** |
| **Coverage** as Sonar sees it | **0.0 %** | **70.1 %** |
| Line coverage | 0.0 % | 83.1 % |
| Condition coverage | not reported | 56.2 % |
| Technical debt (`sqale_index`) | 1490 min (~25 h) | 0 |
| Reliability rating | 3 (C) | 1 (A) |
| Maintainability rating | 1 (A) | 1 (A) |
| Security rating / vulnerabilities / hotspots | 1 (A) / 0 / 0 | 1 (A) / 0 / 0 |
| Duplicated lines | 0.3 % | 0.0 % |

Gate conditions, all five of which now pass:

| Condition | Threshold | Before | After |
| --- | --- | --- | --- |
| `new_reliability_rating` | 1 (A) | **3 (C)** — the only failure | **1 (A)** |
| `new_security_rating` | 1 (A) | 1 (A) | 1 (A) |
| `new_maintainability_rating` | 1 (A) | 1 (A) | 1 (A) |
| `new_duplicated_lines_density` | < 3 % | 0.2 % | 0.0 % |
| `new_security_hotspots_reviewed` | 100 % | 100 % | 100 % |

The gate failure was narrow and fully accounted for: there were exactly **12**
RELIABILITY issues — 10 × `cpp:S8417` in `PartitionBarrier.h` plus 2 × `python:S5863`
in `test_majorana.py` — and those same 2 were the project's only 2 bugs. All 12 are
addressed (10 suppressed with rationale, 2 fixed), which is why
`new_reliability_rating` is back to A.

### One caveat, stated plainly

The "after" column's `violations = 0` is measured on the **PR** analysis, and that is
not the same statement as "the project backlog is now empty". SonarCloud PR analysis
only raises issues on lines the PR touches — confirmed two ways: the PR component's
`ncloc` is 5161 against main's 7688, and the scan log says
`Analyzing 1 new or changed file(s) (PR analysis)`.

So read the table as two distinct claims:

- **Directly measured:** the gate flipped to OK, coverage went from 0.0 % to real
  numbers, and this PR introduces **no new issues on any line it touches**.
- **Accounted for, not yet measured:** the 106 → 0 backlog figure rests on the
  per-rule accounting below (73 fixed + 33 suppressed = 106). It can only be
  *measured* on the first `main` analysis after merge.

Please treat that post-merge number as the real verdict over anything in this
description.

## The analysis was aborting, not degrading — twice, for two different reasons

Worth reading first, because it is why none of the earlier numbers were trustworthy.
The coverage import has **two** separate malformed-input paths, and neither fails
softly: each one takes down the entire analysis, so there is no C++ or Python result,
no quality gate and no PR decoration.

**1. Wrong format.** `sonar.coverageReportPaths` parses SonarQube's *generic* format
(`<coverage version="1">`), but the `code-coverage` job hands it gcovr's
`--cobertura` output:

```
Unknown report version: gcovr 8.6. This parser only handles version 1.
```

**2. Then, after switching to the Cobertura importer: 32-bit overflow.** Cobertura
carries a per-line execution count, and Sonar reads it with a signed 32-bit
`getRequiredIntAttribute`:

```
Expected an integer instead of "2922859194" for the attribute "hits"
in cpp-coverage.xml at line 3
```

`2922859194 > INT32_MAX`. The real report peaks at **3050104586** — a propagation
inner loop in a Coverage build reaches billions of executions routinely, so this is
not a freak value; it would recur unpredictably whenever a hot line crosses 2³¹.

**The fix is gcovr's SonarQube generic format** (`--sonarqube`), which records
`covered="true|false"` and carries no execution counts, so it cannot overflow at any
workload — and it still carries `branchesToCover`/`coveredBranches`, so no branch data
is lost (hence the 56.2 % condition coverage above). Verified before pushing by
downloading the actual failing report from the failed run and converting it. gcovr
emits both formats because Codecov still consumes the Cobertura files. Python stays on
Cobertura, whose per-line counts stay small.

Both properties are now documented in `sonar-project.properties` with the exact error
each produces when fed the wrong input, since they are easy to confuse.

Both bugs were mine, from earlier commits on this branch. Verified fixed by reading
the job log rather than trusting the badge: `Sensor Generic Coverage Report` reports
`Imported coverage data for 46 files`, `Sensor CFamily [cpp]` completes,
**100 % of classes, functions and statements parsed successfully** with 0 partially
analyzed compilation units, and there are zero ERROR and zero WARN lines in the scan.

**Overlap with #202, stated for the record.** This branch originally also repaired
the `src/` → `cpp/` path filters left stale by #187. #202 has since landed the same
repair independently, so that finding is no longer this PR's headline and the merge
takes #202's `.clang-tidy` pattern — `^cpp/.*\.(h|hpp|inl)$` — because it covers
`.inl` and ours did not. What remains unique here is the coverage wiring: the
`coverage-reports` artifact hand-off between the two jobs, and the Cobertura import
property. #202 touched neither, which is why `main` still shows 0.0 % coverage.

## A limitation this work uncovered

**Sonar's C++ analysis covers only 5 translation units.** CFamily analyses TUs, not
files, and only `Evolution.cpp`, `MPFunctions.cpp`, `MPGraph.cpp`, `Validation.cpp`
and `detail/pare/PareGraph.cpp` are both compiled and inside `sonar.sources`. The
scan log confirms it independently: `Analysis cache: 0/5 hits`.

Consequence: the entire `MonomialPropagator` template tree — the core user-facing
engine — is indexed for `ncloc` but yields **zero issues of any rule**, because it
is instantiated only in binding and test TUs (which live under
`build/editable/Release/...`). This is why 22 of the 36 generic throws were never
flagged, and it means "40 remaining" understated the true backlog. Not fixed here:
widening TU coverage would surface a large new backlog mid-PR. Filed as a known
limitation instead.

## Fixed in code

**73 Sonar issues**, plus 8 clang-tidy findings and 1 Python smell.

Earlier on this branch (33): explicit lambda captures in `Evolution.cpp` (8) ·
`MajoranaAlgebra.h` `constexpr`/ranges/nesting (4) · `OperatorIndex.h` complexity
(4) · MPI `void*` → `std::byte*` (8) · `Bitset.h` hidden friend (1) ·
if init-statements (2) · `EvolutionHelpers.h` ternary (1) · structured binding (1) ·
`ranges::lower_bound` (2) · `test_majorana.py` bound instances (2).

This round (40):

| Rule | n | What |
| --- | --- | --- |
| `cpp:S107` | 13 | parameter objects: `AlltoallvArgs`/`FlatAlltoallvArgs<T>`/`AlltoallvResolveArgs<T>`, `EvalRequest`, `CosCallbacks`, `ExchangePayload`, `LayerAngle` |
| `cpp:S112` | 14 | 16 dedicated exception types |
| `cpp:S3608` + `cpp:S5019` | 9 | explicit lambda capture lists in `HybridComm.h` |
| `cpp:S134` | 2 | the two triple loops de-duplicated into one helper |
| `cpp:S5827` | 1 | `auto request = MPI_REQUEST_NULL;` |
| `cpp:S6197` | 1 | `std::ranges::max_element` |
| `python:S3358` | 1 | nested conditional in `Circuit.__add__` |

Five things worth a reviewer's attention:

- **The parameter objects each remove a live miscall hazard, not a hypothetical
  one.** `op`, `gen_coeffs` and `params` are all `const VecD &` and adjacent, and
  `prepare_evolved_operator` took them in a *different* order — a positional swap
  compiled and silently evaluated a different expression. Likewise the
  anonymous-namespace `TrigValues` takes `(param, gen_coeff)` while its caller had
  `(gen_coeff, param)`: same type, adjacent, opposite order. That is what `LayerAngle`
  fixes.
- **`EvalRequest` holds `MPGraphView` by value rather than being templated on the
  graph type.** Templating cannot work for `monoprop_EXPORT` non-template functions
  defined in a `.cpp`: `ev` would take `EvalRequest<MPGraph>` and rebuild an
  `EvalRequest<MPGraphView>` field by field, recreating the very re-splitting hazard
  the struct exists to remove. Holding the view makes `ev`/`ev_and_grad`
  signature-identical to their `_impl`s, so `ev_impl` and `ev_and_grad_impl` are
  **deleted** rather than becoming pass-throughs. Four functions proved to be pure
  adaptors and are gone; net **−80 lines**.
- **The `invoke_result_t` trap is closed, not merely survived.** It spelled out 8
  argument types while the `func(...)` call site passed 10, relying on two trailing
  default arguments — so a partial edit could compile and mis-deduce `R`. The list now
  spells all three types, the call passes three, and the lambdas carry no defaults, so
  no shorter deduction list exists. The deliberately-unnamed 10th padding parameter is
  gone too: with `CosCallbacks`, both lambdas already share an arity.
- **`evolve_step` is *not* deleted** despite looking like an argument-order wrapper.
  It also converts `const Layer &` → `LayerTraversal` and pins `layer_idx = 0`, and it
  is the only header-declared entry point for the single-layer call.
- **The argument bundles live in `Comm.h`, not `HybridComm.h`**, because `ShmComm.h`
  takes the resolve bundle and must still compile with MPI off. That is also why
  `MPI_Datatype` stays a separate parameter: an `#ifdef`-guarded member would make an
  installed header's layout depend on `monoprop_ENABLE_MPI`.

### How the exception pass kept every test green

Every new type derives from the **same** `std` base its site threw before, and every
message string is preserved character-for-character. There is no exception
translation code in this repo — `register_exception_translator` and `nb::exception`
return zero hits — so nanobind's built-in table, which dispatches on the nearest
`std` base, *is* the C++→Python contract: `runtime_error`/`logic_error` →
`RuntimeError`, `invalid_argument` → `ValueError`, `out_of_range` → `IndexError`.
Same base ⇒ same Python type ⇒ **no test file needed editing.**

Message preservation was checked mechanically, not by eye: literals were extracted
from `HEAD` and from the tree, adjacent-literal runs joined into their concatenated
values, and the multisets compared. `clang-format` re-wraps some multi-line literals
when a type name shifts the alignment — value-safe precisely because adjacent
literals concatenate, which is why the check compares joined values rather than raw
lines.

Two deliberate calls: `MPGraphEncodingStorage.h` keeps `std::logic_error` as its base
(nothing observes it, so either was free; keeping it avoids silently turning a Python
`RuntimeError` into a `ValueError`). And the 9 `std::invalid_argument` /
`std::out_of_range` sites are **not** S112 fixes — the rule flags only
`std::exception`, `std::logic_error` and `std::runtime_error`, and its own *compliant*
example throws `std::invalid_argument`. Those sites were already compliant; they are
converted anyway so identical validation code does not diverge by which subtype it
happened to pick. `throw std::runtime_error`, `throw std::logic_error` and
`throw std::exception` now return **zero** hits across `cpp/monoprop` and
`cpp/include`.

## Suppressed as deliberate design (33)

In `sonar-project.properties` rather than "Accepted" in the UI, so the rationale is
version-controlled and reviewable. Every entry is scoped to **one file + one rule**,
never project-wide — accepted knowingly, since a file+rule entry also hides *future*
issues of that rule in that file, which is exactly why they stay narrow.

| Rule | n | Where | Why not a fix |
| --- | --- | --- | --- |
| `cpp:S8417` | 10 | `PartitionBarrier.h` | `seq_cst` would put a full barrier in the spin loop of a measured hotspot (S=112). The `release` store to `gen_` publishes the preceding `relaxed` reset of `arrived_`; the rule's own text sanctions the exception when you have profiled and documented it, and both were already true in-file |
| `cpp:S5827` | 10 | `HybridComm.h` | explicit `const size_t`/`const int` documents the widening, in code whose comments stress bit-identical reproducibility |
| `cpp:S1238` | 6 | `Evolution.cpp`, `MPFunctions.cpp`, `PareGraph.cpp` | `Comm` is a 32–40 byte trivially-copyable non-owning POD, documented in `Comm.h` as passed by value on purpose. **~10 further by-value `Comm` params are unflagged**, so fixing 6 would leave the codebase inconsistent with its own convention |
| `cpp:S5213` | 3 | `Evolution.cpp`, `PareGraph.cpp` | documented type-erasure boundary; templating forces `monoprop_EXPORT` definitions into public headers and instantiates per `NumModes`. Invoked once per *layer*, not per term |
| `cpp:S1709` | 1 | `TypeAliases.h` | allocator rebind converting constructor — `explicit` would break every `DefaultInitVector<T>` instantiation. Textbook false positive |
| `cpp:S1709` | 1 | `Comm.h` | annotated `// implicit on purpose` |
| `cpp:S954` | 1 | `TypeAliases.h` | mid-file include is load-bearing and documented: `OperatorIndex.h` needs `TermIndex`/`MonomialHash` first |
| `cpp:S5414` | 1 | `MPGraphEncodingTypes.h` | `LayerCore` is a public-field aggregate written from ≥4 sites; privatising costs ~8 accessors |

## clang-tidy — 13 fixed, and a pre-existing backlog now visible

When the job first came back to life it reported **10** findings, and this PR fixes
**13**: 5 `readability-math-missing-parentheses` plus 5 unreported siblings in
`OperatorIndex.h` (fixed so byte-identical index arithmetic does not diverge), and 3
`readability-redundant-member-init`.

The `redundant-member-init` three are all plain `std::vector` with the default
allocator, whose default constructor is user-provided and always yields an empty
vector, so dropping `{}` cannot introduce indeterminate values. Specifically **not**
`DefaultInitVector`, which deliberately skips zero-initialization and is left alone.

**Be aware the current run reports 142 warnings / 55 unique locations, and that is
expected, not a regression.** Two things widened the lens at once: #202's
`HeaderFilterRegex` now includes `.inl`, and the job's `-line-filter` covers every
changed C++ file — of which this PR has many. So pre-existing findings in files this
PR happens to touch are now reported for the first time.

**None of them are in code this PR adds.** I checked each location against the diff.
The two that look like ours are not: `Comm.h:49` is the deliberately-implicit
`MPI_Comm` constructor (annotated in-file, and suppressed for Sonar as
`commimplicit`), and `MPFunctions.h:73/75` are `EvalState`'s members, not the new
`EvalRequest`.

Remaining, all pre-existing, for a follow-up:

| Check | unique locations |
| --- | --- |
| `readability-math-missing-parentheses` | 34 |
| `modernize-use-integer-sign-comparison` | 5 |
| `misc-header-include-cycle` | 3 |
| `readability-redundant-member-init` | 2 |
| `readability-qualified-auto` | 2 |
| `modernize-return-braced-init-list` | 2 |
| `misc-use-internal-linkage` | 2 |
| 5 further checks | 1 each |

I deliberately did **not** widen this PR to absorb them: they are a different tool
from the Sonar backlog this PR is scoped to, the job is `continue-on-error` so nothing
is blocked, and folding ~34 unrelated mechanical edits into a PR that already carries
breaking API changes would make review harder, not easier. The counts above should make
a follow-up trivial to scope.

**The `misc-header-include-cycle` findings are deliberately left unsuppressed.** They
are the same root cause as the `cpp:S954` mid-file include in `TypeAliases.h` that this
PR suppresses as deliberate — clang-tidy has independently confirmed the cycle is real,
which is worth recording rather than silencing. Breaking it means splitting
`TypeAliases.h`, which is its own change.

## The newly-in-scope Python

`src/monoprop` was outside `sonar.sources` entirely, which also made importing Python
coverage meaningless. It raises exactly **one** issue — the `python:S3358` nested
conditional, fixed here.

That is a measured result, not an assumption. Because PR analysis is new-code-scoped
and this PR doesn't modify `src/monoprop/*.py`, the smells were invisible on the PR,
so all 14 in-scope files were analysed directly through the analyzer, with a synthetic
smelly file as a positive control (it correctly returned 8 findings across
maintainability, reliability and security — so the clean result is the analyzer
working, not failing silently). Zero reliability, zero security, zero hotspots.
Duplication is the one metric snippet analysis can't measure; a CPD-like estimate
bounds the worst case at ~1.7 %, under the 3 % gate threshold.

Also worth correcting: there are **16** Python files under `src/`, not the ~37 the
scan's file count suggested — that number included `tests/`.

## Verification

- **526 Python tests pass, 9 skipped** — unchanged from baseline, and re-run after the
  `main` merge.
- **210 C++ unit tests pass** in a `monoprop_ENABLE_MPI=ON` build with **0 warnings**
  under `-Wall -Wextra` on a clean rebuild. This matters because the Python build has
  MPI **off**, so it is the only configuration that compiles `HybridComm.h` and the
  full evaluation path at all.
- **17 comm tests pass on each of 2 ranks** under `mpirun --oversubscribe -n 2`,
  exercising the cross-rank staging path the `AlltoallvArgs` change touches.
- The MPI-**off** path was checked by compiling a TU including `Exchange.h` and
  `MPICompat.h` without `monoprop_ENABLE_MPI` — the payoff for keeping MPI types out
  of `Comm.h`.
- **No test file was edited** in the exception or parameter-object work.
- `clang-format`, `ruff`, license headers and workflow validation all pass via `prek`.

### Two things left open, deliberately

- **Performance is not measurable on my machine and needs CI or a pinned host.** The
  same binary run twice varies by up to 68 %, and
  `test_random_build_graph[heisenberg]` has a stddev of 31 % of its median at 25
  rounds. An alternating A/B against a pristine base build was dominated by thermal
  drift, with overlapping ranges. The case for codegen-neutrality is therefore
  **structural**, not empirical: `[[gnu::always_inline]]` is preserved on the
  extracted probe helper, the software pipeline and both `__builtin_prefetch` calls
  are byte-identical, and the parameter objects are aggregates of the same references
  passed before. Please let the bench job gate this.
- **`libmonoprop.dylib` does not link in my local MPI test tree** (missing MPI
  symbols). This is **pre-existing** — reproduced on a clean tree before any edit, and
  a full reconfigure does not fix it. Every TU compiles and both
  `monoprop_unit_tests.x` and `_core` link and run. Worth a separate look; out of
  scope here.

## Review order

1. `fix(sonar)` — the Cobertura import (**read first**; it is why the other numbers
   were untrustworthy)
2. `chore(sonar)` — suppressions with rationale
3. `style(cpp)` — clang-tidy readability
4. `refactor(cpp)` — MPI `void*` → `std::byte*`, then the alltoallv bundles
5. `refactor(cpp)!` — parameter objects (the breaking one)
6. `refactor(cpp)` — dedicated exceptions


cpp/monoprop/Bitset.h
```diff
@@ -163,8 +163,8 @@ class Bitset {
         return *this;
     }
 
-    [[nodiscard]] constexpr auto operator>>(size_t pos) const noexcept -> Bitset {
-        Bitset r = *this;
+    [[nodiscard]] friend constexpr auto operator>>(const Bitset &lhs, size_t pos) noexcept -> Bitset {
```
robertodr: I don't think sonar is right here

tests/test_majorana.py
```diff
@@ -58,9 +58,12 @@ def test_from_unsorted_cancels_repeated_indices(self):
         assert sign_mixed == -1.0
 
     def test_equal_terms_compare_and_hash_alike(self):
-        assert Majorana(4, 5) == Majorana(4, 5)
-        assert hash(Majorana(4, 5)) == hash(Majorana(4, 5))
-        assert {Majorana(4, 5), Majorana(4, 5)} == {Majorana(4, 5)}
+        # Bound to distinct instances rather than repeating the constructor call: equality and
+        # hashing must be by value, and separate objects are what makes that assertion meaningful.
```
robertodr:
```suggestion
```

sonar-project.properties
```diff
@@ -11,3 +11,69 @@ sonar.verbose=true
 sonar.python.version=3.11,3.12,3.13,3.14
 sonar.lang.patterns.c=**/*.c
 sonar.lang.patterns.cpp=**/*.cpp,**/*.h,**/*.hpp,**/*.inl
+
+# Coverage is generated by the code-coverage job in .github/workflows/qa-analysis.yml and
+# handed to the sonarqube-analysis job as the `coverage-reports` artifact. Without these,
+# Sonar reports 0% regardless of what Codecov shows.
+sonar.python.coverage.reportPaths=python-coverage.xml
+sonar.coverageReportPaths=cpp-coverage.xml,cpp-coverage-through-python-bindings.xml
```
robertodr: we know we don't care about the coverage that sonar reports, so this seems redundant. If I recall correctly, @ludmilaasb also deactivated coverage from the sonar quality assessment, i.e. when sonar is red, it's because of the static analysis.

.github/workflows/qa-analysis.yml
```diff
@@ -46,11 +46,11 @@ jobs:
         with:
           matrix: true
           files: |
-            src/monoprop/**/*.cpp
-            src/monoprop/**/*.h
-            src/monoprop/**/*.inl
-            include/monoprop/**/*.h
-            include/monoprop/**/*.inl
+            cpp/monoprop/**/*.cpp
```
robertodr, line 49: I've already got a branch with most of these fixes 😅

.github/workflows/qa-analysis.yml
```diff
@@ -127,13 +127,28 @@ jobs:
           disable_search: true
           fail_ci_if_error: true
 
+      # Handed to the sonarqube-analysis job, which imports them via
```
robertodr: as I commented later on, I don't think this is necessary, because we know that that 0% reported by sonar is bogus.

diagonal-hamiltonian:

## Measured before / after — the analysis now completes

The PR body promised these numbers once the analysis ran. It does now, on `4cc8f3e`.

| Metric | `main` before | PR after |
| --- | --- | --- |
| **Quality gate** | **ERROR** | **OK** |
| `new_reliability_rating` — the only failing condition | **3 (C)** | **1 (A)** |
| Coverage as Sonar sees it | **0.0 %** | **70.1 %** |
| Line coverage | 0.0 % | 83.1 % |
| Condition coverage | — | 56.2 % |
| Issues (`violations`) | 106 | 0 |
| Code smells | 104 | 0 |
| Bugs | 2 | 0 |
| Technical debt (`sqale_index`) | 1490 min (~25 h) | 0 |
| Reliability rating | 3 (C) | 1 (A) |
| Duplication | 0.3 % | 0.0 % |

All five gate conditions pass: `new_reliability_rating` 1, `new_security_rating` 1,
`new_maintainability_rating` 1, `new_duplicated_lines_density` 0.0 %,
`new_security_hotspots_reviewed` 100 %.

### Read `violations = 0` carefully

It does **not** by itself prove the 106-issue backlog is gone. SonarCloud PR analysis
only raises issues on lines the PR touches — confirmed two ways: PR `ncloc` is 5161
against main's 7688, and the scan log says `Analyzing 1 new or changed file(s)
(PR analysis)`.

What it does prove: **this PR introduces no new issues on any line it touches.** The
106 → 0 claim rests on the per-rule accounting in the PR body, and the real verdict is
the first `main` analysis after merge. Please treat that number as authoritative over
this one.

### It took two fixes to get here, and both failed hard rather than softly

The coverage import has two distinct malformed-input paths, each of which aborts the
*entire* analysis — no C++ or Python results, no gate, no decoration:

1. `sonar.coverageReportPaths` wants the generic format and got Cobertura →
   `Unknown report version: gcovr 8.6`
2. `sonar.cfamily.cobertura.reportPaths` reads the per-line `hits` count as a **signed
   32-bit int** → `Expected an integer instead of "2922859194"`. The real report peaks
   at **3050104586**; a propagation inner loop in a Coverage build reaches billions of
   executions routinely, so that would have recurred at random.

Settled on gcovr's `--sonarqube` generic format: `covered="true|false"`, no execution
counts, so overflow is impossible at any workload — and `branchesToCover` survives, which
is why condition coverage is 56.2 % rather than absent. Both properties are now
documented in `sonar-project.properties` with the exact error each produces when fed the
wrong input.

### Verified from the job log, not the badge

`Sensor Generic Coverage Report` → `Imported coverage data for 46 files`;
`Sensor CFamily [cpp]` completes; **100 % of classes, functions and statements parsed
successfully** (0/76, 0/380, 0/1794 errors) with 0 partially analyzed compilation units;
zero ERROR and zero WARN lines in the scan.

### A limitation this measurement exposes

`Analysis cache: 0/5 hits` — CFamily analysed exactly **5 translation units**. Only
`Evolution.cpp`, `MPFunctions.cpp`, `MPGraph.cpp`, `Validation.cpp` and
`detail/pare/PareGraph.cpp` are both compiled and inside `sonar.sources`, so the whole
`MonomialPropagator` template tree — the core engine — is indexed for `ncloc` but yields
zero issues of any rule. That is why 22 of the 36 generic throws were never flagged, and
why "40 remaining" understated the real backlog. Left as a known limitation rather than
widened mid-PR.

.github/workflows/qa-analysis.yml
```diff
@@ -127,13 +129,35 @@ jobs:
...
 
+      # Handed to the sonarqube-analysis job. The C++ side ships the *-sonar.xml pair
+      # (sonar.coverageReportPaths) and not the Cobertura ones: Cobertura carries a per-line
+      # execution count, and Sonar's importer reads it as a signed 32-bit int, so a hot line
+      # crossing 2^31 aborts the entire analysis with
+      #   Expected an integer instead of "2922859194" for the attribute "hits"
+      # which a propagation inner loop in a Coverage build reaches routinely. gcovr's SonarQube
+      # format records covered="true|false" with no counts, so it cannot overflow. Python keeps
+      # Cobertura (sonar.python.coverage.reportPaths) -- coverage.py emits per-line hits that stay
+      # small. Codecov above still reads the Cobertura files, so gcovr emits both.
```
robertodr:
```suggestion
```

.github/workflows/qa-analysis.yml
```diff
@@ -127,13 +129,35 @@ jobs:
...
+          retention-days: 1
+
   sonarqube-analysis:
     # in combination with the PR types selection, this top-level if statement
     # skips running the workflow for Draft PRs
     if: ${{ github.event_name == 'push' || !github.event.pull_request.draft || contains(github.event.pull_request.labels.*.name, 'test-in-draft') }}
 
     name: SonarQube analysis
     runs-on: ubuntu-26.04
+    # Coverage is produced by the code-coverage job; Sonar reports 0% without it.
```
robertodr:
```suggestion
```

cpp/include/monoprop/Evolution.h
```diff
@@ -43,29 +43,36 @@ class MPGraphView;
 
 struct LayerCore;
 
+/// One layer's rotation angle in factored form: the layer rotates by 2·gen_coeff·param.
+// Named rather than two adjacent doubles because a positional swap compiles, and the two factors are not
+// interchangeable downstream: gen_coeff also scales the returned gradient.
```
robertodr:
```suggestion
```

cpp/include/monoprop/MonomialPropagator.h
```diff
@@ -53,6 +54,24 @@ class PartitionGroup;
 } // namespace partition
 } // namespace detail
 
+/// A propagator setting is out of range, or inconsistent with another setting.
+// Covers a crossed atol pair and a logical width outside [1, NumModes]; also thrown from
+// MonomialPropagatorImpl.h, so it is declared here where both the class body and the impl see it.
+// std::runtime_error is load-bearing, not a default -- nanobind's built-in translation dispatches on
+// the nearest std base, so any other base would change the Python exception type these surface as.
```
robertodr:
```suggestion
// Covers a crossed atol pair and a logical width outside [1, NumModes]; also thrown from
// MonomialPropagatorImpl.h
```

cpp/include/monoprop/MonomialPropagator.h
```diff
@@ -53,6 +54,24 @@ class PartitionGroup;
...
+class PropagatorConfigError : public std::runtime_error {
+public:
+    using std::runtime_error::runtime_error;
+};
+
+/// A raw-layout accessor was called on a multi-partition propagator, which owns no operator or graph.
+// Kept apart from PropagatorConfigError because nothing about the configuration is wrong: the caller
+// either uses the partition-transparent accessors or constructs with partitions=1. Same std base.
```
robertodr:
```suggestion
```

cpp/include/monoprop/MonomialPropagator.h
```diff
@@ -262,32 +281,15 @@ class MonomialPropagator {
...
 protected:
-    static inline const auto ev_fn = [](double e_core,
-                                        const EvalState &state,
-                                        const VecD &op,
-                                        const VecZ &parameter_mapping,
-                                        const VecD &gen_coeffs,
-                                        const auto &graph,
-                                        const VecD &params,
+    // Both take the same three parameters so make_functional_ has one call to write; the energy path
+    // simply ignores cos.accumulate.
```
robertodr:
```suggestion
```

cpp/include/monoprop/MonomialPropagator.h
```diff
@@ -458,16 +461,10 @@ class MonomialPropagator {
                               VecD *fused_scale_coeffs = nullptr,
                               bool *fused_scale = nullptr) -> std::shared_ptr<LayerCore>;
 
+    // The argument list is spelled in full, and make_functional_ passes exactly these: deducing R from a
+    // shorter list than the call site uses is how a default argument silently deduces the wrong R.
```
robertodr:
```suggestion
```

cpp/include/monoprop/MPFunctions.h
```diff
@@ -81,28 +81,32 @@ monoprop_EXPORT auto map_params(const VecD &parameters,
...
-                        const VecD &op,
-                        const VecZ &parameter_mapping,
-                        const VecD &gen_coeffs,
-                        const MPGraph &graph,
-                        const VecD &params,
+/// Everything one expectation-value evaluation reads, in one object: the graph plus the arrays that index
+/// into it. Built at the call site and consumed there -- every member but `graph` is a non-owning
+/// reference, and `graph` is itself a view over layers the caller keeps alive.
+// Named rather than passed positionally because `op`, `gen_coeffs` and `params` are all const VecD & and
+// sit next to each other: a positional swap compiles and silently evaluates a different expression.
```
```suggestion
```

cpp/monoprop/detail/evolution/CosineRecomputeCallbacks.h
```diff
@@ -28,4 +28,12 @@ using LayerCosScale = std::function<void(size_t layer, double *coeff, double cos
 using LayerCosAccumulate =
     std::function<double(size_t layer, double *state, double *ham, double cos_val, double sec_val)>;
 
+// The two callbacks travel together from the one builder that makes them to every evaluation entry point,
+// so they are one parameter rather than two adjacent std::functions. Either may be empty; the paths that
+// need one say so (see ev/ev_and_grad).
```
robertodr:
```suggestion
```

cpp/monoprop/detail/graph_encoding/MPGraphEncodingStorage.h
```diff
@@ -28,6 +28,15 @@
 
 namespace monoprop::detail {
 
+// The layer exchange layout and the packed cross-rank storage disagree on the rank count.
+// Kept a logic_error descendant deliberately: both operands are built inside build_graph, so no input
+// can make them disagree -- a mismatch is a violated internal invariant, which is what logic_error
+// means. It also keeps the Python-visible type RuntimeError rather than flipping it to ValueError.
```
robertodr:
```suggestion
```

cpp/monoprop/detail/monomial_propagator/MonomialPropagatorCommon.h
```diff
@@ -23,6 +23,14 @@
 
 namespace monoprop::detail {
 
+// A CutoffType enumerator neither cutoff factory knows. Reachable only through a value cast into the
+// enum, so it stays a std::runtime_error descendant (Python RuntimeError) rather than becoming an
+// argument error: no in-range enumerator can produce it.
```
robertodr:
```suggestion
// A CutoffType enumerator neither cutoff factory knows.
```

cpp/monoprop/detail/monomial_propagator/MonomialPropagatorImpl.h
```diff
@@ -39,6 +40,39 @@
 
 namespace monoprop {
 
+// The ranks disagree on S. Kept out of PropagatorConfigError (declared in MonomialPropagator.h, and
+// used below for the per-propagator settings): the count comes from partitions= or the environment on
+// every rank independently, so the fix is to the launch, and it may belong to a different rank.
```
robertodr:
```suggestion
// The ranks disagree on S. The count comes from partitions= or the environment on
// every rank independently, so the fix is to the launch, and it may belong to a different rank.
```

cpp/monoprop/detail/monomial_propagator/MonomialPropagatorImpl.h
```diff
@@ -39,6 +40,39 @@
...
+// The caller recovers by contracting or rebuilding the graph, not by fixing an isolated argument.
+class GraphStateConflict : public std::runtime_error {
+public:
+    using std::runtime_error::runtime_error;
+};
+
+// The (basis, cutoff_type, basis_change) triple is inconsistent: a Pauli basis with a Length cutoff or
+// a basis change, or a basis-change table that is not 2*logical_num_modes rows.
+// std::invalid_argument, not runtime_error: nanobind's built-in translation dispatches on the nearest
+// std base, and these are Python ValueError today.
```
robertodr:
```suggestion
```

cpp/monoprop/detail/monomial_propagator/MonomialPropagatorImpl.h
```diff
@@ -39,6 +40,39 @@
...
+class CutoffConfigError : public std::invalid_argument {
+public:
+    using std::invalid_argument::invalid_argument;
+};
+
+// A coefficient-informed build_graph() was given fewer parameter values than replaying the stored graph
+// as a seed needs. Same std base as CutoffConfigError for the same Python-visibility reason, but a
+// separate type: this is one short vector, not a misconfigured propagator.
```
robertodr:
```suggestion
// as a seed needs.
```

cpp/monoprop/detail/mpi/Comm.h
```diff
@@ -63,4 +65,64 @@ struct Comm {
...
+    std::byte *recv = nullptr;        // caller-owned, already sized for recv_counts
+    const int *recv_counts = nullptr; // [P] input: the senders' transpose, same contract as MPI_Alltoallv
+    const int *recv_displs = nullptr; // [P] element offsets into `recv`
+    size_t elem = 0;                  // bytes per element; scales every count/displ above
+};
+
+// Typed view of the same six spans, for callers that hold `T` buffers: `bytes()` type-erases it for the
+// in-process transports, which address payloads as raw bytes, while the MPI path keeps the typed
+// pointers it passes alongside a datatype. Same lifetime and element-offset contract as AlltoallvArgs.
+template <class T>
```
robertodr:
```suggestion
template <typename T>
```

cpp/monoprop/detail/mpi/Comm.h
```diff
@@ -63,4 +65,64 @@ struct Comm {
...
+                             .recv_displs = recv_displs,
+                             .elem = sizeof(T)};
+    }
+};
+
+// The fused resolve verb keeps its own shape because its recv side is an output, not a caller-supplied
+// layout: `recv` is a reference the callee resizes, and the two recv arrays are written, not read. Being
+// typed, it derives element bytes as sizeof(T) instead of carrying `elem`. Send side: same lifetime and
+// element-offset contract as AlltoallvArgs.
+template <class T>
```
robertodr:
```suggestion
template <typename T>
```

cpp/monoprop/detail/mpi/HybridComm.h
```diff
@@ -39,6 +41,14 @@
 
 namespace monoprop::mpi {
 
+// The launching environment, not a caller's argument, decides the MPI thread level, so this is kept
+// distinct from CollectiveArgumentError; both stay std::runtime_error so the Python-visible type is
+// unchanged either way.
```
robertodr:
```suggestion
```

cpp/monoprop/Bitset.h
```diff
@@ -163,8 +163,8 @@ class Bitset {
...
+        Bitset r = lhs;
```
robertodr: reverting, as this does not need to be a `friend`. It makes no sense for the arguments to the operator to be symmetric in this case and the `const` signature adheres to the one in the standard 
```suggestion
    [[nodiscard]] constexpr auto operator>>(size_t pos) const noexcept -> Bitset {
        Bitset r = *this;
```

cpp/monoprop/MPFunctions.cpp
```diff
@@ -22,6 +22,22 @@
 
 namespace monoprop {
 
+// A per-layer cosine callback this evaluation path needs was left empty. Derived from
+// std::invalid_argument because nanobind's built-in translation dispatches on the nearest std base:
+// any other base would change the Python exception type these already surface as.
```
robertodr:
```suggestion
// A per-layer cosine callback this evaluation path needs was left empty.
```

cpp/monoprop/MPFunctions.cpp
```diff
@@ -22,6 +22,22 @@
...
+class MissingLayerCallback : public std::invalid_argument {
+public:
+    using std::invalid_argument::invalid_argument;
+};
+
+// A state's rows and values disagree in length, a sparse row names a slot outside the state, or the
+// operator being contracted is shorter than the state. One type because every case is a caller-supplied
+// length or index that does not fit the rest of the call. Same std base, so both stay ValueError.
```
robertodr:
```suggestion
// A state's rows and values disagree in length, a sparse row names a slot outside the state, or the
// operator being contracted is shorter than the state. One type because every case is a caller-supplied
// length or index that does not fit the rest of the call.
```

cpp/monoprop/MPFunctions.cpp
```diff
@@ -53,84 +69,19 @@ auto fill_mapped_params(VecD &result,
...
-auto prepare_evolved_operator(EvalScratch &scratch,
-                              const VecD &op,
-                              const VecD &params,
-                              const VecZ &parameter_mapping,
-                              const VecD &gen_coeffs,
-                              const MPGraphView &graph,
-                              const mpi::Comm &comm,
-                              const detail::LayerCosScale &cos_scale) -> void {
+// Leaves the evolved operator in eval_scratch().op. Reaches for the scratch itself rather than taking it:
+// every caller wants this exact thread_local, and passing it would only invite a second one.
```
robertodr:
```suggestion
```

cpp/monoprop/MPGraph.cpp
robertodr:
```suggestion
        return {layers_, active_end_index() - k, k, true};
```

robertodr:
```suggestion
    return {layers_, active_begin_index(), k, false};
```

cpp/monoprop/Validation.cpp
```diff
@@ -20,14 +20,31 @@
 
 namespace monoprop {
 
+// Every validate_* precondition on a caller-supplied argument reports through this one type. Derived
+// from std::runtime_error because nanobind's built-in translation table dispatches on the nearest std
+// base: changing the base would change the Python exception these surface as.
```
robertodr:
```suggestion
// Every validate_* precondition on a caller-supplied argument reports through this one type.
```

cpp/monoprop/Validation.cpp
```diff
@@ -20,14 +20,31 @@
...
+class ValidationError : public std::runtime_error {
+public:
+    using std::runtime_error::runtime_error;
+};
+
+// The graph was rebuilt after a functional captured its layer count, so the functional's parameter
+// mapping no longer describes the graph. Kept apart from ValidationError because the arguments are
+// well-formed here -- the object moved under them, and the caller recovers by rebuilding the
+// functional rather than by fixing a call. Same std base, so both surface identically in Python.
```
robertodr:
```suggestion
// The graph was rebuilt after a functional captured its layer count, so the functional's parameter
// mapping no longer describes the graph.
```

sonar-project.properties
```diff
@@ -11,3 +11,83 @@ sonar.verbose=true
...
+#   sonar.coverageReportPaths       SonarQube's generic format, <coverage version="1">.
+#                                   Fed a Cobertura file it dies with
+#                                   "Unknown report version: gcovr 8.6".
+#   sonar.cfamily.cobertura.*       Cobertura. Reads the per-line "hits" count as a signed
+#                                   32-bit int, so a hot line past 2^31 dies with
+#                                   'Expected an integer instead of "2922859194"'.
+#
+# So the C++ side uses the generic format (gcovr --sonarqube), which records covered="true|false"
+# with no counts and therefore cannot overflow, whatever the workload. Python stays on Cobertura:
+# coverage.py's per-line hits stay small. gcovr emits both formats because Codecov reads Cobertura.
```
robertodr:
```suggestion
```

robertodr (approved):
LGTM, thanks a lot @diagonal-hamiltonian !

PR #222, opened 2026-08-11
fix(docs): 🐛 make the README quick-start examples runnable

nelyasi:
Both snippets under “Quick example” currently fail with a TypeError.

Circuit.__init__ has the signature

(gates, system_size, parameters=(), initial_state=None)

so system_size is required, but neither example passes it. The equivalent examples under docs/content/docs/ already include it correctly; the README appears to be the only one that drifted, likely because it is not covered by just doctest-docs.

The comments also describe the gates as exp(-i θ/2 · H), but ExpGate applies exp(+i θ · H). The engine rotates by 2 · gen_coeff · param (LayerAngle in Evolution.h, and cos(2 * param) in Evolution.cpp:476), so both the sign and the factor in the comments were incorrect. I fixed the comments only; no code changes were needed.

I also fixed an unterminated </p tag left over from the benchmark images.


diagonal-hamiltonian:
Thank you for the PR, I will look at it tomorrow 🙂

robertodr:
Thanks for the PR! Noting to myself that we should configure the documentation CI lane to add the examples in the README to the `doctest`-s to run, so we avoid these drifts in the future.

Before we merge, can you touch up the PR description to adopt [our template](https://github.com/Algorithmiq/monoprop/blob/main/.github/PULL_REQUEST_TEMPLATE.md)? This is necessary to clarify that you've read and approved the contributor license agreement.

matteoacrossi (approved):

PR #226, opened 2026-08-13
refactor!: :recycle: drop non-type template parameter for number of modes

robertodr:

## Summary

Drops the `NumModes` non-type template parameter from the C++ engine. A `MonomialPropagator` used to be a template instantiated per compile-time mode ceiling, with a generated dispatch table and binder header picking the right instantiation for a given system size. Now there is **one compiled `MonomialPropagator`**: the logical width is a runtime constructor argument, and `detail::storage_modes_for()` sizes monomial storage from it (rounded up to a whole 32-mode block). `Bitset` becomes a runtime-width value type — the first 8 words inline, wider ones spilling to the heap — in place of a `Bitset<NumBits>` template.

Removing the compile-time width also removes the per-term speed that `if constexpr` gave for free, so this PR carries the machinery that earns it back at runtime: a bound-once-per-layer storage word count (`with_kernel_width`) that turns the per-term word loops back into a compile-time trip count, a per-gate kernel seam (`TermProductsFor`) between dense and sparse row representations, and a new sparse ("support-form") row store chosen per propagator from the storage width, alongside the existing dense `OperatorIndex` backend behind the shared `RowAccess.h` accessor seam.

## Changes

**Core engine**
- `MonomialPropagator` takes one runtime `num_modes` (no compile-time ceiling, no per-width instantiation); `logical_num_modes_`/`logical_num_modes()` renamed to `num_modes_`/`num_modes()`.
- `Bitset` has runtime-width: inline storage for the first 8 words, heap-spilled beyond that (>250 modes); per-word loops dispatch through `detail::with_nwords` instead of `if constexpr` arms.
- New sparse row store (`detail::SparseRowStore`, fixed-width mode lanes + a 2-bit `codes` word per row) as a third row-accessor backend beside `std::vector<Bitset>` and `OperatorIndex`, chosen per propagator by `SparseRowStore::preferred_for_modes()` from the target ISA (`monoprop_SPARSE_ROW_MIN_MODES`) and forceable process-wide with `monoprop_ROW_STORE=dense|sparse`. Backend is bound once per layer via `with_store`, never per term.
- `algebra/CodesAlgebra.h`: the structural algebra over a sparse row's `codes` word, one function per dense counterpart, plus `sparse_toggle` for the per-term product; falls back to the dense product on scratch-row overflow rather than truncating.
- `layer_build/TermProduct.h`: the per-term kernel seam (`TermProductsFor<Store, A, W>`), `SparseTermProducts`/`DenseTermProducts`/`DenseTermProductsW<A, W>`, and `with_kernel_width`, which binds the storage word count as a compile-time template parameter once per layer (`kNarrowKernelWords` caps which widths get an instantiation) and throws `KernelWidthMismatch` on a width mismatch instead of compiling it away.
- New query-record wire format (`QueryKeysFor`, `query_payload_words_for`) with a record-count header and a dense escape tail for fully-paired products that overflow a fixed-stride sparse record.
- Row store resizing in place (`OperatorIndex::resized()` / `SparseRowStore::resized()`) so a cutoff-widening setting change migrates existing rows instead of silently overflowing them.
- Generated dispatch table and binder header removed (`tools/generate-binders.py`, `tools/generate-dispatch.py`, `src/monoprop/bindings/bindings.cpp.in`); bindings are now a single, non-generated `src/monoprop/bindings/bindings.cpp`.

**Testing**
- Wide-system test regime: `ModeEmbedding`/`WIDE_EMBEDDING` (Python) and `test_utils::embed_case` (C++) relabel a narrow fixture's modes into a 260-logical/288-storage-mode system, so the fixture's exact energy/gradient still applies without checking in a wider fixture blob.
- New C++ suites: `sparse_row_store_tests.cpp`, `sparse_index_tests.cpp`, `sparse_wire_tests.cpp`, `codes_algebra_tests.cpp`, `codes_product_tests.cpp`, `term_product_tests.cpp`, `word_kernel_tests.cpp`, `row_store_selection_tests.cpp`, `store_interchange_tests.cpp`, `wide_system_tests.cpp` — the whole suite now also runs under `monoprop_ROW_STORE=sparse` (the `sparse-rows` CTest label), plus dedicated MPI CTest variants for the sparse wire format (`monoprop_MPI_SPARSE_ROWS_TEST_PROCS`).
- New Python tests: `tests/test_mode_width.py`, `tests/test_wide_system.py`, `tests/test_deep_circuit_gradient.py`; `tests/test_binding_layout.py` and `tools/_binding_layout.py` removed along with the generated binder they checked.
- `just diff-baseline-sparse` added for comparing the two row-store backends (tolerance-based, since they agree on term sets/values but not term order), distinct from the byte-wise `just diff-baseline`.

**Simplification passes** (no behavior change; verified byte-identical `capture-baseline` and tolerant `diff-baseline-sparse` agreement)
- Collapsed `Bitset`'s inline/spilled word-op arms into a shared `with_words_`/`apply_words_` dispatcher; shared row-materialization and reinsertion helpers across the two row-store backends via `RowHashTable` and `detail::geometric_row_capacity`; hoisted several loop-invariant computations out of per-term/per-layer hot paths.

**Docs**
- `AGENTS.md` rewritten to document the new runtime-width architecture (mode width, row-store seam, per-term kernel seam, query record, partition facade); `docs/content/docs/building.mdx`, `testing.mdx`, `benchmarks.mdx` updated to match.

## Checklist

## AI/LLM disclosure

- [x] I used the following tool to help write this PR description: Claude Code: claude-sonnet-5
- [x] I used the following tool to generate or modify code: Claude Code: claude-opus-5

> [!IMPORTANT]

> [!WARNING]


robertodr:
benchmarks against `main` for the Hubbard model

Read rows `<num_sites>_<cutoff>_<lower_atol>`

## Time

| Operation       | main     | refactor-drop-nttp | + ea64df5 | + 16d8ad8 |
|----------------|---------:|-------------------:|----------:|----------:|
| 32_6_6  | 52.871 s | 61.417 s | 56.467 s | 58.699 s |
| 32_8_5  | 15.661 s | 16.081 s | 15.739 s | 15.896 s |
| 64_6_6  | 61.750 s | 67.608 s | 62.350 s | 64.559 s |
| 64_8_5  | 16.498 s | 17.360 s | 16.546 s | 17.070 s |
| 124_6_6 | 81.586 s | 80.307 s | 78.376 s | 80.690 s |
| 124_8_5 | 19.680 s | 20.404 s | 20.032 s | 20.457 s |

## Memory (peak RSS)

| Operation       | main          | refactor-drop-nttp | + ea64df5 | + 16d8ad8 |
|----------------|--------------:|-------------------:|----------:|----------:|
| 32_6_6  | 18617.14 MiB | 19386.70 MiB | 18851.45 MiB | 17909.15 MiB |
| 32_8_5  | 6700.80 MiB  | 8945.71 MiB | 6786.73 MiB | 6465.30 MiB |
| 64_6_6  | 18929.92 MiB | 19462.99 MiB | 19348.63 MiB | 18270.70 MiB |
| 64_8_5  | 6766.88 MiB  | 9028.16 MiB | 6889.79 MiB | 6505.06 MiB |
| 124_6_6 | 19570.04 MiB | 19675.12 MiB | 19677.80 MiB | 19641.00 MiB |
| 124_8_5 | 8200.38 MiB  | 8295.32 MiB | 9005.57 MiB | 8217.71 MiB |


for the 124 num-sites case, there isn't a noticeable regression. But the smaller cases use much more memory than on `main`. Review the code to pin-point where the regression might come from. Come up with suggestions on how to fix it.

---

### The `+ ea64df5` column

The memory regression is the row store's slot width. `OperatorIndex` used to pick its position type
with a `conditional_t` on `2*NumModes`; dropping the NTTP left one fixed `uint16_t`, which doubles
the operator's largest array for every system at or below 128 modes — where 32 and 64 sites both
sit, and where 124 sites does not. That is exactly the shape of the table above. `ea64df5` sizes the
slot from the width at runtime instead, and the deterministic accounting comes back byte-for-byte:

| `operator_terms_bytes` | main | refactor-drop-nttp | + ea64df5 | + 16d8ad8 |
|-----------------------|-----:|-------------------:|----------:|----------:|
| 32_6_6  | 2222.22 MiB | 4444.44 MiB | 2222.22 MiB | 2222.22 MiB |
| 32_8_5  |  881.70 MiB | 1763.39 MiB |  881.70 MiB | 881.70 MiB |
| 64_6_6  | 2206.03 MiB | 4412.06 MiB | 2206.03 MiB | 2206.03 MiB |
| 64_8_5  |  891.50 MiB | 1783.00 MiB |  891.50 MiB | 891.50 MiB |
| 124_6_6 | 4413.62 MiB | 4413.62 MiB | 4413.62 MiB | 4413.62 MiB |
| 124_8_5 | 1759.98 MiB | 1759.98 MiB | 1759.98 MiB | 1759.98 MiB |

`operator_memory_breakdown().total_bytes` lands within 1.8 MiB of `main` at every config, and the
residual is the new `d_invidx_columns_bytes` accounting field rather than new memory. Peak RSS
improves by less than the operator does at the `_6_6` configs (32_6_6: 535 MiB of the 2222 MiB) because
the peak there is set by transient build buffers, not by the operator.

The commit also recovers most of the time regression, by binding the storage word count once per
layer beside the algebra and the row backend, so the scan's per-term word loops get a compile-time
trip count. Against the branch, down the rows: −8.1%, −2.1%, −7.8%, −4.7%, −2.4%, −1.8%. The `_6_6`
rows gain more than the `_8_5` ones because a looser `lower_atol` rejects a term on its coefficient
*before* the product is computed, so it shrinks the very share of the run this addresses.

#### Caveats on the numbers

The `+ ea64df5` times are means of `--model-rounds 5` (sd 0.25–1.47 s); the other two columns are the
original single samples, so differences below ~5% there are not resolvable. Peak RSS is
`--model-rounds 1` throughout — at 5 rounds it inflates by ~2.5 GiB, because a round's operator is not
freed before the next is built. Re-running `main` and the branch tip today reproduces both original
columns (`main` 124_8_5: 8203.68 vs 8200.38 MiB; branch 32_8_5: 8968.81 vs 8945.71 MiB), so the
August numbers are still comparable to today's.

#### One real regression: 124_8_5 peak RSS, +710 MiB

It is not live data. Resting RSS and every field of `operator_memory_breakdown()` are identical to the
branch. It is glibc per-thread arena retention provoked by a changed allocation interleaving: it
disappears at 8 threads (7288.84 vs 7278.01 MiB) and under
`GLIBC_TUNABLES=glibc.malloc.arena_max=2` (8204.86 MiB, i.e. `main`'s 8203.68). Building the `Bitset`
half of the commit alone reproduces the branch exactly (8283.64 vs 8282.25 MiB), so it is the
`OperatorIndex` half. Worth a follow-up `mallopt`/trim look — it is not the row-width change
reappearing.

All runs on the same host as the original two columns, default thread count,
`--hubbard-observable-site=18`.


### The `+ 16d8ad8` column

`16d8ad8` is the branch tip: `ea64df5` plus the two later branch commits (`9b43667`, `d37c408`) and
the merge of `origin/main` (`cb9a033`). Same host, same `--hubbard-observable-site=18`, dense row
store resolved at every config.

**Memory is now at or below `main` everywhere.** Peak RSS against `main`, down the rows: −708, −235,
−659, −262, +71, +17 MiB. In particular the `124_8_5` +710 MiB peak documented in the section above
is gone (+17 MiB, i.e. `main`). `operator_terms_bytes` is unchanged from the `ea64df5` column, so the
row-slot fix is intact.

What buys it is `main`'s #259, which halves the follower-marking epoch stamp. Resting RSS drops
886 / 411 / 915 / 533 MiB at the four configs at or below 128 modes, and the stamp array is
710 MiB at 260M terms — it was twice that before. `operator_memory_breakdown().total_bytes` *rises*
by exactly that array at every config (+710.22, +253.21, +704.80, +265.74, +731.93, +265.41 MiB)
because #259 also gave it an accounting field, `matched_scratch_bytes`: the memory was always there,
it just had no name. #268's `init_op_map` release contributes nothing on this workload
(`init_operator_bytes` is already 0 on both sides — the branch had its own release, and the merge
kept `main`'s mechanism).

**Time is unchanged.** The column reads 2–4% above `+ ea64df5`, but that is the host, not the code:
re-running `ea64df5` itself today reproduces the same offset (58.789 / 15.955 / 64.924 / 17.310 /
81.728 / 21.091 s against the August 56.467 / 15.739 / 62.350 / 16.546 / 78.376 / 20.032 s). Measured
against that same-day `ea64df5`, `16d8ad8` is −0.2%, −0.4%, −0.6%, −1.4%, −1.3%, −3.0% — flat to
slightly faster. The merge alone, against a same-day `d37c408`, is +0.6%, −0.2%, −0.5%, −4.4%,
−2.1%, −4.2%.

The merge is numerically inert on this workload too: the evolved term count is identical to the
`ea64df5` column at every config (260746764 / 95752255 / 260762912 / 95752255 / 260762912 /
95752255).

#### Caveats

Times are `--model-rounds 5` means, sd 0.45–2.50 s, so a single row's 2–4% is not resolvable — the
attribution above rests on the sign being consistent across all six. Peak RSS is `--model-rounds 1`.
The `main` and `refactor-drop-nttp` columns remain the original single samples.

One fix was needed to run any of this: the merge left `benches/conftest.py` recording
`monoprop.MAX_NUM_MODES`, an attribute this branch removed, which aborted every benchmark run at
`pytest_configure`. Fixed in `c5b4f8d`.

cpp/monoprop/detail/operator/SparseRowStore.h
```diff
@@ -0,0 +1,647 @@
...
+    [[nodiscard]] auto indexed_count() const noexcept -> size_t { return table_.count(); }
+
+    [[nodiscard]] auto index_estimated_memory_bytes() const -> size_t {
+        return sizeof(SparseRowStore) + table_.slot_bytes();
+    }
+
+    [[nodiscard]] auto memory_bytes() const -> size_t {
+        size_t total = modes_.capacity() * sizeof(ModeT);
+        total += codes_.capacity() * sizeof(CodesT);
+        total += overflow_.size() * (sizeof(value_type) + sizeof(size_t) + 24);
```
robertodr: :robot: _AI text below_ :robot:

`memory_bytes()` charges each spilled row `sizeof(Bitset) + sizeof(size_t) + 24` but never adds what the `Bitset` owns off-object. AGENTS.md (this branch) states the rule: *"memory accounting over a container of monomials must add `Bitset::heap_bytes()` per element; `sizeof(Bitset)` per element looks right until someone runs a wide system."*

That is exactly the regime this branch adds: at 288 storage modes a monomial is 9 words, past `Bitset::kInlineWords`, so every `overflow_` entry heap-allocates 72 bytes that `operator_memory_breakdown()["operator_terms_bytes"]` silently drops. `monomial_map_bytes()` right next door already does this correctly for `init_op_map`. `OperatorIndex::memory_bytes()` has the same gap.

```suggestion
    [[nodiscard]] auto memory_bytes() const -> size_t {
        size_t total = modes_.capacity() * sizeof(ModeT);
        total += codes_.capacity() * sizeof(CodesT);
        for (const auto &kv : overflow_) {
            total += sizeof(value_type) + sizeof(size_t) + 24 + kv.second.heap_bytes();
        }
        return total;
    }
```

AGENTS.md
```diff
@@ -36,20 +36,21 @@ monoprop is a high-performance C++/Python hybrid library implementing Majorana a
...
-  per-mode-width `bindings.cpp` and `_dispatch.py` from it (do not hand-edit the generated files).
-  Both generators take the 32-mode storage-block rule from `tools/_binding_layout.py` — they must
-  agree, or dispatch routes at a template the bindings never instantiated.
+- `cpp/include/monoprop/MonomialPropagator.h`: the single C++ engine `MonomialPropagator` (the
+  Majorana/Pauli choice is a runtime `Basis`, not a separate class; so is the mode count).
+  `storage_modes_for()` is the storage-width rule: a logical width rounds up to a whole 32-mode block,
+  never below one block. Pass `storage_num_modes` to override it — the C++ tests do, via
+  `cpp/tests/TestPropagator.h`, because the width is part of a monomial's hash and hence of the order
+  coefficients accumulate in.
+- `src/monoprop/bindings/binder.h`: the binding for that one class, called from
```
robertodr: :robot: _AI text below_ :robot:

`src/monoprop/bindings/binder.h` no longer exists — commit `70d839c` ("remove binder.h") folded it into `src/monoprop/bindings/bindings.cpp`. Line 214 ("Add Python bindings in `src/monoprop/bindings/binder.h`") is stale for the same reason.

The repo's own Documentation Maintenance Policy in this file says to update `AGENTS.md` in the same change and to fix or remove any section that no longer reflects the codebase.

cpp/monoprop/detail/evolution/layer_build/Common.h
```diff
@@ -122,45 +157,384 @@ inline auto decode_value(size_t word) -> double {
...
+    // holds exactly that many lanes. num_bits sizes the escape monomials.
+    auto configure(size_t num_bits, size_t capacity) -> void {
+        if (capacity_ != capacity || num_bits_ != num_bits) {
+            lanes_.clear();
+            keys_.clear();
+            escapes_.clear();
+            retained_lanes_.clear();
+            retained_.clear();
+            capacity_ = capacity;
+            num_bits_ = num_bits;
```
robertodr, line 402: :robot: _AI text below_ :robot:

`configure()` clears `lanes_`, `keys_`, `escapes_`, `retained_lanes_` and `retained_`, but **not** `retained_bases_` or `retained_escapes_`.

`retain()` derives its handle from `retained_.size()` while pushing the matching base onto `retained_bases_`. If `retained_` is cleared and `retained_bases_` is not, the two arrays desynchronise: handle `0` then reads `retained_bases_[0]`, a stale base into a `retained_lanes_` that was just emptied, and `retained()` indexes out of bounds. `retained_escapes_` has the same pairing problem plus unbounded growth.

Today this is latent — the engine's `keys_` is constructed fresh per `build_layer` and the `thread_local` probe batch in `Resolve.h` never calls `retain()` — but the invariant is one `configure()` call away from being live.

```suggestion
        if (capacity_ != capacity || num_bits_ != num_bits) {
            lanes_.clear();
            keys_.clear();
            escapes_.clear();
            retained_lanes_.clear();
            retained_bases_.clear();
            retained_escapes_.clear();
            retained_.clear();
            capacity_ = capacity;
            num_bits_ = num_bits;
        }
```

cpp/monoprop/detail/monomial_propagator/MonomialPropagator.cpp
```diff
@@ -119,65 +94,94 @@ MonomialPropagator<NumModes>::MonomialPropagator(const OperatorDict &initial_ope
...
+    // width, which is only knowable after regenerate_cutoff_fn_() above. set_store() drops the stale lazy
+    // inverted index with it.
+    //
+    // Which backend: the crossover is on the *storage* width, not the logical one, because what the dense
+    // representation costs is one pass per storage word. Rows are sized from the cutoff -- a surviving
+    // term occupies at most `cutoff` modes under either cutoff kind -- and anything wider (a fully paired
+    // term, which escapes the cutoff) spills losslessly.
+    if (use_sparse_rows_()) {
+        mp_op_.set_store(std::make_unique<detail::SparseRowStore>(2 * storage_num_modes_,
+                                                                  detail::SparseRowStore::slots_for_bound(cutoff_)));
```
robertodr: :robot: _AI text below_ :robot:

The dense arm below sizes its rows through `packed_inline_width_()`, which special-cases Schrödinger (`if (schrodinger_) return kDefault;`). The sparse arm bypasses that rule entirely and sizes from `cutoff_` alone.

In Schrödinger the initial term set is every fully paired monomial with up to `max_pairs = ceil(sc/2)` occupied modes, and `sc` is `schrodinger_cutoff` — an independent user knob, not `cutoff`. Whenever `ceil(sc/2) > cutoff_` (e.g. `cutoff=1`, or any explicitly large `schrodinger_cutoff`), every one of those initial rows exceeds `slots_per_row_` and lands in the `overflow_` `unordered_map` — a heap `Bitset` plus a map lookup on the per-term path, for what should be the common case.

The two backends should read the row width from one place; `packed_inline_width_()` already is that place for the dense store.

CMakeLists.txt
```diff
@@ -99,8 +125,21 @@ message(STATUS "   User-appended          : ${EXTRA_CXXFLAGS}")
...
+  "   Sparse rows from       : ${monoprop_SPARSE_ROW_MIN_MODES} modes"
+)
+
+# Testing is enabled from the *top-level* list file on purpose: CTest's root is wherever
+# enable_testing() was called, so called from cpp/ it wrote no top-level CTestTestfile.cmake and every
+# documented entry point (the CMakePresets test presets, `just test-mpi`, `just test-wide`) pointed
+# ctest at a directory with no tests. ctest reports "No tests were found" and exits 0 for that, so those
+# commands were silently running nothing.
+if(monoprop_ENABLE_CXX_UNIT_TESTS)
+  enable_testing()
```
robertodr: :robot: _AI text below_ :robot:

This block duplicates the `if(monoprop_ENABLE_CXX_UNIT_TESTS) enable_testing() / include(CTest) endif()` that already sits at lines 152-158, immediately before `add_subdirectory(cpp)` (added on main by #224). Both are now in the file, each carrying a comment that claims to be *the* reason the call lives where it does — and they give different reasons.

Harmless at build time (`CTestTargets.cmake` guards its own targets with the `CTEST_TARGETS_ADDED` global property), but the next reader has to work out which comment is true. Keep one.

pyproject.toml
```diff
@@ -124,7 +124,7 @@ sdist.exclude = [
   ".prettierignore",
   "cspell.json",
 ]
-sdist.include = ["src/monoprop/_version.py", "src/monoprop/_dispatch.py"]
+sdist.include = ["src/monoprop/_version.py"]
```
robertodr: :robot: _AI text below_ :robot:

Related cleanup this PR misses (line 89, not reachable as an inline anchor):

```toml
[tool.uv]
cache-keys = [
  ...
  { file = "src/monoprop/bindings/bindings.cpp.in" },   # <- deleted by this PR
]
```

This PR deletes `src/monoprop/bindings/bindings.cpp.in` (it became `src/monoprop/bindings/bindings.cpp`), but the cache key still points at the removed template, and no remaining entry covers `src/monoprop/bindings/**` — `cpp/include/**`, `cpp/monoprop/**`, `**/CMakeLists.txt` and `cmake/**` all miss it.

So editing the nanobind bindings no longer invalidates uv's build cache: a later `uv sync` can hand back a wheel built from the previous `bindings.cpp`. Suggest `{ file = "src/monoprop/bindings/*.cpp" }`.

cpp/monoprop/detail/partition/PartitionGroup.h
```diff
@@ -38,17 +38,20 @@
 
 namespace monoprop {
 
-template <size_t NumModes>
-class MonomialPropagator; // completed before any PartitionGroup member body is instantiated (Impl.h)
+// Only MonomialPropagator.inl includes this header, and it does so *after* MonomialPropagator's
```
robertodr: :robot: _AI text below_ :robot:

This comment is written against `MonomialPropagator.inl`, but the same PR renames that file to `cpp/monoprop/detail/monomial_propagator/MonomialPropagator.cpp`. The include is now from the `.cpp`, and the ordering contract the comment describes (include *after* `MonomialPropagator`'s definition, because the member bodies are ordinary functions needing the complete type) is what a reader is meant to be able to check — against a filename that no longer exists.

src/monoprop/bindings/bindings.cpp
```diff
@@ -94,13 +135,14 @@ auto bind_monomial_propagator(nb::module_ &mod) -> void {
         "upper_atol"_a = std::nullopt,
         "cutoff_type"_a = "length",
         "basis_change"_a = std::nullopt,
-        "logical_num_modes"_a = NumModes,
+        "logical_num_modes"_a,
```
robertodr: :robot: _AI text below_ :robot:

`logical_num_modes` has no default but is declared *after* five defaulted arguments (`comm`, `schrodinger_cutoff`, `lower_atol`, `upper_atol`, `cutoff_type`, `basis_change`). nanobind builds the signature from these in order, so the docstring/stub signature is `(..., basis_change=None, logical_num_modes, basis="majorana", ...)` — not valid Python, and any positional call has to pass all six preceding arguments.

`src/monoprop/monomial_propagator.py` only ever calls this with keywords, so nothing breaks today, but moving `logical_num_modes` up next to `initial_state` (matching the C++ constructor, where it already sits third) would make the exposed signature honest.

cpp/monoprop/detail/operator/SparseRowStore.h
```diff
@@ -0,0 +1,647 @@
...
+            [this](const Key &key) { return fold_hash(key); },
+            [this](size_t i) {
+                __builtin_prefetch(&codes_[i], 0, 0);
+                __builtin_prefetch(&modes_[i * slots_per_row_], 0, 0);
+            },
+            [this](size_t i, const Key &key) { return row_eq_key(i, key); });
+    }
+
+    // Rows in index order, as fn(row_index). Not the row itself: a spilled row has no view, so what a
+    // caller wants off the index is the index.
```
robertodr: :robot: _AI text below_ :robot:

"Rows in index order" is not what this does — `RowHashTable::for_each_slot` walks the *slot* array, so rows come out in hash/probe order, not ascending row index. `OperatorIndex::for_each` and `MPOperator::for_each_term` both describe it correctly as "table order" / "slot order", and the class comment 300 lines up is explicit that the two stores deliberately differ in that order.

Given this order is load-bearing (it fixes the user-visible evolved-term order and the FP accumulation order), a comment that says "index order" is the kind that gets believed.

cpp/monoprop/detail/evolution/layer_build/Common.h
```diff
@@ -122,45 +157,384 @@ inline auto decode_value(size_t word) -> double {
...
+}
+
+// Owned storage for a batch of query keys in whichever form a store keys its rows by, plus the record
+// reader that fills it. Both resolve paths -- the self-resolve batch in Engine.h and the incoming probe in
+// Resolve.h -- want exactly this, and both used to hand-roll it: allocate once, keep the storage across
+// layers, overwrite every element before reading it, and hand the whole run to find_batch contiguously.
+//
+// Grow-only and never cleared between layers. That is the measured shape (see Resolve.h): an element is
+// overwritten whole before any read, so a per-layer rebuild bought nothing and cost a construction per
+// query. The trade is that the buffer holds the largest layer's worth until the thread exits; peak RSS is
```
robertodr: :robot: _AI text below_ :robot:

"Grow-only and never cleared between layers ... the buffer holds the largest layer's worth until the thread exits" is true of the `thread_local` batch in `Resolve.h`, but not of the other user of this class: `LayerBuildEngine::keys_` is a plain member of an engine that is constructed fresh inside `build_layer`, i.e. once per layer. Its `configure()` therefore always sees a default-constructed object and `ensure(kResolveBatch)` re-builds 64 elements every layer — the allocation this comment says was measured away.

(It has to be per-layer, because `retain()` handles index into `retained_`, which would otherwise grow without bound across layers — so the fix is the comment, not the lifetime.) `Resolve.h` line ~110 repeats the claim ("the same trade `nz` and `keys_` already make").

cpp/monoprop/detail/monomial_propagator/MonomialPropagator.cpp
```diff
@@ -119,65 +94,94 @@ MonomialPropagator<NumModes>::MonomialPropagator(const OperatorDict &initial_ope
...
-    mp_op_.inverted_index_.reset();
+    // Replaces the store MPOperator's constructor made: same width, but now with the cutoff-derived row
+    // width, which is only knowable after regenerate_cutoff_fn_() above. set_store() drops the stale lazy
+    // inverted index with it.
+    //
+    // Which backend: the crossover is on the *storage* width, not the logical one, because what the dense
+    // representation costs is one pass per storage word. Rows are sized from the cutoff -- a surviving
+    // term occupies at most `cutoff` modes under either cutoff kind -- and anything wider (a fully paired
+    // term, which escapes the cutoff) spills losslessly.
+    if (use_sparse_rows_()) {
```
robertodr: :robot: _AI text below_ :robot:

The store is created once here with a row width derived from `cutoff_`, but `cutoff_` is mutable for the propagator's lifetime: `update_cutoff()` (MonomialPropagator.h) calls `regenerate_cutoff_fn_()` and nothing else, so the store keeps `slots_per_row_ = slots_for_bound(old_cutoff)`.

Raising the cutoff then pushes every row wider than the old bound into `overflow_` — a heap `Bitset` plus an `unordered_map` lookup on the per-term path, for the *typical* row rather than the ~0.07% the spill was sized for. Still correct (the spill is lossless), but it is a silent order-of-magnitude cliff with no diagnostic; `partition_equivalence_tests.cpp` exercises `update_cutoff()` today, just below the sparse crossover where it cannot show up.

Worth either re-creating the store on a cutoff change or documenting that `slots_per_row_` is a construction-time bound.

cpp/monoprop/detail/operator/OperatorIndex.h
```diff
@@ -176,195 +193,65 @@ class OperatorIndex {
...
-        for (;; s = (s + 1) & table_.mask) {
-            const Slot &e = table_.slots[s];
-            if (e.idx == kEmptySlot) {
-                return std::nullopt;
-            }
-            if (e.h == h && row_eq_key(static_cast<size_t>(e.idx), key)) {
-                return static_cast<size_t>(e.idx);
-            }
-        }
+        return table_.find(fold_hash(key), [&](size_t i) { return row_eq_key(i, key); });
```
robertodr: :robot: _AI text below_ :robot:

Note on `memory_bytes()` just above (line 189-193, outside the diff so not directly anchorable):

```cpp
total += overflow_.size() * (sizeof(value_type) + sizeof(size_t) + 24);
```

`value_type` is now a runtime-width `Bitset`, and above `Bitset::kInlineWords` (8 words / 250 modes) each spilled row owns a heap array this never counts. The line was correct when `value_type` was `Monomial<NumModes>`, a fixed-size type with no allocation — de-templating `Bitset` is what invalidated it. `SparseRowStore::memory_bytes()` copies the same shape.

AGENTS.md (added in this PR) states the rule: *"memory accounting over a container of monomials must add `Bitset::heap_bytes()` per element; `sizeof(Bitset)` per element looks right until someone runs a wide system."* And 288 storage modes (9 words) is precisely the regime `tests/test_wide_system.py` adds here. `monomial_map_bytes()` in `MPOperator.h` already does it correctly.

cpp/tests/boostAddTests.cmake
```diff
@@ -206,6 +206,23 @@ foreach(LINE ${LINES})
...
+        "${TEST_EXECUTABLE}"
+        "--run_test=${test}"
+        "--report_level=detailed"
+        "--catch_system_errors=yes"
+        ${extra_args}
+      LABELS
+        serial
+        sparse-rows
+      ENVIRONMENT
+        "monoprop_ROW_STORE=sparse"
```
Panadestein, line 257: Nice that the sparse backend gets a per-case second run here! 

It looks like this env var never reaches the MPI variants, though. The `register_variant` at L253–262 sets only the OMPI root vars, `just test-sparse-rows` runs `pytest -m "not mpi"`, and `sparse_wire_tests.cpp` exercises the codec in-process with no communicator. So the sparse backend is currently covered serially but never across ranks.

That's a bigger gap than it looks, because the sparse form isn't just local storage; it changes the wire format. `query_payload_words_for` gives a different stride per backend, and the escape tail, `kOverflowLane` and `append_escape_tail` have no dense counterpart, so that whole path only ever runs single-rank. It's also the backend that ships for wide systems, which is exactly where MPI gets used.

Could we add `monoprop_ROW_STORE=sparse` MPI variants, even just at 2 ranks? Happy for it to be a follow-up if there's a reason to defer.

cpp/monoprop/Bitset.h
```diff
@@ -16,213 +16,573 @@
...
-    [[nodiscard]] constexpr auto none() const noexcept -> bool { return !any(); }
-
-    [[nodiscard]] static constexpr auto size() noexcept -> size_t { return NumBits; }
+    [[nodiscard]] auto none() const noexcept -> bool { return !any(); }
 
+    // Every binary op below loops *this*'s word count and indexes the other operand unchecked, so a
+    // narrower operand is read past its own width. That is not merely wrong-but-harmless: it only
+    // reads zeros from the inline array while the *result* fits inline, and once *this* is spilled
+    // (> kInlineWords) it reads off the end of the narrower operand's array. Widths must match at
+    // every call site, so this is asserted rather than handled -- Release keeps the loops bare.
```
Panadestein, line 379: The comment here very honest, and I agree it's the right call for the
inner loop. One consequence worth spelling out, since the comment frames it as a call-site obligation: on `main` this was `template <size_t NumBits> class Bitset`, so a width mismatch was a **compile error**. Now it's `assert` only, which compiles out under `NDEBUG`, so in Release a mismatch is a silent OOB read of the inline `std::array<word_type, 8>`, and an OOB read of a heap allocation once the operand has spilled. No crash, no diagnostic, just a wrong number.

The same applies to L282, L344, L361, L378 and L548–549.

I couldn't find a call site that actually violates this, so this isn't a bug report, just an observation. Would a width assert at those *construction* boundaries be worth it?

robertodr:

## Hubbard sweep: the win is the sparse row store, not the runtime width

**Provenance, first.** These numbers come from local commit `7cddf0e`, which is **not** this PR's
current head. Since the run the branch was force-pushed to `49a0a90` — the same work rebased onto a
newer `main` (so onto a later merge-base than the `main@97f95f7` used as the baseline here), plus the
`RowAccess.h` extraction, `9356eb5` and `74b89a1`. The decomposition below is structural and very
unlikely to invert, but treat the exact percentages as indicative rather than final. Say the word and
I'll re-run against `49a0a90`.

32 runs — 2 branches × 2 `monoprop_ENABLE_ARCH_FLAGS` settings × 4 widths × 2 cutoff settings, 5
rounds each — plus 2 runs with `monoprop_ROW_STORE` forced. Term counts match `main` **exactly** in
all 16 cells, so nothing is being traded for accuracy.

### The refactor itself is a uniform small regression

Forcing the row backend on the arch build separates the two things this PR changes at once:

| cell | main (dense) | PR dense | PR sparse | refactor<br>PR dense ÷ main | sparse gain<br>sparse ÷ PR dense | net<br>best PR ÷ main |
|---|---:|---:|---:|---:|---:|---:|
| 256 modes · 124 sites · c6 | 77.85 s | 78.97 s _(auto)_ | 72.57 s _(forced)_ | 0.99× | **1.09×** | 1.07× |
| 1600 modes · 800 sites · c6 | 239.09 s | 245.77 s _(forced)_ | 81.69 s _(auto)_ | 0.97× | **3.01×** | **2.93×** |

Dense-vs-dense at 1600 modes the PR is **2.7% slower and 14% heavier** (31.2 GiB peak RSS against
main's 27.3 GiB) — consistent with the 4–9% it gives up at 64–128 modes. The sparse store then buys
3.01× and 23% less memory over that same dense path, netting 2.93× against main. `main` has no sparse
row store to compare against, so that cell is branch-vs-branch, not backend-vs-backend.

### The matrix

Speedup = main mean ÷ PR mean over 5 rounds. Bold where outside round-to-round spread.

| storage modes | sites | cutoff / lower_atol | ARCH ON | ARCH OFF |
|---:|---:|---|---:|---:|
| 64 | 32 | 6 / 1e-6 | **0.94×** | **0.92×** |
| 64 | 32 | 8 / 1e-5 | **0.93×** | **0.93×** |
| 128 | 64 | 6 / 1e-6 | **0.96×** | **0.96×** |
| 128 | 64 | 8 / 1e-5 | **0.95×** | **0.94×** |
| 256 | 124 | 6 / 1e-6 | 0.99× | **1.09×** _(sparse)_ |
| 256 | 124 | 8 / 1e-5 | 0.97× | **0.94×** _(sparse)_ |
| 1600 | 800 | 6 / 1e-6 | **2.93×** _(sparse)_ | **2.94×** _(sparse)_ |
| 1600 | 800 | 8 / 1e-5 | **1.71×** _(sparse)_ | **1.72×** _(sparse)_ |

### Two things worth acting on

**1. `ENABLE_ARCH_FLAGS` buys nothing directly, on either branch.** Every ON/OFF pair sits inside
run-to-run spread, and four of the eight `main` cells are nominally *slower* with the flag on. This is
not the flag failing to apply: `-march=native` resolves to `znver4` with AVX-512 including
`avx512vpopcntdq`.

**2. The flag's only measurable effect is moving the sparse crossover 256 → 768 modes, and that costs
8%.** At 256 storage modes the arch build picks dense and lands 1.4% behind main; forcing sparse on
that same build gives 72.57 s vs 78.97 s — a 6.8% win over main instead. Since the sparse store is
where *all* the gain lives, `monoprop_SPARSE_ROW_MIN_MODES` is the highest-leverage constant in this
PR, and 768 looks too conservative on znver4. The noarch build's 256 appears to be the better number
on this machine.

<details>
<summary><b>Methodology, environment, and per-round spread</b></summary>

### Builds

Four editable builds from four git worktrees, `monoprop_ENABLE_CXX_UNIT_TESTS=OFF`:

| config | commit | cmake |
|---|---|---|
| `main-arch` | `97f95f7` | `monoprop_MAX_NUM_MODES=2000`, `ENABLE_ARCH_FLAGS=ON` |
| `main-noarch` | `97f95f7` | `monoprop_MAX_NUM_MODES=2000`, `ENABLE_ARCH_FLAGS=OFF` |
| `pr-arch` | `7cddf0e` | `ENABLE_ARCH_FLAGS=ON` (sparse crossover 768) |
| `pr-noarch` | `7cddf0e` | `ENABLE_ARCH_FLAGS=OFF` (sparse crossover 256) |

`main` needs `monoprop_MAX_NUM_MODES=2000` to reach 1600 modes at all; that instantiates 63
width-specialised propagators (32…2016), and 800 sites dispatches to `MonomialPropagator1600`. Both
branches round storage up to a whole 32-mode block, so every cell compares **equal storage widths**
(124 sites = 248 logical → 256 storage).

### Harness

The **PR's** benchmark harness was rsync'd into both `main` worktrees (`benches/` and
`packages/monoprop-bench-tools/src/monoprop_bench_tools/`) so that only the compiled extension
differs between the two sides. This was necessary, not cosmetic:

- `main`'s `conftest.py` has no `--model-rounds` option — it hardcodes `rounds=1`, so `main` would
  have contributed a single sample against the PR's 5.
- `memory/cpu.py` differs by ~115 lines (the PSS sampler), i.e. the measurement itself.

One line was patched for `main`, which has no `rows_are_sparse` binding: `_record_row_store` records
`"n/a"` instead of raising.

### Invocation

```
monoprop_BENCH_LABEL=<label> monoprop_BENCH_RESULTS=<dir> \
  pytest "benches/bench_models.py::test_model[hubbard]" \
  --hubbard-num-sites=<S> --hubbard-cutoff=<C> --hubbard-lower-atol=<A> \
  --hubbard-observable-site=18 --model-rounds 5
```

`trotter_steps` left at its default of 29. Runs were strictly sequential — never concurrent — so no
two runs contended for cores.

### Environment

AMD EPYC 9R14 (Zen 4), 64 physical cores, SMT off, 123 GiB RAM. Single rank, no MPI, default thread
count → 64 partitions pinned 1:1 to cores. GCC, `-O3 -DNDEBUG`. Python 3.11.15, nanobind 3.0.0.

### Per-round spread (mean ± stddev, seconds, n=5)

| modes | c | main-arch | pr-arch | main-noarch | pr-noarch |
|---:|---:|---:|---:|---:|---:|
| 64 | 6 | 52.95 ± 1.75 | 56.42 ± 2.50 | 52.25 ± 1.65 | 56.74 ± 2.13 |
| 64 | 8 | 14.34 ± 0.32 | 15.40 ± 0.72 | 14.21 ± 0.24 | 15.24 ± 0.69 |
| 128 | 6 | 59.90 ± 0.70 | 62.16 ± 1.45 | 60.09 ± 0.78 | 62.69 ± 1.32 |
| 128 | 8 | 15.64 ± 0.28 | 16.51 ± 0.54 | 15.68 ± 0.26 | 16.69 ± 0.59 |
| 256 | 6 | 77.85 ± 0.49 | 78.97 ± 2.24 | 77.08 ± 0.43 | 70.55 ± 2.62 |
| 256 | 8 | 19.30 ± 0.30 | 19.88 ± 0.54 | 18.70 ± 0.32 | 19.86 ± 0.72 |
| 1600 | 6 | 239.09 ± 2.12 | 81.69 ± 3.95 | 243.33 ± 1.52 | 82.86 ± 3.92 |
| 1600 | 8 | 48.99 ± 0.45 | 28.67 ± 0.66 | 50.06 ± 0.41 | 29.08 ± 0.72 |

Forced-backend runs: `pr-arch` 256 modes c6 sparse = 72.57 ± 2.73 s; `pr-arch` 1600 modes c6 dense =
245.77 ± 3.30 s.

### Peak RSS (MiB, process high-water mark over the whole benchmark, construction included)

| modes | c | main-arch | pr-arch | main-noarch | pr-noarch |
|---:|---:|---:|---:|---:|---:|
| 64 | 6 | 19,907 | 20,362 | 19,801 | 20,315 |
| 64 | 8 | 8,808 | 8,830 | 8,769 | 8,758 |
| 128 | 6 | 20,145 | 20,610 | 20,124 | 20,692 |
| 128 | 8 | 8,739 | 8,936 | 8,727 | 8,934 |
| 256 | 6 | 22,293 | 22,870 | 22,100 | 23,927 |
| 256 | 8 | 12,430 | 12,813 | 12,445 | 12,480 |
| 1600 | 6 | 27,277 | 24,051 | 27,588 | 24,105 |
| 1600 | 8 | 12,379 | 11,659 | 12,532 | 11,623 |

### Term counts (identical across all four configs)

cutoff 6: 260,746,764 at 64 modes; 260,762,912 at 128 / 256 / 1600 modes. cutoff 8: 95,752,255 at
every width.

### Caveats

- Single host, single rank, default thread count — no MPI sweep and no thread sweep.
- Timings are `test_model[hubbard]` only: 29 propagate steps plus one `expectation_value()`,
  excluding model construction. Peak RSS *includes* construction, so the two are not directly
  divisible.
- The forced-backend runs were done on the arch build only and at cutoff 6 only, so the
  dense-vs-dense attribution is established at two widths, not four.
- A 1–2% difference sits inside the round-to-round spread at these sizes.

</details>

robertodr:
closing: it's been unpacked into a stack of changes instead, for easier review

PR #227, opened 2026-08-14
refactor(bench): :package: release the benchmark harness as monoprop-bench-tools

Panadestein:

Turns the repository into a [uv workspace](https://docs.astral.sh/uv/concepts/projects/workspaces/) and splits the benchmark harness in two: the reusable half becomes a package we can publish, the benchmarks stay here.

```
monoprop/
├── pyproject.toml            # the library (unchanged in scope)
├── uv.lock                   # one lock for the whole workspace
├── src/monoprop/
├── benches/                  # monoprop's own benchmark suite — stays put
└── packages/
    ├── monoprop-bench-tools/ # → PyPI
    └── bench-third-party/    # → GitHub only, standalone uv project
```

## What moved

| Was | Is |
| --- | --- |
| `benches/_memory_cpu.py` | `monoprop_bench_tools.memory.cpu` |
| `benches/_memory_gpu.py` | `monoprop_bench_tools.memory.gpu` |
| `benches/_builders.py` | `monoprop_bench_tools.models` |
| `benches/report.py` | `monoprop_bench_tools.report` → `monoprop-bench-report` |
| `benches/bmf.py` | `monoprop_bench_tools.bmf` → `monoprop-bench-bmf` |
| `tests/test_bench_*.py` | `packages/monoprop-bench-tools/tests/` |
| `benches/third_party/` | `packages/bench-third-party/` |

All moves are `git mv`, so `git log --follow` still works.

## Design decisions worth reviewing

**`benches/` does not move.** Bencher keys its metric history on the *benchmark name*, so moving or renaming a `bench_*` test orphans its tracked series and resets the thresholds' sample windows. The suite therefore cannot version with a library release, which is precisely the line between "harness" and "benchmarks". `monoprop-bench-tools` contains no benchmarks.

**The bencher.dev tracking stays in the repo.** Only the *renderer* (`bmf.py`) is tooling and moved. `bench_main.yml` carries `BENCHER_API_KEY`, `vars.BENCHER_PROJECT`, the `--branch`/`--testbed` identity and the four threshold blocks — that is this repository's CI identity, and a published package cannot own a secret. The workflow itself is unchanged; `just bench-bmf` now shells out to the console script.

**Three dependency groups instead of two.**

- `test` — monoprop's suite only, deliberately free of workspace members. cibuildwheel installs it against a *built wheel* via `test-groups = ["test"]`, where a `monoprop-bench-tools` requirement would resolve from PyPI (nonexistent) and drag monoprop back in from source.
- `workspace-test` — `test` + the member. What `uv run pytest` at the root needs.
- `bench` — `workspace-test` + `pytest-benchmark`.

**`bench-third-party` is `exclude`d from the workspace.** It pins a narrower `requires-python`, a git dependency and linux-x86_64-only CUDA wheels; folding it in would drag all of that into the root lock. It stays a standalone uv project with its own lockfile (`cd packages/bench-third-party && uv sync`). It now depends on `monoprop-bench-tools` properly instead of poking `benches/` onto `sys.path`.

**No `gpu` extra on the tools package.** Adding one put `cupy-cuda12x` + `cuda-pathfinder` into the root lock. CuPy is installed separately; `memory.gpu` already degrades to `Method.UNAVAILABLE` without it.

## Publishing

`deploy.yml` gains `build-bench-tools` (`uv build --package monoprop-bench-tools` — pure Python, no cibuildwheel matrix) and a **separate** `upload_pypi_bench_tools` job. Separate because it is a separate PyPI project, and because a failure there must not hold back the library's own upload. hatch-vcs reads the same git tags via `raw-options = { root = "../.." }`, so the two distributions share a version line; verified that the wheel still builds from the sdist, where there is no git.

> [!IMPORTANT]
> Before the next release someone has to register a **trusted publisher** for `monoprop-bench-tools` on PyPI (this workflow file, the `pypi` environment), or `upload_pypi_bench_tools` will fail on its first run.

## Validation

- `uv lock --check`, `uv sync --all-groups --all-extras`
- `pytest -m "not mpi"` → 617 passed (both suites)
- `just bench-smoke` and `just bench-bmf smoke` end to end through the console scripts
- Both distributions build; `packages/bench-third-party/uv.lock` re-resolved
- `prek run --all-files` clean

## Deliberate non-changes

- `packages/**` is excluded from Sonar scope, preserving current behaviour. Bringing it in is a separate call.
- `benches/results/*.json` is untouched, so Bencher continuity holds.

## One seam this creates

`benches/conftest.py` *writes* the results artifact schema and `monoprop_bench_tools.report`/`.bmf` *read* it — now across a package boundary. Called out in both docstrings and in `AGENTS.md`: a change to the recorded sections has to land on both sides.

Docs updated in the same change: `AGENTS.md` (new "Workspace layout" section), `README.md`, `docs/content/docs/benchmarks.mdx`, `docs/content/docs/testing.mdx`, `benches/README.md`.


diagonal-hamiltonian:
Good idea 👍

packages/monoprop-bench-tools/src/monoprop_bench_tools/memory/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+"""Peak-memory accounting for benchmarks.
+
+Two backends with a deliberately parallel surface --
+:mod:`monoprop_bench_tools.memory.cpu` for the host (kernel ``VmHWM``) and
+:mod:`monoprop_bench_tools.memory.gpu` for the device (CUDA memory-pool
+counters). Neither is re-exported here: the choice of backend is the caller's,
+and importing the GPU one is what pulls in the optional ``gpu`` extra.
+"""
+
+from __future__ import annotations
```
robertodr, line 25: did you forget to import from the modules and add `__all__` here?

packages/monoprop-bench-tools/src/monoprop_bench_tools/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+"""Benchmark harness for monoprop.
+
+The reusable half of monoprop's benchmark suite: memory accounting
+(:mod:`monoprop_bench_tools.memory`), the builders for the benchmarked problems
+(:mod:`monoprop_bench_tools.models`), and the two renderers of a run's artifacts
+(:mod:`monoprop_bench_tools.report`, :mod:`monoprop_bench_tools.bmf`).
+
+The benchmarks themselves are not here; they live in monoprop's repository under
+``benches/``.
+
```
robertodr: removing some overly verbose comments
```suggestion
```

packages/monoprop-bench-tools/src/monoprop_bench_tools/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+Submodules are not imported eagerly, so ``import monoprop_bench_tools`` stays
+cheap and does not require the optional GPU dependencies.
```
robertodr, line 23: OK, I guess this answers my question above 😅

packages/monoprop-bench-tools/tests/conftest.py
```diff
@@ -0,0 +1,36 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Pytest configuration for the monoprop-bench-tools test suite.
+
+Kept independent of monoprop's own ``tests/conftest.py`` so this package can be
+tested on its own, without the repository around it.
```
robertodr:
```suggestion
```

packages/monoprop-bench-tools/pyproject.toml
```diff
@@ -0,0 +1,91 @@
...
+
+
+[project.urls]
+Homepage = "https://github.com/Algorithmiq/monoprop"
+"Bug Tracker" = "https://github.com/Algorithmiq/monoprop/issues"
+Changelog = "https://github.com/Algorithmiq/monoprop/releases"
+
+
+[project.scripts]
+monoprop-bench-report = "monoprop_bench_tools.report:main"
```
robertodr, line 50: very nice with the entry points 👍

packages/monoprop-bench-tools/pyproject.toml
```diff
@@ -0,0 +1,91 @@
...
+
+# The GPU high-water mark reads CUDA memory-pool counters, so it needs a CUDA
+# build of CuPy. That is deliberately not an extra: the wheels are large and
+# pinned to one CUDA version and one architecture, and declaring them here would
+# pull them into the lockfile of every workspace that depends on this package.
+# Install the CuPy build that matches your toolkit yourself; without it
+# `monoprop_bench_tools.memory.gpu` reports `Method.UNAVAILABLE` rather than
+# failing to import.
+
+
```
robertodr: this information is available in the README.md
```suggestion
```

robertodr: uhm, can you test if - apart from bloating the lockfile - there are downsides into having a `gpu-cupy12x` and `gpu-cupy13x` extras? (maybe versioned like we do in tinite?)

packages/monoprop-bench-tools/README.md
```diff
@@ -0,0 +1,62 @@
...
+| `monoprop_bench_tools.report` | Renders a run's artifacts into a side-by-side Markdown report. |
+| `monoprop_bench_tools.bmf` | Renders a run's artifacts into [Bencher](https://bencher.dev/) Metric Format JSON. |
+
+## Install
+
+```bash
+pip install monoprop-bench-tools
+```
+
+`monoprop_bench_tools.memory.gpu` additionally needs a CuPy build matching your
```
robertodr: if the answer to the above question (can we add it as an extra?) shows that the extra bloats things for the whole workspace, then you should add full instructions on what is needed to access the `gpu` bits. I think cupy and cuquantum-python

pyproject.toml
```diff
@@ -90,6 +96,22 @@ cache-keys = [
...
 
 
+# One lockfile for monoprop and the tooling it releases alongside itself, so a
+# `uv sync` cannot leave the two at incompatible versions.
+#
+# `packages/bench-third-party` is deliberately excluded. It pins a narrower
+# `requires-python`, a git dependency, and CUDA-specific `cupy` / `cuquantum`
+# wheels that exist only for linux-x86_64; folding those into this lock would
+# constrain (and frequently break) resolution for everyone who only wants to
+# build monoprop. It stays a standalone uv project with its own lockfile.
```
```suggestion
```

sonar-project.properties
```diff
@@ -3,7 +3,10 @@ sonar.organization=algorithmiq
 
 sonar.projectName=monoprop
 sonar.sources=cpp/monoprop,cpp/include,src/monoprop
-sonar.exclusions=benches/*,src/monoprop/_version.py,src/monoprop/_dispatch.py
+# `packages/**` holds the benchmark tooling. It was outside Sonar's scope while it
+# lived in `benches/`, and this refactor deliberately does not change that: bringing
+# it in is a separate decision, with its own backlog of first-time findings.
```
robertodr:
```suggestion
```

.github/workflows/deploy.yml
```diff
@@ -117,6 +125,30 @@ jobs:
           name: cibw-sdist
           path: dist/*.tar.gz
 
+  build-bench-tools:
+    # `monoprop-bench-tools` is pure Python and versioned off the same git tags, so it
+    # rides this workflow rather than one of its own. It needs no cibuildwheel matrix:
+    # one sdist and one universal wheel cover every platform.
```
robertodr:
```suggestion
```

.github/workflows/deploy.yml
```diff
@@ -186,11 +218,41 @@ jobs:
       - name: Publish to PyPI
         uses: pypa/gh-action-pypi-publish@v1.14.2
 
+  upload_pypi_bench_tools:
```
robertodr, line 218: I have configure a pending publisher on PyPI for `monoprop-bench-tools` once we make a new release, the package will be created on first upload

.github/workflows/deploy.yml
```diff
@@ -186,11 +218,41 @@ jobs:
...
+    # Separate from `upload_pypi` because it is a separate PyPI project: it needs its own
+    # trusted-publisher entry (this workflow, the `pypi` environment), and a failure here
+    # must not hold back the library's own release.
```
robertodr:
```suggestion
    # Separate from `upload_pypi` because it is a separate PyPI project, with its own
    # trusted-publisher entry
```

robertodr (commented):
Overall LGTM! Thank you @Panadestein I think we'll gain very valuable experience with uv workspaces. I have some comments on the extras for the other workspace packages.

packages/monoprop-bench-tools/src/monoprop_bench_tools/memory/__init__.py
Panadestein: It is deliberate as importing `monoprop_bench_tools.memory.cpu` executes the parent `__init__` first, so re-exporting here would run `memory/gpu.py` on every CPU-only import.

Panadestein: That docstring still says importing gpu "pulls in the optional gpu extra", which is stale since we dropped the extra. I will fix that.

packages/monoprop-bench-tools/pyproject.toml
Panadestein: I tested this. Running `uv sync --all-extras` at the root does not activate a member's extras, so our CI, devcontainer and VS Code tasks are unaffected currently. Running `uv sync --package monoprop-bench-tools --all-extras` installs `cupy-cuda12x` and `cupy-cuda13x` into one environment. They ship the same top-level `cupy` package, so whichever unpacks last wins and you silently get a CuPy built against the wrong CUDA runtime. I guess the best is to keep it undeclared and document it, so we avoid the conflict?

packages/monoprop-bench-tools/README.md
Panadestein: Will do. At the moment `memory.gpu` only ever does `import cupy` so we need only that. `cuquantum-python` is a dependency of the third-party comparison scripts, and it's already pinned where it belongs in there.

.github/workflows/deploy.yml
Panadestein: Thanks! I'll drop the ⚠️ callout from the PR description and trim the comment in `deploy.yml`.

packages/monoprop-bench-tools/pyproject.toml
robertodr: `uv` has a notion of conflicting dependencies https://docs.astral.sh/uv/concepts/resolution/#conflicting-dependencies maybe it helps here? Sorry for insisting, but keeping the installation documentation as short as possible usually helps avoiding "it doesn't work for me" situations

Panadestein: Took the extras, you're right that the one-liner is worth it, and the README section is now a third of the length. I did try conflicts thing, but uv validates the conflict set eagerly, so declaring the pair makes plain `uv sync --all-extras` fail for the whole workspace, even though root `--all-extras` never activates a member's extras anyway.

Panadestein: So in short: left undeclared, root `--all-extras` resolves clean and pulls no CuPy, and `--extra gpu-cuda12` installs exactly one build. What do you think of this way?

packages/monoprop-bench-tools/pyproject.toml
```diff
@@ -0,0 +1,90 @@
...
+  "numpy>=2",
+  "psutil>=7.2",
+  "tabulate>=0.9",
+]
+
+
+[project.optional-dependencies]
+# Deliberately not declared as a `[tool.uv] conflicts` pair: that would make
+# `uv sync --all-extras` fail for the whole workspace, and asking for both is
+# already nonsense (they ship the same top-level `cupy` package).
```
robertodr:
```suggestion
```

robertodr (approved):
Thank you!

packages/monoprop-bench-tools/pyproject.toml
robertodr: It's great :)

PR #230, opened 2026-08-15
fix(benches): 🐛 record placement for every benchmark, not just the random ones

diagonal-hamiltonian:

## Summary

The benchmark suite could not tell a fast run from an unpinned one, and several of its instruments were measuring the harness rather than the engine. This fixes three of them. No engine code changes.

The one that matters most is **placement**. `pinned_thread_summary()` was called from inside `OpMemory.stop()`, and the fixed-model benchmarks (`bench_models.py`) never open a memory window — so they recorded no placement at all. An unpinned multi-rank run is not a slow run, it is a **meaningless** one: co-located ranks collapsed onto the same cores produce timings that look like code changes. Placement is a property of the run, not of the memory window that happened to observe it, so it moves to module scope and is called from `record_model_stats` too. It still has to be sampled while the engine's partition threads are alive, which is why it stays inside an instrumentation hook rather than moving to configure time.

## Changes

- `_record_placement(comm)` at module scope in `benches/conftest.py`, called from both `OpMemory.stop()` and `record_model_stats`. min/max across ranks, because a *partial* failure — some ranks placed, some not — is the interesting case and the shape a batch-system cpuset confinement actually takes.
- `barriered()` no longer times its own entry barrier. It was charging each rank with the skew accumulated in the *preceding* setup rather than with the operation — worth ~1.44× of a 2.86× observed spread on `build_graph`. The residual is still unexplained, so this makes the wall *less* untrustworthy, not trustworthy.
- `PssSampler` is now opt-in behind `--bench-pss-sampler`. It reads `/proc/self/smaps_rollup`, whose cost is O(address space) and which takes `mmap_lock`; above a few GiB it perturbs the run it is measuring, and because it only samples when the GIL is released, the **faster** side of an A/B is sampled less and reports a *lower* peak. `HighWaterMark` is exact, needs no thread, and is what to quote.
- Adds a per-operation memory window (`op_memory`) spanning the timed call alone — the autouse `record_memory` window includes `setup`, so for a benchmark with a setup it reports the construction transient rather than the operation.
- Adds a `test_random_propagate` benchmark. **Adding** a benchmark is safe for Bencher's history key; nothing here renames or moves one.
- `AGENTS.md` records the traps: `pedantic`'s `setup=` builds round *k+1* before releasing round *k* (so `rounds > 1` roughly doubles peak memory), pytest's fd-level capture hides C++ diagnostics without `-s`, and asserting an instrument's emission count *at the point of collection* is what stops "the two arms measured the same" and "the instrument never fired" from being the same observation.

## Verification

`pytest benches/bench_random.py` and `benches/bench_models.py` both run and now record `pinning` on either path (checked at 2 threads and at 8 ranks × 16 partitions, `16..16 of 19` under `srun --cpu-bind=cores`). `prek run --from-ref origin/main --to-ref HEAD` clean.

## Checklist

## AI/LLM disclosure


diagonal-hamiltonian:

PR #231, opened 2026-08-15
fix(partition): 🐛 place partitions inside the cgroup the batch system carved

diagonal-hamiltonian:

## Summary

**Placement must not divide an already-divided machine.**

`enumerate_physical_cores` reports only the cores inside the calling thread's affinity mask. When a launcher has already given each co-located rank its own disjoint slice — `srun --cpu-bind=cores`, or any cgroup-confined launch — that slice **is** the rank's share. Passing the node-wide ranks-per-node through as `group_count` then asks for `group_count × n` cores out of a list that only ever held `n`, `placement_order` correctly refuses, and **every rank silently runs unpinned**. The two-level barrier loses its domains at the same time, because `cpuset_domains` derives them from the placement.

Measured on 8 ranks × 16 partitions: **437 µs/sync unplaced against 15.5 µs/sync placed.**

This is a correctness-of-measurement bug more than a performance one — it does not change any result, but it silently invalidated four separate A/B jobs before it was found, because an unpinned run looks exactly like a slow code path.

## Changes

- `PartitionGroup` allgathers the affinity masks over its node-local communicator, and `classify_node_mask` **measures** disjointness; a `NodeMask::PerRank` result collapses `group_count` to 1.
- Mask *width* deliberately cannot substitute for that measurement. "8 ranks holding 16 cores each" and "8 ranks sharing one 16-core mask" both leave a rank seeing 16 of 128, and they need **opposite** placement. Collapsing in the wrong direction pins every co-located rank to the *same* cores, so `Shared` is the default and the safe error.
- `cpu_topology_policy_per_rank_slice_starves_without_collapse` pins the mechanism without needing live hardware, which is what makes it a regression test rather than a machine-specific one.

## A note for reviewers

This regressed **once already**, when topology discovery was rewritten onto hwloc, because the guard lives in the placement policy rather than in discovery. Any rework of that layer has to re-check it — `AGENTS.md` now says so.

## Checklist

## AI/LLM disclosure

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -73,6 +73,21 @@ Key files:
   for the mutating/collecting paths, which run on the partitions' own pinned masters; `sum_partitions_`,
   `fold_partitions_`, `first_partition_` for reads off quiescent partitions) rather than hand-rolling a
   `run_on_all` loop — the declarations record which helper is legal where.
+- **Placement must not divide an already-divided machine.** `enumerate_physical_cores` reports only cores
+  inside the calling thread's affinity mask, so when a launcher has given each co-located rank its own
+  disjoint slice (`srun --cpu-bind=cores`), the slice *is* the rank's share. Passing the node-wide
+  ranks-per-node through as `group_count` then asks for `group_count × n` cores out of a list that only
+  held `n`, `placement_order` correctly refuses, and every rank silently runs unpinned — which also costs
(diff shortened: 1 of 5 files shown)


robertodr:
I'd say we leave it to the user to decide affinity and binding of threads. Doing it programmatically is becoming a can of worms...

diagonal-hamiltonian:

PR #232, opened 2026-08-15
feat(evolution): 🔍 attribute the layer build behind monoprop_LAYER_PROFILE

diagonal-hamiltonian:

## Summary

Adds `monoprop_LAYER_PROFILE`: a per-partition attribution of the layer build, off by default.

Before this there was no way to say where a layer build's time went. The instrumented phases summed to ~6% of wall, so a phase split described almost nothing, and every question about the engine had to be answered with the end-to-end wall — an instrument whose observed spread on `build_graph` is **2.86×**, and which needs **~1590 paired repetitions** to resolve a 3% effect. The phase timers resolve the same 3% in about **5**.

This is the instrument the rest of this work is measured with, and it is useful on its own for anyone profiling the engine.

## Changes

Two instruments with deliberately different costs:

- **Phase timers and per-gate counters** — O(1) clock reads per gate, with per-term counters kept in locals and folded into the slot once per gate, so the emit loop keeps its registers.
- **A store population sweep** (k/d histograms, paired and overflow fractions) — O(n), so it runs only on power-of-two gate indices, which also yields the population's growth curve for free.

`layer_ns` brackets the **whole** of `build_layer`, so `layer_ns` minus the phases is the unattributed remainder. Without it the split is not falsifiable: phases that sum to a small fraction of wall can be individually accurate and collectively meaningless. `index_ns` covers the inverted-index rebuild, which is not a layer phase but is charged per gate and had never been attributed to anything.

Slots are heap-allocated per thread and owned by the registry, **never freed**: a partition master can outlive or predecease the dump, and a dangling slot would be a use-after-free in a diagnostic. The registry dumps from its destructor rather than via `atexit` so the ordering is sound — `slot()` evaluates `config::get()` before `Registry::add()`, so `Settings` is constructed first and therefore destroyed last.

## Cost when off

`slot()` returns `nullptr` when the knob is unset, so every call site is one predictable branch and no clock is read. It changes no result.

Output goes to **stderr**, which pytest's fd-level capture discards without `-s`. Documented in `docs/content/docs/features/parallelism.mdx`.

## Verification

`ctest -L serial` — 208/208 green. `pytest tests --with-mpi` green at 1, 2 and 4 ranks.

## Checklist

## AI/LLM disclosure

### DIFF
--- a/cpp/monoprop/detail/EnvConfig.h
+++ b/cpp/monoprop/detail/EnvConfig.h
@@ -24,6 +24,7 @@
 //   monoprop_NUM_THREADS    positive int (1..1e6), else ignored                → num_threads
 //   monoprop_PARTITION_PINNING  bool, default ON; 0/false disables per-core pinning → partition_pinning
 //   monoprop_PARTITIONS         int N | "auto" | "off"; parsed where it is used (resolve_partition_count_)
+//   monoprop_LAYER_PROFILE      bool, default OFF; per-partition layer-build attribution to stderr → layer_profile
 
 namespace monoprop::config {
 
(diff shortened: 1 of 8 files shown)


diagonal-hamiltonian:

PR #233, opened 2026-08-15
perf(mpi): ♻️ lease alltoallv staging buffers instead of allocating per call

diagonal-hamiltonian:

> **Stacked.** Based on `pr/layer-profile` (#232) so the diff shown here is this change alone. It retargets to `main` when that one merges.

## Summary

`begin_alltoallv` resized a fresh send buffer every call and copied every per-rank vector into it, and `wait_into` allocated each per-source block. Both sit inside the exchange timer, once per gate per partition.

Fitting `exchange = A + B·bytes` across two problem sizes with the collective count held constant isolates a fixed component **A = 9.38 partition-seconds — 21.4% of a leg that is itself 68.2% of layer time**. Per-call heap traffic is what that component is made of.

## Changes

The buffers now come from a per-thread free list. Ownership is the partition master, which is the per-partition scope: each partition runs its whole gate sequence on one pinned master, so capacity carries gate to gate and every buffer stays on the NUMA node whose thread first grew it. The list is typed on the element type, so the query leg can never be handed a bundle belonging to the response leg.

**What makes reuse safe is the handle's lifetime, not a caller obligation.** `Comm.h` requires a buffer be unmodified until the verb's second barrier. `PendingAlltoallv` is now move-only and completes its MPI request **in the destructor**, before the bundle returns to the pool. Recycling a buffer under a live `Ialltoallv` is exactly the silent-wrong-answer failure this class of change invites, so it is closed structurally rather than by convention.

## Tests

Three reuse tests, all in the shape that actually fails — sizes going **large, small, large**, so a logical size left at the high-water mark or a counts tail never rewritten shows up. Verified by mutation: dropping the `recv_counts` clear fails 12 cases; appending instead of assigning in `wait_into` fails 1.

**Not covered:** recycling under a live `MPI_Ialltoallv`, since `ShmComm` never sets a request. That path needs the multi-rank equivalence run, which is green at 1/2/4 ranks here but is not a targeted test of it.

## Checklist

## AI/LLM disclosure

### DIFF
--- a/cpp/monoprop/detail/evolution/layer_build/Engine.h
+++ b/cpp/monoprop/detail/evolution/layer_build/Engine.h
@@ -388,22 +388,31 @@ struct LayerBuildEngine {
             sendp = &sink.send_buffer(queries_r, src_val_r, combined_qv_);
         }
         std::vector<VecZ> &send = *sendp;
-        std::vector<std::vector<size_t>> inc_q;
-        {
+        // The handle owns the unpacked per-source blocks (leased from the partition master's scratch
+        // pool), so it has to outlive its reader rather than being a temporary: the whole point is that
+        // those blocks keep their capacity from gate to gate. Holding it across resolve_incoming also
+        // keeps the leg's send buffer leased for as long as `send` is meaningful, which is what makes
+        // the pooling invisible to the second-barrier contract in Comm.h.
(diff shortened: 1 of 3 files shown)


diagonal-hamiltonian:

PR #234, opened 2026-08-15
perf(algebra): ⚡ decide the structural cutoff from a (k, d) digest

diagonal-hamiltonian:

> **Stacked.** Based on `pr/mpi-lease-staging` so the diff shown here is this change alone. It retargets to `main` as the stack merges.

## Summary

The structural cutoffs only ever needed two integers about a monomial: its popcount `k`, and the number of modes holding both of their Majorana bits, `d`. Computing them meant calling out to `cutoff_sums`, which walks the dense words to build a sum it then throws away.

`paired_mode_count` computes `d` inline from the dense words and `passes_with_popcount` decides from `(k, d)` directly, dropping the `popcount <= cutoff` early-out. The dense representation is **untouched**: this changes no layout and moves no data, which is why it is a pure decision change and why its counters must stay bit-identical between arms.

## Measured

Default **ON**. `emit_s` **0.9200×**, 11/12 paired reps over two nodes, sign-test p=0.0063; pooled with 128- and 512-mode shapes **0.9214×, 19/20, p=4.0e-5**. Counters (`emit`/`reject`/`push`) bit-identical between arms in all 20 pairs, memory flat.

Worth **~0.8% of `build_graph`** — *not* the ~3% the microbenchmark projected. In isolation the kernel measures far better than it delivers, because in the real scan the loop is bound by the latency of the random row read rather than by the arithmetic beside it. Quoted at what it delivers.

Measure this on `emit_s`, not on the wall: at the observed variance the end-to-end benchmark needs ~1590 paired reps to resolve 3%, and the phase timer needs ~5.

## The test that matters more than the perf

`env_config_member_defaults_match_parsed_defaults`.

Every knob declares its default **twice** — once as a `Settings` member initialiser, once as `parse_flag`'s fallback in `get()` — and `get()` unconditionally overwrites the first with the second. So changing only the member silently does nothing.

That is not hypothetical. It shipped a measured **7.2% regression enabled by default**, on a knob that had been "turned off" in the member, in the header comment and in the docs — but not in the fallback. Any knob whose two declarations disagree now fails a test.

## Verification

`ctest -L serial` green; `pytest tests --with-mpi` green at 1/2/4 ranks.

## Checklist

## AI/LLM disclosure

### DIFF
--- a/cpp/monoprop/algebra/AlgebraCommon.h
+++ b/cpp/monoprop/algebra/AlgebraCommon.h
@@ -24,6 +24,7 @@
 
 #include "monoprop/TypeAliases.h"
 #include "monoprop/Utilities.h"
+#include "monoprop/core/SparseMonomial.h"
 
 namespace monoprop {
 
@@ -95,6 +96,9 @@ auto is_paired(const VecZ &mono) -> bool {
     return is_paired<NumModes>(indices_to_bitset<NumModes>(mono));
 }
 
+// The (k, d) digest form of the predicate above is is_paired(size_t, size_t) in SparseMonomial.h,
+// pulled into this namespace by the include; it is not repeated here so the two cannot drift.
+
 template <size_t NumModes, typename Rows>
 auto is_fully_paired(const VecZ &inds, const Rows &op) -> VecZ {
     VecZ result;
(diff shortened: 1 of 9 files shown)


diagonal-hamiltonian:

PR #242, opened 2026-08-18
refactor(c++)!: :art: de-tangle picture decisions

robertodr:

## Summary

Two related refactors, no behaviour change. Both are verified **bit-identical** against `main` on every
floating-point result the pictures reach.

**1. State each picture rule once, in a stateless policy.** The Heisenberg/Schrödinger choice was one
`bool schrodinger_`, derived from `std::optional<unsigned int> schrodinger_cutoff`, and read at 17
decision points inside `MonomialPropagator`. The gate-order rule alone was written four times, and
`contract_partially` held two near-identical ~28-line arms. `HeisenbergPicture` and
`SchrodingerPicture` are now sibling policy models over shared primitives, exactly as
`MajoranaAlgebra`/`PauliAlgebra` already were, bound to the runtime value once per public entry point
by `with_picture`. The whole private layer is written against one picture and never re-tests which one
it is in.

The choice also becomes visible in the type system: the constructor takes a
`PictureSpec = std::variant<Heisenberg, Schrodinger>`, so only a Schrödinger run carries a state
cutoff. That deletes the "an empty optional means Heisenberg" encoding, along with the dead
`schrodinger_cutoff.value_or(cutoff + 2)` fallback the old spelling made unreachable.

**2. Take the picture out of `MPGraph`, then simplify what was left.** `MPGraph` carried a `Picture`
and branched on it, which put a simulation concept inside a container. It now takes an `ArrivalOrder`
naming the slope of the optimizer slots a build hands to `append()`. Following that through surfaced
five further problems in how layers were stored and addressed, each fixed in its own commit:

- **The partial-slice machinery was dead.** `contract_partially` was the only production caller of
  `slice_graph`/`slice_view`, and it always passed `key == layers()`. Both meant one view and one
  drain, so `front_offset_`, `maybe_compact_layers`, the 4096 compaction threshold and the erase arm
  had no production reachability at all. Gone, with the tests that were their only reachers.
- **A slice's metadata could contradict its own layout.** `slice_graph` stamped its source's growth
  onto a slice whose layer order disagreed with it, guarded only by a prose warning. It is now
  unrepresentable: `MPGraph` has one constructor.
- **`MPGraphView` reimplemented `std::span`.** Its `(container pointer, base, count)` trio is one
  `std::span` member; `sizeof` drops 40 → 24 bytes on a type copied by value into an `EvalRequest` on
  every evaluation, and the class stops naming `std::vector`. The header now also records why this is
  not a `std::ranges` adaptor: consumers index it rather than iterate it, one type must serve both
  directions and `std::ranges::any_view` is C++26, and the type sits in exported signatures.
- **`n - 1 - x` was open-coded at several sites**, each re-deriving the storage invariant from a comment.
  They all call `slot_of_layer()` now, which is its own inverse, with two `static_assert`s pinning it. It
  and the bounds-checked `checked_layer_offset()` live below `MPGraph`, so a graph and the views over it
  share one spelling of the mapping and one diagnostic instead of a copy each.
- **`pare_graph` took the picture to answer a question the graph already knew.** It read a
  `bool schrodinger` to choose which end to sweep from. The sweep in fact starts at the last gate the
  build applied under *either* picture — Heisenberg reaches it via layer `n-1` (slot 0, its last
  simulation step), Schrödinger via layer 0 (slot `n-1`, also its last) — because the paring seed is
  the result of the whole evolution, and reachability runs backwards from a result through the circuit
  that produced it. `pare_graph` now takes no direction at all and asks
  `MPGraph::layer_of_unbuild_step()`. Nothing previously guarded the direction beyond one hand-passed
  argument in a test; a new case asserts both arrival orders yield the same sweep sequence.
- **An ascending-slot build inserted at `layers_.begin()` once per gate**, so appending was O(n) each
  and the build was quadratic. `layers_` now only ever `push_back`s and `get_layer()` maps a layer
  index onto arrival order instead. Measured on one machine at 20k/40k/80k layers:
  **177 / 716 / 2863 ms before, 4.6 / 7.5 / 14.7 ms after** — quadratic to linear, ~195× at 80k.

**3. A cleanup pass over the result.** A review of the finished branch for reuse, duplication, wasted
work and misplaced logic turned up a further set, all in the last commit and all bit-identical:

- **Two policy constants were not independent facts.** `is_schrodinger` restated `picture`, and
  `contract_phase` was `apply_sign` under a second name — both are the sign an applied angle carries,
  and `contract_partially` replays the very angles the build applied, so they cannot differ.
  `contract_reverse` is likewise `arrival_order == DescendingSlot`, because `map_params` must invert
  `gate_slot`. The `PicturePolicy` concept shape-checks exactly these constants and none of the
  behavioural members, so a picture whose bits disagreed still compiled. They are derived now.
- **Three private members took the policy but used it for nothing.**
  `apply_initial_operator_`, `build_evolve_result_` and `propagate_one_` read `P::picture` straight
  back out as the runtime value their callee already takes, duplicating a whole body per mode width
  for no specialization. They take the runtime value.
- **`FusedApply.h` included `picture/Picture.h`.** `apply_fused_contract` only tests a dependent
  `P::is_schrodinger`, which needs no definition of `P`, so the include put the evaluation API and the
  graph behind a header in the layer-build hot path. Dropped.
- **`build_graph`'s seed vector was copied twice** — once into the call, once inside it — though it is
  dead at the call site. Moved.

## Breaking changes (C++ only)

The Python API is untouched: `schrodinger_cutoff` keeps its historical spelling and the binder
resolves it into a `PictureSpec` at the boundary, and the `schrodinger` property still reads the same.

| Declaration | Change |
|---|---|
| `MonomialPropagator` ctor | `std::optional<unsigned int> schrodinger_cutoff` → `const PictureSpec &picture` |
| `~MonomialPropagator`, `update_initial_operator` | no longer `virtual` (nothing overrode either) |
| `MPGraph` ctor | takes `ArrivalOrder`; the `(order, std::vector<Layer>)` overload is removed |
| `MPGraph::slice_graph`, `slice_view`, `growth` | removed — use `contraction_view()`, `clear()`, `replace_layer()` |
| `LayerGrowth` | renamed `ArrivalOrder`, enumerators `DescendingSlot`/`AscendingSlot` |
| `MPGraphView` ctor | `(std::span<const Layer>, bool)` instead of `(const std::vector<Layer> &, size_t, size_t, bool)` |
| `pare_graph` | drops `bool schrodinger`; takes no sweep direction at all |
| `detail::build_layer` | `bool schrodinger` → `Picture picture` |
| `detail::apply_fused_contract` | takes the picture policy as a template parameter, drops `bool schrodinger` |
| `detail::MPOperator::update_initial_operator` | `bool schrodinger` → `Picture picture` |

## Checklist

- [X] Tests added or updated to cover the changes
- [X] Documentation updated (docstrings, `docs/`, `CONTRIBUTING.md`) if needed

## AI/LLM disclosure

- [ ] I used the following tool to help write this PR description:
- [X] I used the following tool to generate or modify code: ClaudeCode:claude-opus-5

> [!IMPORTANT]

> [!WARNING]


cpp/include/monoprop/MPGraph.h
```diff
@@ -80,27 +86,32 @@ class monoprop_EXPORT MPGraph {
         storage->param_index = param_index;
         storage->gen_coeff = gen_coeff;
         storage->gate_index = gate_index;
-        append_layer(Layer(std::move(storage)));
+        layers_.push_back(Layer(std::move(storage)));
```
robertodr:
```suggestion
        layers_.emplace_back(std::move(storage));
```

cpp/monoprop/detail/graph/MPGraphViews.h
```diff
@@ -56,34 +57,38 @@ struct GraphMemoryBreakdown final {
...
+//
+// Not a std::ranges adaptor, and not iterable, for three reasons:
+//  - Index i addresses three parallel things at once -- the layer, params[i] in evolve_operator(), and
+//    the recipe cache in build_cos_callbacks(). ev_and_grad() passes a computed index, not an iteration
+//    step. The consumers need an index space, not a sequence.
+//  - One type must serve both directions. views::counted(...) and views::reverse(views::counted(...))
+//    are different types, and std::ranges::any_view is C++26.
+//  - The type appears in the exported signatures of ev(), ev_and_grad(), evolve_operator() and
+//    state_operator_derivative_local(). A ranges adaptor would put a library-version-dependent
+//    template soup into their mangled names.
```
robertodr, line 92:
```suggestion
```

cpp/include/monoprop/MPFunctions.h
```diff
@@ -107,10 +107,11 @@ monoprop_EXPORT auto ev_and_grad(const EvalRequest &request,
                                  const detail::CosCallbacks &cos = {}) -> std::pair<double, VecD>;
 
 /// Prune `graph` to the subgraph reaching `nonzero_inds`; `full_cos_of_layer(i)` supplies layer i's full cosine set.
+// Takes no sweep direction: the seed is the result of the whole evolution under either picture, so the
+// sweep is always the graph's own build reversed. MPGraph::layer_of_unbuild_step is that order.
```
```suggestion
```

cpp/monoprop/core/Picture.h
```diff
@@ -0,0 +1,42 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+// Deliberately dependency-free: MonomialPropagator.h carries a Picture in its signatures, and MPOperator.h
+// in a parameter. The behaviour behind each value lives in picture/Picture.h.
```
robertodr:
```suggestion
```

robertodr:
I'd still like to see this change happen, but it's not the most fruitful refactor at this point.

PR #244, opened 2026-08-19
test(bench): split the fixed-model benchmarks into four timed operations

diagonal-hamiltonian:

Branch `pr/bench-and-test-harness` @ `a2f543e`, merged up to `origin/main` @ `d26b23c`.
`main`'s #227 moved the shared bench modules into `packages/monoprop-bench-tools/`, so the
`resting_rss_bytes` and `MODELS` edits land there — visible as renames with detection on.

## Summary

The four fixed-model benchmarks were one blended figure per model, which cannot answer where the
memory goes: a single timed region that builds a graph *and* evolves an operator reports one peak
RSS for both. This splits each model into four separately timed operations — `build_graph`,
`propagate`, `energy`, `gradient` — each with its own time and its own memory window. `build_graph`
and `propagate` are mutually exclusive per process, so a `propagate` row's peak RSS *is* operator
memory (its graph ledger reads exactly `0`) and a `build_graph` row's *is* graph memory. `test_model`
itself is untouched: renaming a tracked series orphans its Bencher history.

It also adds **`monoprop_TEST_EXCLUDE_MPI_FABRIC`**, a CMake option, ON for MPI builds: the
single-process test variants stop initialising every fabric device in `MPI_Init`. **2.03 s → 0.61 s
per case**, faster on all 208 cases the two runs share. **This changes the CI default there**, and it
is safe only there — the same setting hangs multi-rank runs, so it is scoped to the `serial` variants
through `discover_tests`' new `SERIAL_ENVIRONMENT` argument. Heads-up for #166: it carries the same
option and `SERIAL_ENVIRONMENT` block, and this one is a strict subset of it (0 non-comment lines here
are absent there), so whichever lands second should drop its own copy rather than merge both.

### Measurement — this PR is the null control for the rest of the stack

It adds no library code. `pr3` and `main` differ only by an upstream `class`→`typename` sweep that is
provably codegen-identical, yet their modules hash differently, so the A/B harness measures them as
two genuine arms — measuring its own false-positive rate rather than assuming it.

    main _core.so  f73d17f264e62f52987591dea0c79b50   (origin/main 6abd839, pinned)
    port _core.so  216d7e2f0593aa46cbe12d133dd814bd   (pr/bench-and-test-harness cd58352)

**0 of 24 timing tests resolved**, best agreement anywhere 8 of 10 reps. Every operation flat on both
workloads, layouts and node counts; term counts identical in all 12 cells.

That campaign predates the entry-barrier change below, so the four model operations were measured with
no entry barrier on either side. Both arms shared that one instrument, so the false-positive-rate
statement stands. The barrier has since been measured directly, by timing its own `comm.Barrier()` per
rank over 10 fresh-process reps: it removes **1.09% of makespan CV on `propagate` and 7.46% on
`build_graph`** against materiality thresholds of 0.55% and 0.48%, leave-one-out stable 10 of 10 on
both. `energy` and `gradient` measure 0.07% and 0.01%, 3–20× below threshold — they pass
`barrier_setup` with no inner setup, and the graph fixture's closing allreduces already resync the
ranks. So the recorded ratios stand, but "unresolved at this rep count" on the first two operations was
read against an avoidably noisy instrument.

The grid: layout A = 1 rank/node × 128 partitions, B = 8 × 16, at N = 1 and 2; hubbard c10 /
`lower_atol=1.25e-5` and pauli c14 / `5e-5`, plus a `build_graph[hubbard]` rung at
`--hubbard-trotter-steps=2`. A and B hold the flat world P constant, so A-vs-B is transport alone.
`--cpu-bind=cores --distribution=block:block` on both arms, recorded per cell. 10 interleaved reps,
order flipped per (rep, cell); median of **paired** per-rep ratios, sign test, Holm across the 24-test
family. Ten reps because six floors the sign test at p=0.031, which Holm over 24 makes 0.75.

**Not reproducible from this diff.** Measured at `cd58352`, before the merge, on Deucalion by a private
harness driving two prebuilt venvs whose hashes are above. The batch scripts and report generators live
in a separate private repo and are not proposed for this one.

## Changes

- **Four timed operations per fixed model** (`benches/bench_models.py`): `test_model_build_graph`,
  `test_model_propagate`, `test_model_energy`, `test_model_gradient`, each with its own `op_memory`
  window. Run one group per pytest process — the four together do not fit a node at these sizes.
- **`monoprop_TEST_EXCLUDE_MPI_FABRIC`** (`cpp/tests/CMakeLists.txt`) plus `SERIAL_ENVIRONMENT` in
  `discover_tests` (`cpp/tests/boost-test.cmake`, `boostAddTests.cmake`), so the exclusion reaches the
  single-process variants only.
- **The benchmarks now assert what they measured**: each of the four asserts the operation produced
  terms, and the two `build_graph` cells that the graph is non-empty — no new collective, since the
  count is already reduced for the report, and `allreduce` makes the assertion rank-uniform rather than
  aborting ranks unevenly.
- **`opsize` no longer leaks into Bencher's `operator[...]` history** (`bmf.py`): it is keyed by pytest
  node id, which without a guard enters that permanent series beside the per-picture entries. Covered
  by this branch's first `bmf.py` test.
- Supporting harness fixes: the entry barrier moves out of the timed region into `pedantic`'s `setup`,
  so it no longer charges each measurement for skew in the preceding setup; an A/B arm is identified by
  its `_core.so` hash, not a dist-info version that goes stale after a rebuild; per-rank peak RSS and
  thread affinity come from `/proc` and `/usr/bin/time -v`; and model benchmarks whose graph cannot fit
  skip instead of OOM-killing the run.
- Notes for the next reader: `AGENTS.md` and `docs/content/docs/testing.mdx`.

## Checklist

- [x] Tests added or updated to cover the changes — `packages/monoprop-bench-tools/tests/test_bmf.py`
      covers the `opsize` guard; the four benchmarks assert their own output
- [x] Documentation updated (docstrings, `docs/`, `CONTRIBUTING.md`) if needed —
      `docs/content/docs/testing.mdx` for the fabric option, `AGENTS.md` for the bench invariants
- [ ] `CHANGELOG` / release notes updated if applicable — not applicable, no library change

## AI/LLM disclosure

- [x] I used the following tool to help write this PR description: ClaudeCode / claude-opus-5
- [x] I used the following tool to generate or modify code: ClaudeCode / claude-opus-5


benches/bench_models.py
```diff
@@ -12,21 +12,46 @@
...
 
 import pytest
 from monoprop_bench_tools.memory.cpu import resting_rss_bytes
-from monoprop_bench_tools.models import MODELS, barriered
+from monoprop_bench_tools.models import MODELS, barrier_setup, barriered
+
+# `build_graph` extends the graph, so a driver that re-applies its circuit retains one
+# layer-set per step: 3 steps of hubbard exceed a 242 GiB node where 2 fit in 15.4 GiB.
+# Skipping beats OOM-killing the machine of whoever runs `just bench`, which does not
+# pass -m "not slow". Pauli's step count is 1 and is unaffected.
```
Panadestein:
```suggestion
```

Panadestein: Some verbose comments.

benches/bench_models.py
```diff
@@ -12,21 +12,46 @@
...
 
+import os
 from typing import Any
 
 import pytest
 from monoprop_bench_tools.memory.cpu import resting_rss_bytes
-from monoprop_bench_tools.models import MODELS, barriered
+from monoprop_bench_tools.models import MODELS, barrier_setup, barriered
+
+# `build_graph` extends the graph, so a driver that re-applies its circuit retains one
```
Panadestein:
```suggestion
# `build_graph` extends the graph, so a driver that re-applies its circuit retains one layer-set per step
```

benches/conftest.py
```diff
@@ -88,6 +91,33 @@ def _reduce_sum(comm: Any, value: int) -> int:
...
+    """Gather per-rank CPU-id lists to rank 0. Collective; off root returns ``[]``."""
+    if comm is None or comm.Get_size() == 1:
+        return [values]
+    gathered = comm.gather(values, root=0)
+    return gathered if gathered is not None else []
+
+
+def _spread(comm: Any, value: int) -> dict[str, int]:
+    """Reduce a per-rank number to ``sum`` (bounds the job) and ``max`` (bounds a node)."""
+    return {"sum": _reduce_sum(comm, value), "max": _reduce_max(comm, value)}
```
Panadestein, line 116: Nonblocking: Do you think this bunch of MPI wrappers could be reused somewhere else in the codebase?

cpp/tests/CMakeLists.txt
```diff
@@ -51,10 +51,40 @@ target_link_libraries(
...
+# message. Measured on a compute node: 2.03 s per case without the exclusion, 0.61 s with it.
+#
+# A per-case launch is ONE process (world size 1, `*_World` cases skip themselves, every
+# other communicator is MPI_COMM_SELF), so the fabric can only cost startup time. The MPI
+# variants exchange real messages and MUST keep the full component set -- `OMPI_MCA_pml=^ucx`
+# hangs a 2-rank run indefinitely -- hence SERIAL_ENVIRONMENT rather than a global setting.
+#
+# Exclusions (`^`), never a positive component list: `vader` became `sm` in Open MPI 5, so
+# naming a component breaks on the next machine. Unknown OMPI_MCA_* variables are ignored by
+# other MPIs, so this is inert under MPICH.
```
Panadestein: I would condense this to two lines, really:
```suggestion
# CTest launches each Boost case as a world-size-1 process, so excluding fabric components cuts MPI_Init from 2.03 s to 0.61 s without affecting communication; real MPI tests must keep the full component set.

# Use SERIAL_ENVIRONMENT with exclusions (`^`), not a positive component list: Open MPI component names change across versions, and unknown OMPI_MCA_* variables are harmless under other MPIs.

```

Panadestein (dismissed):
Thanks! Take a look at the comments.

benches/conftest.py
diagonal-hamiltonian: Nah i dont think so to be honest.. most of the mpi is c++

benches/bench_models.py
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Done — those three paragraphs went in 2638b2b, before this push. The module docstring is now one line ("Run one group per pytest process: build_graph/propagate and energy/gradient do not fit a node together."), and every test docstring in the file is one line too.

benches/bench_models.py
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Taken verbatim in a2f543e. The line that had replaced it ("Not a free knob: the benchmark grid is calibrated on this value") is gone; the 242 GiB figure and the retain-per-step count stay in `skip_if_graph_will_not_fit`'s `pytest.skip` message right below, which is where a reader who trips the skip meets them.

cpp/tests/CMakeLists.txt
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Your two sentences verbatim in a2f543e, wrapped to the file's comment width (three + two lines rather than two long ones). That restores the 2.03 s → 0.61 s figures and the "harmless under other MPIs" note, and drops the `OMPI_MCA_pml=^ucx`-hangs-a-multi-rank-run detail — that one survives in `docs/content/docs/testing.mdx` and the PR body.

benches/conftest.py
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Checked rather than assumed: `_reduce_sum`/`_reduce_max`/`_reduce_min`/`_gather_lists` are the only `comm.allreduce`/`comm.gather` callers anywhere outside `tests/` — `src/monoprop/` has none, and `monoprop_bench_tools` has no counterpart to fold them into. So there is no second caller to share them with, and they stay private to `benches/conftest.py`.

diagonal-hamiltonian:
@Panadestein Good to go?

robertodr (approved):

PR #225, opened 2026-08-12
fix(propagator): :bug: invalidate functionals after `update_initial_operator`

JoshCudbyAlgo:
## Summary

expectation_value_functional and expectation_value_and_gradient_functional snapshot the initial operator's coefficients. update_initial_operator re-weighted the operator in place without invalidating functionals built beforehand, so a stale functional would silently answer for coefficients the propagator no longer holds. It now bumps a re-weight epoch counter that every functional call checks, raising the same StaleFunctionalGraph error used for a rebuilt graph.

## Changes

- Added initial_operator_epoch_ to MonomialPropagator (cpp/include/monoprop/MonomialPropagator.h), bumped on every update_initial_operator re-weight (cpp/monoprop/detail/monomial_propagator/MonomialPropagatorImpl.h)
- Added validate_expected_initial_operator, raising StaleFunctionalGraph when the epoch a functional captured no longer matches the propagator's current one (cpp/monoprop/Validation.{h,cpp})
- Wired the epoch check into make_functional_, so every call through expectation_value_functional/expectation_value_and_gradient_functional validates against it alongside the existing graph-layers check
- Updated docstrings on MonomialPropagator.expectation_value_functional, expectation_value_and_gradient_functional, and update_initial_operator, plus the MajoranaPropagator/PauliPropagator update_initial_operator overrides, to document that functionals are invalidated by a re-weight
- Updated docs/content/docs/features/evaluation.mdx to describe functional invalidation on graph mutation or initial-operator update
- Added update_initial_operator_invalidates_gradient_functional (C++) and test_expectation_value_functional_invalidated_after_initial_operator_update / test_expectation_value_and_gradient_functional_invalidated_after_initial_operator_update (Python) regression tests

## Checklist

- [X] Tests added or updated to cover the changes
- [X] Documentation updated (docstrings, `docs/`, `CONTRIBUTING.md`) if needed
- [ ] `CHANGELOG` / release notes updated if applicable

## AI/LLM disclosure

- [ ] I did not use LLM tooling, or used it only privately for ideation
- [X] I used the following tool to help write this PR description: ClaudeCode:claude-sonnet-5
- [X] I used the following tool to generate or modify code: ClaudeCode:claude-sonnet-5

> [!IMPORTANT]
[...]

PR #251, opened 2026-08-20
feat(propagator)!: :sparkles: explicit types for functionals and tighter invalidation control

robertodr:

## Summary

`make_functional_` returned an anonymous `std::function` whose twelve-item capture list carried its ownership and lifetime rules in comment prose, because the object had no name and no declaration. This gives it two named classes over one named plan, so those rules become a member list a reader can check. One `detail::FunctionalControl` block — structure revision, alive flag, name of the last structural change, published weights — replaces the two ad-hoc counters, and every mutator bumps it through `detail::MutationGuard` so no exit path can forget. `MonomialPropagator::num_mutating_methods` states the roster and `cpp/tests/functional_validity.cpp` `static_assert`s its row count against it, so a new mutator breaks the build until its effect on a live functional is recorded.

Reading the validity checks closely enough to move them turned up three defects: the graph-layer check could never fire for a pared functional (equal layer counts by construction, so a later `build_graph` passed both checks and the call read the cosine callbacks' raw pointers into a rebuilt inverted index — UB); `set_parameter_mapping` moved neither check, so a functional built before it returned a wrong number silently; and the epoch check dereferenced the propagator to decide whether dereferencing the propagator was safe.

A review pass over the replacement found five more, fixed in `fix(propagator): a call that commits nothing invalidates nothing`. They share one rule: **a call that commits nothing invalidates nothing, and a partition facade must reach that verdict where a single partition does.**

1. **An outright rejection invalidated a live functional on a facade but not on one partition.** `build_graph()` armed the facade's guard before the checks that reject a call up front, so a bad `gate_indices`, a wrong-length `parameter_mapping` or a too-short seed went stale at `monoprop_PARTITIONS=auto` and stayed valid at `=off`. Every such rejection is now decided above the fan-out.
2. **A re-weight naming an absent term retired a partitioned propagator.** Only the partition owning the term can see it is missing, so it was refused there with its siblings already committed — divergence, latched fault, propagator dead over a typo, and whether it happened depended on the host's core count through `auto`. Each partition now judges its own share before any of them writes.
3. **A fanned-out plan reached the partition group directly**, skipping the fault check every other door onto the partitions performs.
4. **`set_parameter_mapping` bumped by hand rather than through the scope**, so a part-way relabel recorded nothing. The unscoped `bump_structure_` is gone, which keeps the rule in one place.
5. **The identity-row rule was undocumented.** See below.

## Changes

- **`cpp/include/monoprop/Functional.h`** — `ExpectationValueFunctional<NumModes>` and `ExpectationValueAndGradientFunctional<NumModes>`, each a handle on one shared `detail::FunctionalPlan<NumModes>` whose `std::variant` carries both the single-partition and the facade shape. Bound with `nb::keep_alive<0, 1>`; `follows_weights` on both, so a caller can read the contract off the object.
- **`cpp/monoprop/detail/functional/Control.h`** — the shared validity block, the published `OperatorWeights`, and the `live_plans` count a re-weight consults before publishing.
- **`Validation.h/.cpp`** — `validate_expected_graph_layers` / `validate_expected_initial_operator` give way to `validate_functional_state` / `validate_weight_refresh`. Still `StaleFunctionalGraph`, so the promised `RuntimeError` is unchanged.
- **`MPOperator::validate_initial_operator`** — rejects exactly what `update_initial_operator` would while writing nothing, which is what the facade's dry pass runs per partition.
- **The plan owns its layers**, because `cos` holds a raw `CosMask` pointer per layer that a later `append_layer`, `slice_graph` or `maybe_compact_layers` must not move or free. Plus an operator-layout backstop — store pointer and inverted-index row count, re-derived before any use of that index — so a missing bump reports staleness instead of folding a rebuilt index through a stale pointer.
- **`cpp/tests/functional_validity.cpp`** — the mutation table: one row per public mutating method, for value and gradient, pared and unpared, both pictures; plus the rejection rows, where an outright rejection and a refused re-weight must leave the functional answering on both partition counts. `tests/test_parameter_validation.py::TestFunctionalValidityTable` mirrors it over `monoprop_PARTITIONS=off` and `=auto`.
- **Docs** — the mutation table on the evaluation page, the re-weight rule and its price in both factory docstrings and in `update_initial_operator`, the identity-row rule and the pre-check on the initialisation page, and what a failed mutation leaves behind on the parallelism page.

## Behaviour changes

Two, both breaking:

1. **`update_initial_operator` no longer invalidates a live functional.** The coefficient vector and core term travel together as a `detail::OperatorWeights` set published into the control block, read with one atomic load so `op` and `core_term` cannot come from two re-weights. The price: a functional is a live view of the weights, so two calls with the same parameters straddling a re-weight give two answers, and a caller needing a frozen value must rebuild. One case still throws — a Schrödinger functional with a `pare_threshold`, whose keep-set was thresholded from the coefficients the re-weight replaced. A *refused* re-weight leaves the functional valid, because it commits nothing.
2. **`update_initial_operator` replaces rather than patches.** Every existing term the dict omits is zeroed, **the identity term included**, so a re-weight carrying only the non-identity terms drops the constant offset the propagator started with. This is what the store already did with every other omitted row, but it is a silent numerical change for a caller re-weighting a Hamiltonian with a constant term: pass the identity row again to keep it.

## Verification

Green: 293 C++ tests, 826 Python tests, the runnable doc snippets, and `prek run --all-files` over every file this PR touches.

Before the review fixes: the layer-ownership change was proved bit-identical across eight configurations via hex-float fingerprints; the weight refresh is asserted bit for bit rather than to a tolerance, value and gradient, pared and unpared, at `partitions=1` and the default, under MPI at 2 and 4 ranks; an ASan+UBSan run over 232 cases was clean, and it is what caught the third defect's fix being incomplete in an intermediate commit, where the ordinary build read plausible freed memory and passed all 213 tests. The module shrank ~16 KiB and a call got 2–3% faster — losing a layer of type erasure pays for two extra classes per mode width.

**The review fixes have not been re-measured and have not been through the sanitizer legs.** They are covered by the suites plus three negative checks, each confirming its regression test fails when the fix is reverted: the `build_graph` hoist, the `set_parameter_mapping` hoist, and the weights-slot clear.

## Notes for the reviewer

- **The divergence latch is now unreachable on demand.** With the dry pass in place no public call diverges the partitions deterministically, so the latch is asserted where it is decided — the guard that latches and the validator it feeds. Its real-world trigger remains a gate generator too large for the system, bounds-checked only as the gate loop reaches it, so which partitions committed depends on which one poisons the collective first.
- **A `parameter_mapping` of the wrong length is only reachable from C++.** The Python front end refuses it first with `ValueError`, so that rejection row lives in `functional_validity.cpp` alone.
- **No `AGENTS.md` rule for the weights.** `num_mutating_methods` gates the row count, but nothing tells a future contributor that a new mutator must also decide its effect on the published weights. Worth adding; not in this PR.
- **No changelog entry**: this repository has no changelog file, and `.github/release.yml` generates release notes from PR titles. The `feat!` title and the behaviour changes above carry them instead.

## Checklist

- [ ] `CHANGELOG` / release notes updated if applicable — no changelog file exists; see the note above

## AI/LLM disclosure

- [x] I used the following tool to generate or modify code: Claude Code (claude-opus-5), noted inline by the `Assisted-by:` trailer on every commit

> [!IMPORTANT]

## Summary by CodeRabbit

* **New Features**
  * Added reusable expectation-value and expectation-value-with-gradient functionals with parameter metadata.
  * Functionals now follow updated operator coefficients when supported, while structural changes invalidate them.
  * Added partition-aware validation and failure handling for propagator mutations.
  * Python functional objects now retain their originating propagator and expose read-only metadata.

* **Bug Fixes**
  * Initial-operator updates are validated atomically, including omitted and identity terms.

* **Documentation**
  * Expanded guidance on functional validity, coefficient updates, and partitioned mutation failures.


src/monoprop/monomial_propagator.py
```diff
@@ -55,6 +55,37 @@
...
+
+    def __getattr__(self, name: str) -> object:
+        # Private names are answered by the instance dictionary alone: reaching the engine object for
+        # one would recurse here through `self._functional` before __init__ has set it.
+        if name.startswith("_"):
+            raise AttributeError(name)
+        return getattr(self._functional, name)
+
+
+class _BoundGradientFunctional(_BoundFunctional):
```
adamglos92: nitpick: Liskov principle is violated. Should there be common base generic class with the output?

tests/test_parameter_validation.py
```diff
@@ -186,18 +197,413 @@ def test_functional_invalidated_after_initial_operator_update(
...
 
-        # The functional snapshotted the old coefficients, so it must reject the call rather than
-        # keep answering for the operator the propagator no longer holds.
-        with pytest.raises(RuntimeError, match=r"MP object has been modified"):
+        # A re-weight moves no structure, so the functional built before it follows the new
+        # coefficients instead of refusing the call: it now answers what the propagator answers.
+        after = _value(functional(parameters))
+        assert after != pytest.approx(before)
+        assert after == pytest.approx(mp.expval(parameters))
+        assert after == _value(getattr(mp, functional_name)()(parameters))
```
adamglos92: question: would the test be cleaner if the reference value would be computed with a fresh mbs?

tests/test_parameter_validation.py
```diff
@@ -163,7 +174,7 @@ def test_functional_invalidated_after_graph_mutation(self, serial_comm):
             "expectation_value_and_gradient_functional",
         ],
     )
-    def test_functional_invalidated_after_initial_operator_update(
+    def test_functional_follows_initial_operator_update(
```
adamglos92, line 230: perhaps this question should also compare gradients? for some reason it only checks expectation values

tests/test_parameter_validation.py
```diff
@@ -186,18 +197,413 @@ def test_functional_invalidated_after_initial_operator_update(
...
-        with pytest.raises(RuntimeError, match=r"MP object has been modified"):
+        # A re-weight moves no structure, so the functional built before it follows the new
+        # coefficients instead of refusing the call: it now answers what the propagator answers.
+        after = _value(functional(parameters))
+        assert after != pytest.approx(before)
+        assert after == pytest.approx(mp.expval(parameters))
+        assert after == _value(getattr(mp, functional_name)()(parameters))
+
+
+class TestFunctionalValidityTable:
```
adamglos92, line 268: I'm completely lost with this tests

cpp/monoprop/Validation.h
```diff
@@ -38,11 +39,27 @@ monoprop_EXPORT auto validate_parameters_length(const VecD &params, const VecZ &
 
 monoprop_EXPORT auto validate_functional_call(const VecD &parameters, size_t expected_num_params) -> void;
 
-// The graph must still have the layer count the functional was built against.
-monoprop_EXPORT auto validate_expected_graph_layers(size_t current_layers, size_t expected_layers) -> void;
+/// What a functional must be able to say about its propagator before it reads anything it borrowed.
+// Assembled from detail::FunctionalControl plus, for a single-partition plan, two facts derived from the
+// operator itself. The derived pair is the backstop: it holds even for a mutation that forgot to bump.
```
robertodr:
```suggestion
```

cpp/monoprop/Validation.cpp
```diff
@@ -99,20 +99,49 @@ auto validate_functional_call(const VecD &parameters, size_t expected_num_params
...
 
-auto validate_expected_graph_layers(size_t current_layers, size_t expected_layers) -> void {
-    if (current_layers != expected_layers) {
-        throw StaleFunctionalGraph(std::format("MP object has been modified since the functional was created. "
-                                               "Previous number of graph layers was {} and now is {}.",
-                                               expected_layers,
-                                               current_layers));
+auto validate_functional_state(const FunctionalState &state) -> void {
+    // Aliveness first: every other handle a functional holds points into the propagator, so there is
+    // nothing else it may legally read once this is false.
```
robertodr:
```suggestion
```

cpp/monoprop/Validation.cpp
```diff
@@ -99,20 +99,49 @@ auto validate_functional_call(const VecD &parameters, size_t expected_num_params
...
+                                   "again from a live one.");
+    }
+    if (state.current_revision != state.expected_revision) {
+        throw StaleFunctionalGraph(std::format(
+            "MP object has been modified since the functional was created: {} changed the "
+            "graph or operator the functional replays. Create a new functional.",
+            state.last_structural_change != nullptr ? state.last_structural_change : "a structural mutation"));
+    }
+    // No bump, but the operator moved anyway: some mutation is missing its bump_structure_ call. Report
+    // it as staleness rather than read a rebuilt inverted index through a pointer to the old one.
```
robertodr:
```suggestion
```

cpp/monoprop/Validation.cpp
```diff
@@ -99,20 +99,49 @@ auto validate_functional_call(const VecD &parameters, size_t expected_num_params
...
-    if (current_epoch != expected_epoch) {
-        throw StaleFunctionalGraph("MP object has been modified since the functional was created. "
-                                   "The initial operator was re-weighted, so the coefficients the functional "
-                                   "snapshotted are stale; create a new functional.");
+auto validate_weight_refresh(const WeightRefresh &refresh) -> void {
+    // The caller passes its own follows_weights(), the property it advertises, so what is enforced here
+    // and what is advertised cannot drift apart. It is false for exactly one shape: a pared Schrodinger
+    // plan holds a keep-set thresholded from the operator coefficients themselves, so new coefficients
+    // would need a different keep-set and replaying this one would answer for a paring nobody asked for.
+    // Heisenberg thresholds the state, which a re-weight does not touch, so it follows exactly.
```
robertodr:
```suggestion
```

cpp/monoprop/Validation.cpp
```diff
@@ -99,20 +99,49 @@ auto validate_functional_call(const VecD &parameters, size_t expected_num_params
...
+    if (!refresh.may_follow_weights) {
+        throw StaleFunctionalGraph("MP object has been modified since the functional was created: the "
+                                   "initial operator was re-weighted, and this functional pares its graph "
+                                   "against the operator coefficients (a Schrodinger picture functional "
+                                   "with a pare_threshold), so it cannot follow the new weights. Create a "
+                                   "new functional.");
+    }
+    // Unreachable while every structural mutation bumps and every publication carries the revision it
+    // was made at: a functional whose revision still matches the propagator's cannot see weights from
+    // another revision. Kept as the backstop for the day one of those two stops being true.
```
robertodr:
```suggestion
```

cpp/monoprop/detail/functional/Control.h
```diff
@@ -0,0 +1,70 @@
...
+
+// The validity block a propagator shares with every functional plan it makes.
+//
+// A plan borrows from its propagator, so before it reads any of those handles it needs two facts its
+// own snapshot cannot supply: whether the propagator is still there, and whether the structure the
+// snapshot describes is still the propagator's. Both live here, behind one shared_ptr, so a plan
+// answers them without dereferencing the propagator at all. The propagator holds the only mutating
+// handle; plans hold shared_ptr<const>.
+//
+// A copied propagator gets its own block: a copy carries no functionals, so it starts at revision 0.
```
robertodr:
```suggestion
// A plan borrows from its propagator, so before it reads any of those handles it needs two facts: whether the propagator is still there, and whether the structure the
// snapshot describes is still the propagator's.
//
// A copied propagator gets its own block: a copy carries no functionals, so it starts at revision 0.
```

tests/test_parameter_validation.py
robertodr: no, that would defeat the point of the changes in this PR. Which is to ensure that functionals taken from a single propagator stay valid, under certain conditions. Basically, hardening what was done in #225

tests/test_parameter_validation.py
robertodr: this PR adds ways to keep a functional valid when a limited set changes are made to the parent propagator. This test checks that this limited set of changes produces the desired behavior.

tests/test_parameter_validation.py
robertodr: isn't it doing it already? this line looks suspicious:
```python
        assert after == pytest.approx(mp.expval(parameters))

robertodr: fixed.

robertodr:
@coderabbitai full review

PR #253, opened 2026-08-20
feat(graph): 🔍 separate the graph's flat-world slot cost from its traffic

diagonal-hamiltonian:

## Summary

`graph_memory_breakdown()` reported the graph's cross-rank storage as one number,
`cross_rank_bytes`, which mixes two costs that scale with different things:

- **one record per slot of the flat world** (`ranks × partitions`), whether or not that slot carries
  any traffic, and
- **the endpoints actually crossing**.

On a partitioned run the first grows with a `P` the rank count never shows, so a breakdown that adds
them cannot say which of the two a graph is spending on. This splits them.

## What changed

Five fields on the existing `GraphMemoryBreakdown`, deliberately **outside `total_bytes()`**,
because each is a count or a slice of a field already summed there:

| field | what it is |
|---|---|
| `slot_record_bytes` | the slot-proportional slice of `cross_rank_bytes` |
| `layer_cores` | distinct `LayerCore`s; `slot_records / layer_cores` recovers `P` |
| `slot_records` | summed over cores, so `P` per core |
| `occupied_slots` | slots carrying any traffic; `/ slot_records` is the occupancy |
| `cross_rank_endpoints` | the traffic itself, and a `P`-independent ceiling on `occupied_slots` |

Three helpers in `MPGraphEncoding.cpp` derive them. `cross_rank_storage_bytes` now **calls**
`cross_rank_slot_record_bytes` rather than repeating its expression, so the slice cannot drift from
the total it is a slice of.

Surfaced to Python as five `d_`-prefixed keys on `graph_memory_breakdown()`, matching the `d_`
convention `operator_memory_breakdown()` already uses for diagnostics excluded from the sum.

## Tests

Both sides of the binding, because they fail differently.

- **`cpp/tests/graph_encoding_tests.cpp`** — five cases. `occupied_slots` against zeros at the
  front, the interior and the back (the three places a scan loses count); `slot_record_bytes`
  against a pair of storages with **identical traffic and different world sizes**, which is the
  whole point of the field; `total_bytes()` unchanged by any diagnostic; and `+=` summing all five,
  since a partitioned propagator merges per-partition breakdowns.
- **`tests/test_memory_breakdown.py`** — the dicts are built inside nanobind lambdas with no C++
  entry point returning the map, so *which struct field each string key names* is observable only
  from Python. It also pins `monoprop_PARTITIONS`; left alone, `P` resolves to the host's physical
  core count and every slot-record expectation becomes machine-dependent.

`d_slot_record_bytes / d_slot_records` is asserted to be exactly 32 or 40 bytes —
`CrossRankPartnerRange` is two `size_t` and three `TermIndex`, the latter 4 B or 8 B under
`monoprop_WIDE_TERM_INDEX`, and nothing in between.

## Scope

**This PR was much larger and is now much smaller.** It previously also carried the
`monoprop_PROFILE` instrument. That instrument is held back: whether to ship an internal profiler at
all is a separate decision from whether the graph can report its own slot cost, and the two were
only ever in one PR because they were written in one sitting. The instrument is kept as a
cherry-pickable commit for measurement work and is not part of this diff. Consequences:

- **Nothing here is behind a build flag or an environment variable.** Five always-on counters, and
  `grep -r 'monoprop_PROF\|profile::' ` over this diff returns nothing.
- **@robertodr's `fprintf` comment no longer applies to this PR** — that code is not here. It was
  **six** call sites, not eight (two of the grep hits were the word inside a comment), and they
  became one `noexcept` helper over `std::format` on the branch that still has them.
- **The earlier A/B on this branch measured a superset of this diff** and should not be read as this
  PR's result. For the record it resolved 0 of 24 timing tests against `48cadcb`, with one memory
  cell at 1.01x and 15 flat — but that was the instrument plus this, not this.

Reviewers who remember closed #232 `pr/layer-profile`: that was the instrument's ancestor, not this.

## Verification

`ctest -L unit` 220/220, `-L serial` 219/219, `-L mpi`, and the Python MPI suite at 1x1, 1x16, 2x8
and 8x16 — 600 passed at every layout, all suites green. Extension `3abe91dcace4eef73c43526d04e95a6e`.
`clang-format --output-replacements-xml` reports 0 replacements on every changed C++ file.

### DIFF
--- a/cpp/monoprop/MPGraph.cpp
+++ b/cpp/monoprop/MPGraph.cpp
@@ -52,6 +52,14 @@ auto layer_storage_memory_usage(const LayerCore &storage) -> GraphMemoryBreakdow
     breakdown.layer_storage_object_bytes = sizeof(LayerCore);
     breakdown.cross_rank_bytes = detail::cross_rank_storage_bytes(storage.cross_rank);
     breakdown.exchange_layout_bytes = detail::layer_exchange_layout_storage_bytes(storage.evolution_exchange_layout);
+
+    // Diagnostics. The graph does not partition -- its per-layer arrays are indexed by the FLAT world
+    // (ranks x partitions), so on a partitioned run these grow with a P the rank count never shows.
+    breakdown.slot_record_bytes = detail::cross_rank_slot_record_bytes(storage.cross_rank);
+    breakdown.layer_cores = 1;
+    breakdown.slot_records = storage.cross_rank.rank_count();
+    breakdown.occupied_slots = detail::cross_rank_occupied_slots(storage.cross_rank);
+    breakdown.cross_rank_endpoints = detail::cross_rank_endpoint_count(storage.cross_rank);
     return breakdown;
 }
 
--- a/cpp/monoprop/detail/graph/MPGraphViews.h
+++ b/cpp/monoprop/detail/graph/MPGraphViews.h
@@ -40,6 +40,14 @@ struct GraphMemoryBreakdown final {
     size_t cross_rank_bytes = 0;
     size_t exchange_layout_bytes = 0;
 
+    // Diagnostics, deliberately OUTSIDE total_bytes(): each is a count or a slice of cross_rank_bytes,
+    // separating the part of the graph that scales with the flat world size P from the traffic part.
+    size_t slot_record_bytes = 0;    // the slice of cross_rank_bytes that is one record per world slot
+    size_t layer_cores = 0;          // distinct LayerCores; slot_records / layer_cores recovers P
+    size_t slot_records = 0;         // summed over cores, so P per core
+    size_t occupied_slots = 0;       // slots carrying any traffic; / slot_records is the occupancy
+    size_t cross_rank_endpoints = 0; // traffic itself; P-independent ceiling on occupied_slots (>= 1 endpoint each)
+
     auto total_bytes() const -> size_t {
         return layer_descriptor_bytes + layer_storage_object_bytes + cos_data_bytes + cross_rank_bytes
                + exchange_layout_bytes;
@@ -52,6 +60,11 @@ struct GraphMemoryBreakdown final {
         cos_data_bytes += o.cos_data_bytes;
         cross_rank_bytes += o.cross_rank_bytes;
         exchange_layout_bytes += o.exchange_layout_bytes;
+        slot_record_bytes += o.slot_record_bytes;
+        layer_cores += o.layer_cores;
+        slot_records += o.slot_records;
+        occupied_slots += o.occupied_slots;
+        cross_rank_endpoints += o.cross_rank_endpoints;
         return *this;
     }
 };
--- a/cpp/monoprop/detail/graph_encoding/MPGraphEncoding.cpp
+++ b/cpp/monoprop/detail/graph_encoding/MPGraphEncoding.cpp
@@ -14,6 +14,7 @@
 
 #include "monoprop/detail/graph_encoding/MPGraphEncodingStorage.h"
 
+#include <algorithm>
 #include <format>
 #include <limits>
 #include <memory>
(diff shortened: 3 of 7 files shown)


cpp/monoprop/detail/Profile.h
```diff
@@ -0,0 +1,698 @@
...
+ *   total > gate > {apply, append}      (outside build_layer)
+ *   total > contract > {cacheop, cachestate, index, cacheshrink}
+ * `index` is accumulated from BOTH Scan.h and the caches step, so it appears under two parents. Three
+ * out-of-tree harness parsers key on `fold_s=` and friends, so those names never change. */
+inline auto emit_layer_line(std::FILE *out, bool want_layer, size_t index, const Slot &s) -> int {
+    if (!want_layer || !s.any_nonzero()) {
+        return 0;
+    }
+    // `row` is print order over the rows that printed, not the partition index, and is not stable run to run.
+    std::fprintf(out,
```
robertodr: c'mon... `fprintf`?? Use `std::print`

diagonal-hamiltonian: :robot: _AI text below_ :robot:

Fair, and it's gone — though I should be straight that it left this PR rather than being fixed in it.

**This PR no longer contains any of that code.** It has been cut down to the graph memory coverage alone (5 diagnostic counters on `GraphMemoryBreakdown` plus their tests); `Profile.h` and the whole `monoprop_PROFILE` instrument are held back as a separate commit, because whether we ship an internal profiler at all is a different decision from whether the graph can report its own slot cost. So `cpp/monoprop/detail/Profile.h` isn't in the diff you're looking at any more, and neither is the `fprintf`.

For the record on the branch that still carries it: it was **six** call sites, not eight — two of the grep hits were the word `fprintf` inside a comment — and they collapsed into one helper:

```cpp
template <typename... Args>
auto write_line(std::FILE *out, std::format_string<Args...> fmt, Args &&...args) noexcept -> void;
```

which `std::format`s into a string and does a single `fwrite`. Not `std::print`, for two reasons I'd rather have on the record than discover later: (1) two of the callers are **destructors** that can run while unwinding, so the emit path has to be `noexcept`, and `std::print` can throw on a format or I/O error; (2) one `fwrite` per line is what stops concurrent partition masters interleaving mid-line, which `std::print` doesn't guarantee. The `std::format_string` parameter still gets the compile-time check against the argument pack, which was the real content of your objection.

Happy to revisit if you'd rather see `std::print` with the destructor paths restructured — but it's not blocking this PR either way, since none of it is here.

diagonal-hamiltonian:

Closing this — the five counters have moved to the profiling-tool branch rather than to `main`, because nothing that ships reads them.

**What actually decides the memory results is kernel peak RSS** from `/usr/bin/time -v` around each rank, measured outside the code under test. The engine's own byte ledger is a secondary diagnostic and has disagreed with the kernel by up to 23x in both directions. So the case for landing `d_slot_record_bytes`, `d_layer_cores`, `d_slot_records`, `d_occupied_slots` and `d_cross_rank_endpoints` in `main` rested on them being useful to something downstream — and checked, the only consumers are five collators in the private benchmark harness (`sparse_collate.py`, `graph_world_report.py`, `graph_layout_collate.py`, `nocache_collate.py`, `sparse_stack_collate.py`). No library code, no benchmark, no shipping test outside this PR's own two suites. Diagnostics whose only readers are measurement tools belong with the measurement tools.

**Where the work went:** branch `tool/profile-instrument`, tag `tool/profile-instrument-v2` — one squashed commit on `main` `48cadcb` carrying the `monoprop_PROFILE` instrument *and* these counters, with both test suites intact. It is cherry-picked onto a branch under test, never merged. Nothing is lost: that tree is byte-identical to the pre-split branch, all at tree `321e8701a8995fe20b27d6f12be5aa2ebeb99329`.

**If a PR later needs these fields to make its own case, it should carry them.** One that already would: `perf/sparse-world-slots` (#238) added assertions on all five to `tests/test_monoprop_smoke.py`, since its whole claim is that per-rank graph memory falls as the world grows. Those counters are the natural way to show it, so they should ship *with* that argument rather than ahead of it.

**Two review threads this closes out:**

- @robertodr's `fprintf` point — already answered above: that code was never going to land here, and on the tool branch it is one `noexcept` helper over `std::format` with a single `fwrite` per line.
- The A/B figures quoted earlier on this PR measured a superset of the final diff and should not be read as anyone's result.

For the record, the diff was green throughout: `ctest -L unit` 220/220, `-L serial` 219/219, `-L mpi` and the Python MPI suite at 1×1, 1×16, 2×8 and 8×16 (600 passed each), 22/22 CI checks, and `clang-format` clean.

PR #254, opened 2026-08-20
ci: :construction_worker: add a bare-metal benchmark testbed via RunsOn

Panadestein:

## Summary

The benchmarks tracked by Bencher currently run on a shared GitHub-hosted runner. Its timing noise is why alerts on `main` are recorded but do not fail the build — the noise floor is wide enough that a real regression and a noisy neighbour look alike.

This adds a second testbed on a dedicated AWS node provisioned per run by [RunsOn](https://runs-on.com/). The node exists only while the workflow runs, so it is billed per commit to `main` rather than continuously, and nothing else is scheduled on it meanwhile. That lower noise floor is the point: it is what a testbed needs before its timing thresholds can be tightened enough to gate on.

To avoid a second copy of the ~100-line pipeline, the build/benchmark/upload steps move into a reusable workflow. Both callers now differ only in the three things that actually differ: the runner, the testbed name and the result label.

**The AWS side is deliberately not set up here** — that is still to be agreed. Everything in this PR is inert until it is, in two independent ways:

- The job is gated on a `BENCH_BARE_METAL` repository variable. Without a RunsOn stack answering the label the job would sit queued until it timed out, once per commit, so it is skipped rather than hung.
- On public repositories RunsOn reads `.github/runs-on.yml` from the **default branch only**, so the runner definition here has no effect until this PR merges.

Not Bencher's own [Self-Hosted Bare Metal Runners](https://bencher.dev/docs/) product, which is a much heavier thing (its own runner daemon, OCI images, Firecracker sandboxes). This repo only ever hands Bencher a JSON file over HTTP, and that does not change.

## Changes

- **`.github/workflows/bench.yml`** (new): the reusable build → benchmark → BMF → upload pipeline, taking `runner`, `testbed`, `label` and optional `bench-args`. Inputs reach the shell through `env:` rather than `${{ }}` interpolation.
- **`.github/workflows/bench_main.yml`**: now a caller — GitHub-hosted `ubuntu-26.04`, testbed `ubuntu-26.04`, label `ci-linux`. Triggers, concurrency and every `bencher run` flag are unchanged, so the existing history continues uninterrupted. It gains an explicit `contents: read` because a reusable-workflow call has to name the permissions it grants.
- **`.github/workflows/bench_bare_metal.yml`** (new): the gated caller — testbed `aws-c7i-4xlarge`, label `ci-bare-metal`.
- **`.github/runs-on.yml`** (new): the `bench` runner shape, `c7i` / 16 vCPU / 32 GB / gp3, on-demand.
- **`justfile`**: `bench-ci` takes `*ARGS` and forwards them to `pytest`.
- **Docs**: a "Testbeds" subsection in `docs/content/docs/benchmarks.mdx`, plus an `AGENTS.md` key-files entry.

## Notes for review

**Why every runner field is pinned rather than ranged.** Bencher keys history on (branch, testbed, measure), so a testbed name means one machine shape. Letting RunsOn pick from a family range, or fall back to spot, would let the hardware change under a fixed name — which shows up as a step change in every series rather than a data point. Spot would cost us twice: an interruption also loses the run outright.

**Problem sizes are unchanged, on purpose.** A bigger machine invites bigger problems, but the sizes cannot be calibrated against a node that does not exist yet, and once a testbed's history starts its sizes are fixed forever. So this adds the *capability* (`bench-args` → `bench-ci` → `pytest`) without baking in numbers; e.g. `-m ""` to include the slow fixed models. Worth doing as a follow-up once the node is real and we can see the timings.

**A new testbed starts empty.** Its first runs raise no alerts, and its thresholds want setting once there are enough points to see the noise. The `--thresholds-reset` flags in `bench.yml` apply per testbed, so the bare-metal one gets the same starting thresholds as the shared runner; tightening them is the follow-up that justifies the whole exercise.

**Deferred:** pinning `monoprop_NUM_THREADS` for the bare-metal node. An empty-string env var is worse than an unset one here, so it wants a real value chosen against the real machine.

## Checklist

No tests: the change is workflow plumbing. The `bench-ci` passthrough was verified locally against `pytest --collect-only` (extra args reach pytest; a later `-m` overrides the recipe's `-m "not slow"`).

## AI/LLM disclosure

- [x] I used the following tool to help write this PR description: GitHub Copilot (Claude Opus 4.5)
- [x] I used the following tool to generate or modify code: GitHub Copilot (Claude Opus 4.5)


.github/workflows/bench.yml
```diff
@@ -0,0 +1,114 @@
...
+
+    steps:
+      - uses: actions/checkout@v7.0.1
+        with:
+          fetch-depth: 0
+
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y just libopenmpi-dev openmpi-bin libboost-dev libboost-test-dev libmsgpack-cxx-dev
```
robertodr:
```suggestion
          sudo apt-get install -y just libopenmpi-dev openmpi-bin libboost-dev
```

robertodr: this slipped at the previous iteration, but the benchmarks are Python-only and we don't need to compile the C++ unit tests for it. We can skip installing these additional dependencies

.github/workflows/bench.yml
```diff
@@ -0,0 +1,114 @@
...
+
+      - name: Install the latest version of uv
+        uses: astral-sh/setup-uv@v10.0.0
+        with:
+          enable-cache: true
+          python-version: "3.11"
+
+      - name: Install package
+        run: |
+          uv sync --no-progress --group bench --all-extras -v
```
robertodr, line 65: you can set `monoprop_ENABLE_CXX_UNIT_TESTS` to OFF here (check that the name of the CMake toggle is correct 😅)

Beware that `uv sync` does NOT remember the CMake settings, so if you call `uv run` in the `just` recipes, you need to add `--no-sync` to avoid another rebuild (which would use default values for the cmake flags)

.github/workflows/bench.yml
```diff
@@ -0,0 +1,114 @@
...
+          BENCHER_PROJECT: ${{ vars.BENCHER_PROJECT || 'monoprop' }}
+          BENCHER_API_KEY: ${{ secrets.BENCHER_API_KEY }}
+          BENCHER_TESTBED: ${{ inputs.testbed }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          bencher run \
+            --branch main \
+            --adapter json \
+            --file bmf.json \
+            --github-actions "$GH_TOKEN" \
```
robertodr: can this work? one less secret in `env` is one less secret that might leak.
```suggestion
            --github-actions "${{ secretes.GITHUB_TOKEN }}" \
```

.github/workflows/bench_bare_metal.yml
```diff
@@ -0,0 +1,39 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false
+
+jobs:
+  bare-metal:
+    # Inert until a maintainer sets the BENCH_BARE_METAL repository variable to
+    # 'true'. Without a RunsOn stack behind the label the job would sit queued
+    # until it timed out, once per commit.
+    if: ${{ vars.BENCH_BARE_METAL == 'true' }}
```
robertodr, line 26: will we really need this once all the pieces of  the puzzle are in place?

.github/runs-on.yml
```diff
@@ -0,0 +1,22 @@
...
+  # this testbed's own history, so a different instance type or core count is a
+  # step change in the series rather than a data point -- widening any of these
+  # means retiring the testbed and starting a new one.
+  bench:
+    family: ["c7i"]
+    cpu: 16
+    ram: 32
+    image: ubuntu24-full-x64
+    volume: 40gb:gp3
+    # Spot would trade that comparability for cost twice over: an interruption
```
robertodr: can you make the prose of the comment a little less flowery? 💮

AGENTS.md
```diff
@@ -84,6 +84,12 @@ Key files:
   change to the recorded sections has to land on both sides of the package boundary. Benchmark
   names are Bencher's history key, so renaming or moving a `bench_*` test orphans its tracked
   series.
+- `.github/workflows/bench.yml`: the reusable build/benchmark/upload workflow; `bench_main.yml` and
```
robertodr: I don't think the agents need this information?

justfile
```diff
@@ -166,15 +166,17 @@ bench-smoke:
...
-bench-ci LABEL:
+bench-ci LABEL *ARGS:
     @mkdir -p "{{ bench_results }}"
-    label="$1"; \
+    label="$1"; shift; \
     monoprop_BENCH_LABEL="$label" monoprop_BENCH_RESULTS="{{ bench_results }}" \
         uv run --no-sync python -m pytest benches -o filterwarnings=default \
         --benchmark-json="{{ bench_results }}/time-$label.json" \
-        -m "not slow" --bench-rounds 5
+        -m "not slow" --bench-rounds 5 "$@"
```
robertodr: for the bare metal instance, I think we want to run the slow ones as well. Maybe with less rounds, since they're likely to be more stable

docs/content/docs/benchmarks.mdx
```diff
@@ -484,3 +484,42 @@ recorded but do not fail the build.
...
+|---|---|---|---|
+| `bench_main.yml` | GitHub-hosted `ubuntu-26.04` | `ubuntu-26.04` | `ci-linux` |
+| `bench_bare_metal.yml` | [RunsOn](https://runs-on.com/) EC2 node | `aws-c7i-4xlarge` | `ci-bare-metal` |
+
+The bare-metal workflow provisions a dedicated node per run, so the machine is
+billed per commit rather than continuously and nothing else is scheduled on it
+while the benchmarks run. That lower noise floor is the point: it is what a
+testbed needs before its timing thresholds can be tightened enough to gate on.
+
+It is skipped unless the `BENCH_BARE_METAL` repository variable is set to
```
robertodr, line 514: see my question above regarding the need for the repo variable

robertodr (commented):
this is very exciting 🤓 I have comments

.github/workflows/bench_bare_metal.yml
Panadestein: Well, if for some reason the benchmarks break we would be wasting resources with each commit to main. It is a kill switch in that case.

.github/workflows/bench.yml
Panadestein: Thanks for this, will fix!

.github/workflows/bench.yml
matteoacrossi: I don't see why it shouldn't

Panadestein: But this interpolation splices the token's literal content in the shell script, isn't that more dangerous?

.github/runs-on.yml
Panadestein: Fixed, you know who did this :rofl:

AGENTS.md
Panadestein: True, killed.

.github/runs-on.yml
matteoacrossi: Also, we should probably switch to spot once we know this is working. For these kinds of machines, and for short lived jobs, the likelihood of preemption is very low.

docs/content/docs/benchmarks.mdx
Panadestein: I commented above, what do you think? Perhaps I am being paranoid, but I thought having the possibility of killing the runs in AWS while development continues could be useful.

justfile
Panadestein: The single most important thing to fix here. Thanks!

.github/runs-on.yml
```diff
@@ -0,0 +1,22 @@
...
+  #
+  # Every field is pinned rather than ranged. Bencher compares each run against
+  # this testbed's own history, so a different instance type or core count is a
+  # step change in the series rather than a data point -- widening any of these
+  # means retiring the testbed and starting a new one.
+  bench:
+    family: ["c7i"]
+    cpu: 16
+    ram: 32
+    image: ubuntu24-full-x64
```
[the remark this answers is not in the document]
Panadestein: This one made sense. I fixed it by spawning an Ubuntu 26 machine instead.

.github/workflows/bench.yml
Panadestein: For some reason this doesn't show as outdated, but the fix was applied in [7984709](https://github.com/Algorithmiq/monoprop/pull/254/commits/79847095b8950022e7dcbaf582160e088172633d)

.github/runs-on.yml
Panadestein: Done!

.github/workflows/bench.yml
robertodr: the example in the documentation does not use `env` at all https://bencher.dev/docs/how-to/github-actions/#base-branch

Panadestein: You are right, I will fix that

diagonal-hamiltonian (approved):

PR #259, opened 2026-08-20
perf(evolution): ⚡ halve the follower-marking epoch stamp

diagonal-hamiltonian:

# Halve the follower-marking epoch stamp (u32 → u16)

Narrows `detail::MatchedEpochSet`'s per-term stamp to `uint16_t` — 2 B/term of operator row storage — with wrap handled explicitly and no shrink of the array anywhere. `perf/epoch-stamp-noshrink-v2` `1f350ff`: one commit of substance (`8dbe2a5`) plus merges of `main` and two review commits, 6 files +103/−11 against `origin/main`.

1. **The stamp width.** One stamp per term, for a counter that never leaves the struct — never serialised, never exchanged between ranks, never compared against anything but `cur_`. The wrap branch is now live (once per 65535 gate applications instead of once per 2^32) and the `std::fill` is what keeps it correct; `matched_epoch_stamp_wrap_reached_by_gate_count` cycles a whole period rather than assigning to `cur_`, which is the only way to prove the fill runs rather than that the branch is reachable.
2. **`matched_scratch_bytes` joins the memory breakdown, inside `total_bytes()`.** The stamp array is propagator-owned, so `estimate_memory_usage()` cannot see it and only `MonomialPropagator::operator_memory_usage()` can fill it in — the partitioned path sums per-partition breakdowns, so each picks up its own array.

## Memory: main `48cadcb` vs this branch, both `ENABLE_PROFILE=OFF`
Kernel `/usr/bin/time -v` peak RSS, summed over the ranks on the node, produced **outside** the code under test. 16 cells × 10 interleaved reps in one allocation, order flipped per (rep, cell); paired per-rep ratios, median of ratios; `agree` = reps pointing the same way.

| cell | main GiB | port GiB | port/main | agree |
|---|---:|---:|---:|---:|
| hubbard · A `1x128` · N=1 · `fresh` | 9.54 | 9.49 | 0.9948 | 10/10 |
| hubbard · A `1x128` · N=2 · `fresh` | 10.33 | 10.02 | 0.9702 | 10/10 |
| hubbard · B `8x16` · N=1 · `fresh` | 11.06 | 10.96 | 0.9911 | 10/10 |
| hubbard · B `8x16` · N=2 · `fresh` | 14.18 | 13.92 | 0.9813 | 10/10 |
| hubbard rung · A `1x128` · N=1 · `fresh` | 26.61 | 26.05 | 0.9788 | 10/10 |
| hubbard rung · A `1x128` · N=2 · `fresh` | 31.23 | 30.46 | 0.9750 | 10/10 |
| hubbard rung · B `8x16` · N=1 · `fresh` | 28.41 | 27.84 | 0.9797 | 10/10 |
| hubbard rung · B `8x16` · N=2 · `fresh` | 34.83 | 34.04 | 0.9772 | 10/10 |
| pauli · A `1x128` · N=1 · `fresh` | 10.39 | 10.31 | 0.9922 | 10/10 |
| pauli · A `1x128` · N=1 · `graph` | 18.13 | 17.93 | 0.9890 | 10/10 |
| pauli · A `1x128` · N=2 · `fresh` | 10.90 | 10.88 | 0.9987 | 10/10 |
| pauli · A `1x128` · N=2 · `graph` | 36.70 | 36.53 | 0.9952 | 10/10 |
| pauli · B `8x16` · N=1 · `fresh` | 12.17 | 12.03 | 0.9891 | 9/10 |
| pauli · B `8x16` · N=1 · `graph` | 19.68 | 19.48 | 0.9902 | 10/10 |
| pauli · B `8x16` · N=2 · `fresh` | 14.94 | 14.92 | 0.9969 | 8/10 |
| pauli · B `8x16` · N=2 · `graph` | 40.78 | 40.57 | 0.9951 | 10/10 |

**Peak RSS falls in 16 of 16 cells, 14 at unanimous 10/10, and 15 of 16 clear Holm** across the memory family — largest adjusted p = **0.043** (`pauli-B-N1 fresh`, 9/10); the only failure is `pauli-B-N2 fresh` (8/10), the smallest effect present. **That unanimity is node-sum only**: on worst-rank RSS two cells reverse direction (`pauli-A-N2 fresh` 1.0000, `pauli-B-N2 fresh` 1.0027). Hubbard beating pauli is what a per-term saving must do — but **2 B/term is a floor on the *array's* saving, not on the peak-RSS delta**, because peak RSS is a maximum over *time* and the stamp array need not be at its own maximum at that instant: hubbard `propagate` at N=1 implies only 0.54 and 1.05 B/term of node-sum delta, below the array's own floor, which is a statement about *when* the peak lands and not about the width.

## Timing: a null
**1 of 24 tests resolved**; the other 23 span 0.964–1.014x. The one survivor is `gradient[pauli]`, layout A `1x128`, N=1: **11355.2 → 11172.4 ms, 0.9836x, 10/10, Holm-adjusted p = 0.0469** — exactly at the boundary, and on an operation that reaches only the allreduce and never touches a table this diff changes, so it reads as a null either way. Absolute medians at N=1 below; the N=2 half and the full 24-row breakdown will follow in a comment. Grid: layout A = 1 rank/node × 128 partitions, B = 8 × 16, N = 1 and 2 nodes; hubbard `--hubbard-cutoff=10 --hubbard-lower-atol=1.25e-05` (96,981,051 terms), pauli `--pauli-cutoff=14 --pauli-lower-atol=5e-05` (91,273,861 terms), plus a `build_graph[hubbard]` rung at `--hubbard-trotter-steps=2` with **`--hubbard-lower-atol=1e-04`, not the main grid's `1.25e-05`**.

| cell | operation | main ms | port ms | port/main | agree |
|---|---|---:|---:|---:|---:|
| hubbard · A `1x128` | `propagate[hubbard]` | 17624.0 | 17600.3 | 0.9950 | 7/10 |
| hubbard · B `8x16` | `propagate[hubbard]` | 19194.4 | 19150.7 | 0.9974 | 6/10 |
| hubbard rung · A `1x128` | `build_graph[hubbard]` | 3976.1 | 3917.2 | 0.9928 | 6/10 |
| hubbard rung · B `8x16` | `build_graph[hubbard]` | 3093.7 | 3089.3 | 0.9988 | 6/10 |
| pauli · A `1x128` | `build_graph[pauli]` | 13947.9 | 14100.9 | 1.0065 | 9/10 |
| pauli · A `1x128` | `propagate[pauli]` | 11852.8 | 11755.8 | 0.9937 | 5/10 |
| pauli · A `1x128` | `energy[pauli]` | 2473.0 | 2418.7 | 0.9867 | 6/10 |
| pauli · A `1x128` | `gradient[pauli]` | 11355.2 | 11172.4 | **0.9836** | **10/10** |
| pauli · B `8x16` | `build_graph[pauli]` | 13300.9 | 13270.3 | 0.9984 | 8/10 |
| pauli · B `8x16` | `propagate[pauli]` | 10585.8 | 10533.2 | 0.9962 | 8/10 |
| pauli · B `8x16` | `energy[pauli]` | 2860.2 | 2860.8 | 1.0046 | 7/10 |
| pauli · B `8x16` | `gradient[pauli]` | 11583.9 | 11549.9 | 0.9988 | 7/10 |

## Caveats and scope
- **`total_bytes()` is not comparable across this commit**: a build without `matched_scratch_bytes` reports a total *lower* by roughly the stamp array while holding the same or more resident memory, so subtract the field or re-measure the baseline.
- **Not reproducible from this diff**: Deucalion, 2× AMD EPYC 7742 / 128 cores / SMT off / NPS4 / 242 GiB per node; a private benchmark harness that does not ship; two prebuilt venvs (main `1250a27e…`, port `6f14ecb1…`).
- **The measured binaries carry one mechanism this branch's own diff does not — and `main` now carries it too.** They were built from a cut that also released `init_op_map` in `get_operator()`. That mechanism shipped separately as #268 and is in this branch's base as of the merge below, so `get_operator()` here is `main`'s in both arms. Either way it is worth **1,189 B** (8357 → 7168 B in this campaign's own ledger) against cells of 9.5–40 GiB — order 1e-7, far below the smallest resolved effect. Re-gated, not re-measured.

## Gates
Re-gated on the merged tree (`09c5e67`, extension md5 `4a32d5c3…`; the two commits since are a doc line and a code comment, neither reaching the binary, `ENABLE_PROFILE=OFF`): `ctest -L unit` **224/224** and `-L serial` **223/223**, the four Python MPI layouts 1×1, 1×16, 2×8 and 8×16 at **592 passed each**, and `clang-format --output-replacements-xml` 0 replacements on every changed C++/binding file.

`main` has moved under this branch since the `48cadcb` measurement, and two of those commits are C++ rather than docs and build config: #267 bounds the epoch-stamp mark by `combined_size`, and #268 is the `init_op_map` release named in the caveat above. Neither is in the measured `main` arm. #268 is the 1,189 B already bounded there; #267 adds one compare per self-resolve hit and allocates nothing. The figures stand.

Merging `main` took three conflicts, all additive. Main modernised `MPOperatorMemoryBreakdown`'s member initialisers to `{0uz}`, so `matched_scratch_bytes` joins in that form; and each side had appended a test case at the same point in `evolution_detail_tests.cpp` and in `mp_operator_tests.cpp`, so both survive in each file. The wrap test and #267's guard test both pass on the merged build.


diagonal-hamiltonian:

## A/B campaign at this SHA

Campaign `pr7v2`, 12 cells × 10 interleaved reps, one allocation per cell, order flipped per (rep, cell). Ratios are the **median of per-rep paired ratios**, never a ratio of medians.

    base  bbd1473eccd55cecb22e9f5bf5637a72   main 48cadcb + #253 29ab66d + the profiling instrument aa77ed5
    port  802c69efbfe7e73994540a7a20dc7246   the same, plus f1f3067

**The result is memory, and it is consistent: peak RSS falls in 16 of 16 memory cells.** 15 of them at unanimous 10/10 rep agreement.

| cell | layout | N | main sum (GiB) | port sum (GiB) | ratio | agree |
|---|---|---|---:|---:|---|---|
| hubbard propagate | A 1×128 | 1 | 9.66 | 9.36 | **0.97x** | 10/10 |
| hubbard propagate | A 1×128 | 2 | 10.37 | 9.94 | **0.96x** | 10/10 |
| hubbard propagate | B 8×16 | 1 | 11.17 | 10.85 | **0.97x** | 10/10 |
| hubbard propagate | B 8×16 | 2 | 14.22 | 13.85 | **0.97x** | 10/10 |
| hubbard build_graph | A 1×128 | 1 | 26.60 | 26.07 | 0.98x | 10/10 |
| hubbard build_graph | A 1×128 | 2 | 31.22 | 30.36 | 0.97x | 10/10 |
| hubbard build_graph | B 8×16 | 1 | 28.40 | 27.84 | 0.98x | 10/10 |
| hubbard build_graph | B 8×16 | 2 | 34.86 | 34.05 | 0.98x | 10/10 |
| pauli fresh | A 1×128 | 1 | 10.44 | 10.33 | 0.99x | 10/10 |
| pauli graph | A 1×128 | 1 | 18.12 | 17.93 | 0.99x | 10/10 |
| pauli fresh | A 1×128 | 2 | 10.93 | 10.87 | 1.00x | 10/10 |
| pauli graph | A 1×128 | 2 | 36.71 | 36.53 | 1.00x | 10/10 |
| pauli fresh | B 8×16 | 1 | 12.18 | 12.04 | 0.99x | 10/10 |
| pauli graph | B 8×16 | 1 | 19.68 | 19.49 | 0.99x | 10/10 |
| pauli fresh | B 8×16 | 2 | 14.96 | 14.92 | 1.00x | 7/10 |
| pauli graph | B 8×16 | 2 | 40.78 | 40.58 | 0.99x | 10/10 |

Kernel peak RSS from `/usr/bin/time -v` per rank, summed over the node — measured outside the code under test. `sum` is the only figure comparable across layouts.

**Statistics.** A 10/10 sign test is p = 2·(1/2)^10 = 0.00195, the floor at this rep count. Holm step-down across the 16-test memory family: **15 of 16 survive**, largest surviving adjusted p = **0.031**. Only the pauli B/N=2 `fresh` cell (7/10, p = 0.34) does not, and it is the smallest effect in the table.

**Hubbard beats pauli, which is what a per-term saving must do.** Implied bytes/term at N=1, from the node-sum delta over the cell's term count:

| cell | terms | implied B/term |
|---|---:|---:|
| hubbard propagate A | 96,981,051 | 3.32 |
| hubbard propagate B | 96,981,051 | 3.54 |
| hubbard build_graph A | 278,722,051 | 2.04 |
| hubbard build_graph B | 278,722,051 | 2.16 |
| pauli fresh A | 91,273,861 | 1.29 |
| pauli graph A | 91,273,861 | 2.24 |
| pauli fresh B | 91,273,861 | 1.65 |
| pauli graph B | 91,273,861 | 2.24 |

The stamp width saves exactly 2 B/term at `capacity == size`, and up to ~4 B/term with geometric overshoot — which this PR deliberately does not shrink. Every cell lands inside that bracket, and the two Hubbard `propagate` cells sit highest, consistent with 29 Trotter steps growing the array repeatedly and retaining the overshoot. Two caveats on reading this table too literally: peak RSS is a **maximum over time** of the whole process, so the stamp array need not be at its own maximum at that instant — the implied B/term is an estimate, not an accounting — and it is why the pauli `fresh` cells can come out under 2.

**Timing is flat, as intended.** 0 of 24 timing tests resolved; every operation lands between 0.98x and 1.01x and nothing clears Holm at m = 24. `P(X ≥ 0)` = 1.00. PR 3, which adds no library code, resolved 0 of 24 on this same grid — that is the false-positive control.

**Placement verified per cell before any of the above was read**: both arms pinned 128 threads/rank at layout A and 16 at layout B, in all 12 cells, with identical term counts arm-to-arm. Slurm-confined ranks have silently defeated the engine's pinning before and every timing above it was junk.

### Two honest limits on this campaign

**1. Both arms are `ENABLE_PROFILE=ON` builds.** This branch's stack was developed on an instrumented base, so rather than pretend otherwise the instrument is in *both* arms, where it is common-mode. The arm-to-arm diff is line-for-line identical to `29ab66d..f1f3067` in all ten files (both hash `5a694b12`), so the **ratios isolate this PR** — but the **absolute** GiB and ms columns are a profiling build's and are not production figures. `monoprop_PROFILE` was unset, so no timer fired; the arms differ from a default build in codegen only.

**2. This campaign does not measure the `init_op_map` release at all.** Both grid models have single-site observables, where the map never grows past its initial buckets and the release is worth ~1.2 kB total. The 39.58 B/term (~1.15 GB) figure in the description comes from a large-observable workload, and no cell on this grid can corroborate it. So of the PR's three mechanisms, the campaign speaks to the stamp width and nothing else.

### Reproducibility

Not reproducible from this diff. Deucalion, 2× AMD EPYC 7742 / 128 cores / SMT off / NPS4 / 242 GiB per node; `--cpu-bind=cores --distribution=block:block`, `--cpus-per-task=128/ranks` on both arms; private harness at `4553c02`; the two prebuilt venvs hashed above.

diagonal-hamiltonian:

**Rebased back onto `main` and the base retargeted.** This was briefly stacked on #253; that PR has been withdrawn — its five graph memory counters had no consumer outside the private benchmark harness, so they moved to the profiling-tool branch (`tool/profile-instrument`, tag `tool/profile-instrument-v2`) instead of landing in `main`.

Consequence for this PR: it is back to **one commit on `main` `48cadcb`, 9 files +283/−25**, and the 6-line breakdown-key classification hunk moved to the tool branch beside the test that needs it. `main` has no `tests/test_memory_breakdown.py`, so nothing here enumerates breakdown keys. **The tree is identical to the version originally cut against `main`** and the mechanism never changed.

One correction to the A/B comment above, in the interest of not overclaiming: I said the arm-to-arm diff was line-for-line identical to the shipping diff. With the classification hunk no longer in this PR, the measured arms differ by **this diff plus those six lines in a Python test file**. A Python test cannot reach the compiled extension, so the two binaries still differ in exactly the nine files here — the ratios stand unchanged — but "identical" is now the wrong word for the diff.

Gates on this tree: `ctest -L unit` 221/221, `-L serial` 220/220, `-L mpi` and the Python MPI suite at 1×1, 1×16, 2×8, 8×16 (600 passed each), `clang-format` clean.

diagonal-hamiltonian:

## The OFF/OFF pair: the memory win holds, the headline figure does not

The measurement tables posted earlier were taken with the profiling instrument compiled into **both**
arms. That was necessary to attribute the change to a mechanism, but an instrument is not what ships,
so it leaves one question open: does the same pair measure the same way with it out? It does not,
quite — and the difference lands on the cells the headline quoted, so this supersedes that number.

**The pair.** `48cadcb` (main, and exactly this branch's parent) against this branch, both built
**OFF**, on the same 16-cell grid, 10 interleaved reps per cell in one allocation, order flipped per
`(rep, cell)`. Nothing was rebuilt: both arms were built and gated out of band, and the A/B driver
compiles nothing. Extension md5s `1250a27e…` (main) and `6f14ecb1…` (this branch); the latter carries
`matched_scratch_bytes` and not `d_slot_record_bytes`, which is how it is confirmed to be this PR
alone rather than a tree with the graph-counter work mixed in.

### Per cell, ON against OFF

Paired per-rep ratios, median of ratios, `port/main` — under 1.00 is this branch using less. Both
campaigns re-read with **one** script over both run directories, so the comparison is not itself
across two instruments.

| cell | ON | OFF | shift |
|---|---:|---:|---:|
| `hubbard-A-N1` fresh | 0.9695x | 0.9948x | +2.5 pts |
| `hubbard-A-N2` fresh | 0.9582x | 0.9702x | +1.2 pts |
| `hubbard-B-N1` fresh | 0.9712x | 0.9911x | +2.0 pts |
| `hubbard-B-N2` fresh | 0.9746x | 0.9813x | +0.7 pts |
| `hubbard-graph-A-N1` | 0.9800x | 0.9788x | −0.1 pts |
| `hubbard-graph-A-N2` | 0.9726x | 0.9750x | +0.2 pts |
| `hubbard-graph-B-N1` | 0.9805x | 0.9797x | −0.1 pts |
| `hubbard-graph-B-N2` | 0.9777x | 0.9772x | −0.1 pts |
| `pauli-A-N1` fresh | 0.9897x | 0.9922x | +0.3 pts |
| `pauli-A-N1` graph | 0.9894x | 0.9890x | −0.0 pts |
| `pauli-A-N2` fresh | 0.9954x | 0.9987x | +0.3 pts |
| `pauli-A-N2` graph | 0.9950x | 0.9952x | +0.0 pts |
| `pauli-B-N1` fresh | 0.9880x | 0.9891x | +0.1 pts |
| `pauli-B-N1` graph | 0.9903x | 0.9902x | −0.0 pts |
| `pauli-B-N2` graph | 0.9950x | 0.9951x | +0.0 pts |
| `pauli-B-N2` fresh | 0.9988x | 0.9969x | −0.2 pts |

**What holds.** Direction is unanimous in all 16 cells in both configurations, and **15 of 16 clear
Holm in both** — with the same single cell failing in each (`pauli-B-N2` fresh, the one cell whose
rep agreement is not unanimous either way). The saving is real and is not an artifact of measuring
with the instrument in.

**What does not.** The four hubbard `propagate` cells are **0.7 to 2.5 points weaker OFF**, all four
moving the same direction. Twelve of sixteen cells agree to within a few tenths of a point, so a spot
check on any of those twelve would have looked like confirmation. **The production figure for hubbard
`propagate` is 0.97–0.99x, not 0.96–0.97x**, and the body has been corrected. The `build_graph` rung
and pauli figures are unchanged.

**Why, honestly: not established.** An instrument is usually assumed to *dilute* a ratio, since it
adds bytes to both arms; amplifying instead would require its own state to scale with something this
PR also shrinks, which would put it on the epoch/stamp structures. That is a hypothesis and this
experiment does not test it. Two confounds block attributing the gap to the instrument alone: the ON
pair is a different pair of binaries, and one of its arms carries test-only lines that still change
the `.so`. It is also one group of four cells. It should not be carried forward as a mechanism.

### Timing is still a null

**1 of 24 resolved OFF**, against 0 of 24 ON: `gradient[pauli]` at layout A / N=1, 0.9836x, 10/10,
Holm-adjusted **p = 0.0469** — at the boundary. Across two 24-test families a single survivor at a
0.05 family-wise rate is what the correction is priced to allow, and `gradient` reaches only the
allreduce and never touches a table this diff changes. Read as the false positive it almost certainly
is. This PR's claim is memory, not speed.

### Instrument health, read before any timing

`distinct_pinned_cpus` is 128 at layout A and 16 at layout B in every cell, symmetric across arms
(80/80 observations each). Involuntary context switches run 0.88–1.24x port/main across all 12 cells,
against the ~59x that unpinning leaves on this machine — so no arm ran unconfined and no wall clock
here is unresolved.

**Not reproducible from this diff:** Deucalion, 2× AMD EPYC 7742 / 128 cores / SMT off / NPS4 / 242
GiB per node, private harness, two prebuilt venvs with the md5s above.

diagonal-hamiltonian:

_Measured at `b01e6fa`, which also carried an `init_op_map` release in `get_operator()`. That mechanism has since been split out of this PR, so the branch is now `8dbe2a5` and `get_operator()` matches `main`; the release was worth 1,189 B against cells of 9.5–40 GiB, so these figures stand. See the body's caveats._

## Full timing breakdown — all 24 tests, OFF/OFF

Campaign `pr7v2off`: main `src/mp-main-48cadcb` (`1250a27e…`, `48cadcb`) vs port `src/pr7v2` (`6f14ecb1…`, `b01e6fa`), 10 interleaved reps per cell. `port/main` is the median of the **per-rep paired** ratios, not the ratio of the two side medians, so it can disagree in sign with the `main ms`/`port ms` columns — `propagate[hubbard]` at A / N=2 is the one row where it does. `raw p` is one two-sided sign test in isolation, uncorrected.

| cell | N | operation | main ms | port ms | port/main | agree | raw p |
|---|---:|---|---:|---:|---:|---:|---:|
| hubbard · A `1x128` | 1 | `propagate[hubbard]` | 17624.0 | 17600.3 | 0.9950 | 7/10 | 0.344 |
| hubbard · A `1x128` | 2 | `propagate[hubbard]` | 46373.7 | 46670.9 | 0.9996 | 5/10 | 1.000 |
| hubbard · B `8x16` | 1 | `propagate[hubbard]` | 19194.4 | 19150.7 | 0.9974 | 6/10 | 0.754 |
| hubbard · B `8x16` | 2 | `propagate[hubbard]` | 13761.7 | 13843.4 | 1.0032 | 7/10 | 0.344 |
| hubbard rung · A `1x128` | 1 | `build_graph[hubbard]` | 3976.1 | 3917.2 | 0.9928 | 6/10 | 0.754 |
| hubbard rung · A `1x128` | 2 | `build_graph[hubbard]` | 8298.2 | 8312.4 | 1.0027 | 6/10 | 0.754 |
| hubbard rung · B `8x16` | 1 | `build_graph[hubbard]` | 3093.7 | 3089.3 | 0.9988 | 6/10 | 0.754 |
| hubbard rung · B `8x16` | 2 | `build_graph[hubbard]` | 2935.8 | 2807.5 | 0.9643 | 8/10 | 0.109 |
| pauli · A `1x128` | 1 | `build_graph[pauli]` | 13947.9 | 14100.9 | 1.0065 | 9/10 | 0.021 |
| pauli · A `1x128` | 1 | `propagate[pauli]` | 11852.8 | 11755.8 | 0.9937 | 5/10 | 1.000 |
| pauli · A `1x128` | 1 | `energy[pauli]` | 2473.0 | 2418.7 | 0.9867 | 6/10 | 0.754 |
| pauli · A `1x128` | 1 | `gradient[pauli]` | 11355.2 | 11172.4 | **0.9836** | **10/10** | 0.002 |
| pauli · A `1x128` | 2 | `build_graph[pauli]` | 28878.9 | 28858.4 | 0.9988 | 6/10 | 0.754 |
| pauli · A `1x128` | 2 | `propagate[pauli]` | 23310.2 | 23467.5 | 1.0142 | 7/10 | 0.344 |
| pauli · A `1x128` | 2 | `energy[pauli]` | 4659.1 | 4756.0 | 1.0120 | 8/10 | 0.109 |
| pauli · A `1x128` | 2 | `gradient[pauli]` | 15958.2 | 15720.6 | 0.9858 | 6/10 | 0.754 |
| pauli · B `8x16` | 1 | `build_graph[pauli]` | 13300.9 | 13270.3 | 0.9984 | 8/10 | 0.109 |
| pauli · B `8x16` | 1 | `propagate[pauli]` | 10585.8 | 10533.2 | 0.9962 | 8/10 | 0.109 |
| pauli · B `8x16` | 1 | `energy[pauli]` | 2860.2 | 2860.8 | 1.0046 | 7/10 | 0.344 |
| pauli · B `8x16` | 1 | `gradient[pauli]` | 11583.9 | 11549.9 | 0.9988 | 7/10 | 0.344 |
| pauli · B `8x16` | 2 | `build_graph[pauli]` | 9248.5 | 9250.4 | 1.0007 | 5/10 | 1.000 |
| pauli · B `8x16` | 2 | `propagate[pauli]` | 7197.5 | 7162.5 | 0.9944 | 8/10 | 0.109 |
| pauli · B `8x16` | 2 | `energy[pauli]` | 756.5 | 760.1 | 1.0010 | 5/10 | 1.000 |
| pauli · B `8x16` | 2 | `gradient[pauli]` | 4612.5 | 4607.9 | 0.9996 | 5/10 | 1.000 |

Holm across all 24 leaves **one** survivor, `gradient[pauli]` at A `1x128` / N=1, adjusted p = 0.0469. The next smallest raw p — `build_graph[pauli]` at A `1x128` / N=1, 0.021 at 9/10 — adjusts to 0.494.

### Worst-rank peak RSS, for the OOM question

Node-sum is the only cross-layout figure and is what the body tables; worst-rank is what an OOM is decided by. At layout A the two are equal (1 rank/node). The direction holds on worst-rank everywhere except two pauli cells, which reverse: `pauli-A-N2 fresh` 5.5067 → 5.5141 GiB (1.0000x, 5/10) and `pauli-B-N2 fresh` 0.9682 → 0.9703 GiB (1.0027x, 7/10). Both are unresolved and neither contradicts the node-sum result, but the body's "16 of 16" is a node-sum claim and does not transfer.

cpp/include/monoprop/MonomialPropagator.h
```diff
@@ -132,7 +132,11 @@ class MonomialPropagator {
         if (partition_group_) {
             return partitioned_operator_memory_usage_();
         }
-        return detail::estimate_memory_usage(mp_op_);
+        // matched_scratch_ is propagator-owned, so estimate_memory_usage() cannot see it; on a facade the
+        // partitioned path above already picks it up per partition.
+        auto breakdown = detail::estimate_memory_usage(mp_op_);
+        breakdown.matched_scratch_bytes = matched_scratch_.memory_bytes();
+        return breakdown;
```
robertodr: I really don't understand what is happening here?

docs/content/docs/benchmarks.mdx
```diff
@@ -273,7 +273,8 @@ For every engine, `results.json` collects, indexed by Trotter step:
 - `native_memory`: where an engine accounts for itself, its own figure (monoprop's C++
   operator-memory accounting, cuPauliProp's peak device memory over the step,
   `PauliPropagation.jl`'s `Base.summarysize` of the Pauli sum). These are reference values
-  only: they measure different things and must not be compared across engines. `QuEra ppvm`
+  only: they measure different things and must not be compared across engines. monoprop's
+  own total also steps up when a container gains a field that counted it nowhere. `QuEra ppvm`
```
robertodr: yeah, we don't need to know this
```suggestion
  only: they measure different things and must not be compared across engines. `QuEra ppvm`
```

robertodr (dismissed):
I just don't understand one bit, but otherwise seems harmless. It's good to be thrifty with type widths

cpp/include/monoprop/MonomialPropagator.h
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Fair — it was unexplained, and it looks like the line above it is being overwritten. Comment rewritten in `1f350ff`.

What it is: `matched_scratch_` is a member of **this class**, not of the operator. `detail::estimate_memory_usage(mp_op_)` takes the operator, so it has no way to see the stamp array and returns `matched_scratch_bytes == 0`; this level is the only one that holds both the number and the struct to put it in, so it fills the field in after the fact.

Why the field lives in the operator's breakdown struct at all, rather than being returned separately: the array is one stamp per operator row, so it belongs with the other per-term figures, and `total_bytes()` has to include it or the total under-reports resident memory. The partitioned branch two lines up needs no such patch — `partitioned_operator_memory_usage_()` sums each partition's own breakdown, and each partition is a propagator with its own array, so each summand already carries its own bytes.

The two tests pinning it are `mp_operator_breakdown_counts_matched_scratch_in_total_and_sum` (the field is summed by `total_bytes()` and accumulated by `operator+=`, and an operator on its own reports 0) and `mp_operator_breakdown_matched_scratch_nonzero_after_a_gate` (0 before any gate, nonzero after one, and the delta in `total_bytes()` equals the field).

docs/content/docs/benchmarks.mdx
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Agreed, and you already committed it as `c3fb2bc` — that file is now identical to `main` and out of this diff, which is down to 6 files. The aside was aimed at whoever compares a `native_memory` curve across this commit, but the caveat belongs in the PR body (where it is), not in the engine-comparison doc.

diagonal-hamiltonian:
Are you satisfied
@robertodr

robertodr (approved):

PR #260, opened 2026-08-20
chore(docs): ⬆️ move the API reference to griffe 2.x

matteoacrossi:

## Summary

Moves the API-reference pipeline to griffe 2.x, and in doing so removes an unsound dependency resolution.

griffe 2.0 split itself into three distributions: `griffe` is now a metapackage over `griffecli` + `griffelib`, and the importable `griffe` package ships in **`griffelib`**. `griffe-typingdoc` 0.3.1 switched to `griffelib>=2.0`, so a `griffe<2` pin does not actually keep griffe 2.x out — it invites *both* in. On #258's branch the lock resolved griffe 1.15.0 **and** griffelib 2.2.0, two distributions owning the same 44 files under `griffe/`; which one wins is install-order luck. (#258 shipped a `griffe-typingdoc<0.3` cap so it stays coherent on its own; this PR removes the need for that cap.) That failure mode is not hypothetical — while preparing this change, `uv sync` uninstalling `griffelib` deleted files griffe 1.15.0's RECORD also owned, leaving `import griffe` working but `griffe.parse_google` missing until a forced reinstall.

The upgrade itself is cheap. Every griffe symbol the walker uses is unchanged in 2.x — `load` (same 16-kwarg signature), `load_extensions`, `JSONEncoder`, `Expr`, `ParameterKind`, `parse_google`, and every model attribute read — so `gen_api_dump.py` needs no porting. `griffe-typingdoc` goes away entirely: it reads PEP 727 `Doc[...]` annotations and `src/monoprop` has none, so dumps with and without it are byte-identical. Bisecting the 2.x line, griffe 2.0.0 through 2.1.0 reproduce the 1.15 dump exactly; only 2.2.0 differs, by three hunks, and both are griffe bug fixes.

One of those fixes needed a companion change. griffe 2.2 correctly splits the `Note:` on `MonomialPropagator` out of the class description into an `admonition` section — but `convertClass` never rendered the docstring remainder, so the Note would have **silently disappeared from the page**. `convertClass` and `convert` now render it, which turns the Note into a `<Callout>`. `convertDoc` also gained a warning for section kinds it does not handle, since which sections griffe splits out is version-dependent and silent dropping is exactly how content goes missing across an upgrade.

## Changes

- `pyproject.toml` — replace `griffe>=1.15,<2` + `griffe-typingdoc>=0.2.8,<0.3` with `griffelib>=2.2,<3`; rewrite the pin comment (every clause of the old one is now wrong).
- `docs/scripts/gen_api_dump.py` — drop the `griffe_typingdoc` import and the `extensions=` argument. No other change.
- `docs/scripts/api-mdx.mjs` — render class- and module-level docstring remainders; warn on unhandled section kinds.
- `docs/content/docs/documenting.mdx` — fix a pre-existing claim that griffe resolves `[Symbol][]` links (it is `generate-api.mjs` that does).

Second commit, independent of the upgrade: qualify the ambiguous cross-reference targets in `features/cutoff.mdx` and `features/evaluation.mdx`. `build_graph`, `propagate` and `evolved_operator` each name three symbols, so the bare form was dropped from the ambiguous-leaf index and rendered as literal text. `evolved_operator` also moves to the two concrete front-ends, which is where it is documented — the abstract base has no page anchor for it.

## Verification

The generated tree is gitignored, so the evidence is below. Baseline captured from `main` (`412315e`), all diffs version-normalised for setuptools-scm.

**Dump — exactly the three expected hunks:**

<details><summary><code>diff before.json docs/monoprop.json</code></summary>

1. `MonomialPropagator.description` loses the trailing `Note:` block.
2. `MonomialPropagator.docstring` gains `{"kind": "admonition", "title": "Note", "value": {"annotation": "note", ...}}`.
3. `FermiOperator.terms` value: `[(t if isinstance(t, FermiString) else FermiString(t)) for t in terms]` → `[t if isinstance(t, FermiString) else FermiString(t) for t in terms]`.

</details>

**Rendered MDX — 2 files, 13 changed lines:**

<details><summary><code>diff -r before/api docs/content/docs/api</code></summary>

```diff
--- before/api/fermi/FermiOperator.mdx
+++ docs/content/docs/api/fermi/FermiOperator.mdx
-<PyAttribute name={"terms"} type={null} value={"[(t if isinstance(t, FermiString) else FermiString(t)) for t in terms]"} />
+<PyAttribute name={"terms"} type={null} value={"[t if isinstance(t, FermiString) else FermiString(t) for t in terms]"} />

--- before/api/monomial_propagator/MonomialPropagator.mdx
+++ docs/content/docs/api/monomial_propagator/MonomialPropagator.mdx
-Note:
-    Heisenberg evolution consumes each [`build_graph`](...) / [`propagate`](...) call's gates
-    back-to-front, so splitting one circuit across several calls is *not* equivalent to a
-    single call; in the Schrodinger picture (front-to-back) it is.
+<Callout title={"Note"} type={"note"}>
+
+Heisenberg evolution consumes each [`build_graph`](...) / [`propagate`](...) call's gates
+back-to-front, so splitting one circuit across several calls is *not* equivalent to a
+single call; in the Schrodinger picture (front-to-back) it is.
+
+</Callout>
```

</details>

- **Anchor set identical** — 143 `PyFunction`/`PyAttribute` ids, same set, so no `#member` fragment moved.
- **Cross-reference map identical** — 227 entries, byte-for-byte. This is the only artifact that would expose a prose-link regression, since `remark-xref.mjs`'s bare-name index silently drops leaves that map to two URLs.
- **Build clean** — `npm run build`, 150 pages, no prerender error; the `<Callout>` renders on `/api/monomial_propagator/MonomialPropagator` with its two xrefs intact.
- **`just doctest-py` and `just doctest-docs`** pass (they share the edited dependency group).
- **Resolution is coherent again** — `uv.lock` lists only `griffelib` 2.2.0; the venv has a single `griffe*` dist-info; `griffe.parse_google` no longer carries the 1.x `**options`.
- **Zero unresolved-xref warnings** — the first fully clean docs build here. Three were firing on `main` (two in `cutoff.mdx` since before #258, one in `evaluation.mdx` from #256's prose links); the second commit fixes them. They were never affected by the upgrade itself, since the xref map is byte-identical. Verified beyond the warning going away: all four references render as real links, and every `#fragment` target exists on its destination page.

## Follow-ups

Unchanged from #258's list, none of them unblocked or blocked by this (the prose-xref follow-up it mentioned is done here): the `returns_multiple_items` fix, the private-attribute filter, rendering the 19 `Raises:` sections (now at least *reported* rather than silent), dump pruning, sharing the pruning policy, and the `fdpy-` token rename.

## Checklist

## AI/LLM disclosure

- [x] I used the following tool to generate or modify code: Claude Code (dependency swap, the docstring-remainder rendering, and the before/after verification)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,13 +55,7 @@ dev = ["prek>=0.4", "ruff==0.15.20"]
 # The documentation site is built with Fumadocs (Next.js); see `docs/`.
 # Python's only roles are converting the tutorial notebooks to Markdown
 # (`nbconvert`) and dumping the API surface from docstrings (`griffe`, driven by
-# `docs/scripts/gen_api_dump.py` in the `gen-api` recipe). griffe 2.0 split
-# itself into three distributions: `griffe` is now a metapackage over
-# `griffecli` + `griffelib`, and the importable `griffe` package lives in
-# `griffelib` -- which is what we depend on, since the API reference never
-# shells out to the CLI. The 2.2 floor is where `Note:`-style admonitions are
-# detected in Google docstrings and where expressions stringify faithfully; both
-# show up in the generated pages.
+# `docs/scripts/gen_api_dump.py` in the `gen-api` recipe).
 docs = [
   "griffelib>=2.2,<3",
   "nbconvert",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,15 +55,15 @@ dev = ["prek>=0.4", "ruff==0.15.20"]
 # The documentation site is built with Fumadocs (Next.js); see `docs/`.
 # Python's only roles are converting the tutorial notebooks to Markdown
 # (`nbconvert`) and dumping the API surface from docstrings (`griffe`, driven by
-# `docs/scripts/gen_api_dump.py` in the `gen-api` recipe). griffe is held to the
-# 1.x line, and `griffe-typingdoc` (which reads the `Doc[...]` annotations) below
-# 0.3 with it: from 0.3.1 it depends on `griffelib`, the distribution griffe 2.x
-# ships its library in, and that would install a second copy of the `griffe`
-# package on top of this one. The floor is 1.15 because earlier 1.x releases
-# infer different return annotations, which changes the generated pages.
+# `docs/scripts/gen_api_dump.py` in the `gen-api` recipe). griffe 2.0 split
+# itself into three distributions: `griffe` is now a metapackage over
+# `griffecli` + `griffelib`, and the importable `griffe` package lives in
+# `griffelib` -- which is what we depend on, since the API reference never
+# shells out to the CLI. The 2.2 floor is where `Note:`-style admonitions are
+# detected in Google docstrings and where expressions stringify faithfully; both
+# show up in the generated pages.
 docs = [
-  "griffe>=1.15,<2",
-  "griffe-typingdoc>=0.2.8,<0.3",
+  "griffelib>=2.2,<3",
   "nbconvert",
   "pytest-markdown-docs",
   { include-group = "interactive" },
--- a/docs/content/docs/documenting.mdx
+++ b/docs/content/docs/documenting.mdx
@@ -163,8 +163,9 @@ def my_function(x: int) -> str:
 
 Docstrings may cross-reference other symbols with the same `[Symbol][]` /
 `[Display][fully.qualified.path]` syntax described in
-[Cross-references to the API](#cross-references-to-the-api); griffe resolves
-them into links when the reference is regenerated.
+[Cross-references to the API](#cross-references-to-the-api). `generate-api.mjs`
+resolves them into links when the reference is regenerated — griffe only records
+the docstring text.
 
 Inside a class docstring a bare `[name][]` only reaches that class's *own*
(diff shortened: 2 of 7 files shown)


pyproject.toml
```diff
@@ -55,15 +55,15 @@ dev = ["prek>=0.4", "ruff==0.15.20"]
...
-# ships its library in, and that would install a second copy of the `griffe`
-# package on top of this one. The floor is 1.15 because earlier 1.x releases
-# infer different return annotations, which changes the generated pages.
+# `docs/scripts/gen_api_dump.py` in the `gen-api` recipe). griffe 2.0 split
+# itself into three distributions: `griffe` is now a metapackage over
+# `griffecli` + `griffelib`, and the importable `griffe` package lives in
+# `griffelib` -- which is what we depend on, since the API reference never
+# shells out to the CLI. The 2.2 floor is where `Note:`-style admonitions are
+# detected in Google docstrings and where expressions stringify faithfully; both
+# show up in the generated pages.
```
```suggestion
# `docs/scripts/gen_api_dump.py` in the `gen-api` recipe).
```

robertodr (approved):
great stuff

PR #263, opened 2026-08-20
perf(evolution): ⚡ carry cross-rank queries as a width-adaptive record

diagonal-hamiltonian:

# Carry cross-rank queries as a width-adaptive record

`pr/query-wire-v3`, on `main` `999caf18`. The layer-build query path carried one dense word per 64
modes per query, at a fixed stride. This replaces it with a per-term-minimal record and stops the two
paths that never needed the wire from using it.

## What changes

1. **Cross-rank queries travel as a width-adaptive record.** The fixed stride the old `CompactQuery`
   design assumed is gone; a record is gap-coded ascending positions, sized by its own popcount.
2. **A rank resolves its own queries from positions**, not through the cross-rank path.
3. **`Sink::kStride` becomes `incoming_form()` / `querier_form()`**, fixing a latent inconsistency:
   `ContractSink::kStride` was the fused width while `on_response_block` got the plain buffer and
   used the plain-stride default — correct only by accident. A wrong form reads a neighbouring
   record's phase, which is a coefficient sign flip, not a crash.
4. **`bulk_insert` gains a hashed entry point**, so a miss found by position reuses the hash the
   probe already folded instead of rebuilding a monomial to re-fold it.

1 and 2 ship together because they share the `OperatorIndex` position store.

## Timings

Arms: `main` `999caf18` — this branch's own merge-base — against this branch, both
`ENABLE_PROFILE=OFF`. Extension md5s `b201ec44` (main) and `e39ec0db` (branch).

**`port/main`, so every row below 1.00 is this branch faster.** 10 interleaved reps per cell in one
allocation, order flipped per rep; ratios divided per rep then median, never a ratio of medians —
which is why the ratio column does not equal the two median columns divided.
`rung` is `lower_atol`, with Trotter steps where they differ, and separates cells that are different
model sizes. Every `p` is a raw two-sided sign test.

### Production layout (8 ranks x 16 partitions, 1 node)

| operation | layout | rung | main (ms) | port (ms) | port/main | agree | p |
|---|---|---|---:|---:|---:|---|---:|
| `build_graph[hubbard]` | B_8x16 | 1e-04/2st | 2752.8 | 1926.6 | **0.70x** (1.43x faster) | 10/10 | 0.00195 |

### Single thread (1 rank x 1 partition)

| operation | layout | rung | main (ms) | port (ms) | port/main | agree | p |
|---|---|---|---:|---:|---:|---|---:|
| `build_graph[hubbard]` | X_1x1 | 1e-3/2st | 41102.6 | 31110.0 | **0.76x** (1.32x faster) | 10/10 | 0.00195 |
| `propagate[hubbard]` | X_1x1 | 3e-5 | 48034.0 | 49811.2 | 1.04x | 9/10 | 0.0215 |
| `build_graph[pauli]` | X_1x1 | 1e-4 | 542.0 | 615.8 | **1.13x slower** | 10/10 | 0.00195 |
| `propagate[pauli]` | X_1x1 | 1e-4 | 408.9 | 489.9 | **1.20x slower** | 10/10 | 0.00195 |

## What these cells show, and where they do not

**The production win is the point, and it is intact.** `build_graph[hubbard]` at B_8x16 runs 1.43x
faster, 10 of 10 reps.

**`build_graph` wins at one rank too, and not for the reason an earlier revision of this description
gave.** That revision credited the X_1x1 gain to an inline `(k, d)` cutoff digest. This branch
*deletes* that digest path outright, and the cell got **faster** — 0.76x here against 0.899 there. So
the digest was not what made it win; it was the record, and plausibly the merge losing the `d`
bookkeeping from its inner loop. The digest bought ~0.8% of `build_graph` at 128 partitions, below
its own noise floor, and it is gone.

**`propagate` at one rank still costs.** Pauli is 1.13x / 1.20x, 10 of 10 each. The record's fixed
per-query cost needs something to amortise against, and at one partition there is no wire traffic to
save. This is not a narrow-system effect — the axis is amortisation, not width. It is also *smaller*
than it was: those cells were 1.43x and 1.54x before the follow-up commits and this cleanup.

**A correction to `RESULTS-qpos-stack.md`.** That file reads as though the follow-up commits removed
the X_1x1 regression. They reduce it by roughly a quarter, and it is still there. The stronger
reading came from composing a stack-vs-#263 ratio onto a #263-vs-main one; the two campaigns have
different baselines, so the product is not a measurement. The rows above are measured directly.

**Scope.** Four cells, chosen to answer two questions: does the production layout still win, and what
does one rank still cost. The A_1x128 rows, the 2-node rows, `energy`/`gradient` as an in-cell null
control, and peak RSS were measured on an earlier cut of this branch against `48cadcb` — a baseline
the rebase moved off — and are **not** re-measured here. Treat that earlier grid as evidence about
the commit it measured, not about this one.

## Gates

`ctest` on this branch: **unit 271/272, serial 270/270**. The single unit failure is
`monoprop_link_export_probe`, which fails **identically on untouched `main` `999caf18`** (243/244) —
it aborts with `MPI_Comm_size() called before MPI_INIT`. That is #313's own regression test failing
on `main`, not a defect of this branch, and it is worth a separate issue.

Python MPI suite: **600 passed** at each of ranks x partitions 2x1, 2x16, 4x8 and 16x16 (world 2 to
256), plus the C++ MPI ctest 1/1.

## Not reproducible from this diff

Deucalion, 2x AMD EPYC 7742 / 128 cores / SMT off / NPS4 / 242 GiB per node; a private harness that
never ships; two prebuilt venvs at the md5s above.


robertodr (commented):
I _think_ this duplicates what Claude&I have been doing in #226 🤔

robertodr:
you did indeed tell me to stack my stuff on top of this one

diagonal-hamiltonian:
Rebased onto current `main` (was 13 commits behind, at `cb9a033a`). The five commits are unchanged in
content; the pre-rebase tip is kept at the tag `archive/query-wire-v3-pre-rebase`.

The commit that mattered is **#270 (`cf84009e`), the exchange-layout / `HybridComm` rewrite** — it
derives the exchange layouts and stores only occupied world slots, and deletes `RecvLayout.h`. This
branch predated it, so it could not be benchmarked against anything current. Two conflicts, both
trivial: `AlgebraCommon.h` (keep both includes) and `cpp/tests/fused_query_codec_tests.cpp` (this
stack deletes it; `main` had only added a `TypeAliases` include, so the delete stands).

Verified after the rebase: **270/270 `ctest`**, and clean under real MPI at world 2 and 4.

#296 stacks GF(2)-linear rank routing on top of this and is now based on this branch.

cpp/monoprop/algebra/AlgebraCommon.h
```diff
@@ -155,6 +159,27 @@ template <size_t NumModes>
     return {(first_pair ^ second_pair).count(), active_mono.count(), (first_pair | second_pair).count()};
 }
 
+// The same sums from a (k, d) digest; no logical_num_modes because the masking above is inert for a
```
diagonal-hamiltonian: make clearer and 1 line

cpp/monoprop/algebra/AlgebraCommon.h
```diff
@@ -155,6 +159,27 @@ template <size_t NumModes>
...
+// well-formed monomial (every set bit at physical position >= 2 * (NumModes - logical_num_modes)).
+[[nodiscard]] inline constexpr auto cutoff_sums(size_t k, size_t d) noexcept -> CutoffSums {
+    return {k - (2 * d), k, k - d};
+}
+
+// d alone: mode m owns bits (2m, 2m+1) LSb0, so `w & (w >> 1)` masked to even bits counts each
```
diagonal-hamiltonian: simplify comment and shorten

cpp/monoprop/algebra/AlgebraCommon.h
```diff
@@ -170,6 +195,11 @@ auto length_cutoff(const Monomial<NumModes> &mono, unsigned int cutoff) -> bool
     return length_cutoff<NumModes>(mono, cutoff, NumModes);
 }
 
+// Digest form, on the same precondition as cutoff_sums(k, d). Width-independent, hence not templated.
```
diagonal-hamiltonian: remove

cpp/monoprop/algebra/AlgebraCommon.h
```diff
@@ -181,6 +211,11 @@ auto support_cutoff(const Monomial<NumModes> &mono, unsigned int cutoff) -> bool
     return support_cutoff<NumModes>(mono, cutoff, NumModes);
 }
 
+// Digest form, on the same precondition as cutoff_sums(k, d). Width-independent, hence not templated.
```
diagonal-hamiltonian: remove

cpp/monoprop/algebra/AlgebraCommon.h
```diff
@@ -243,6 +278,29 @@ class CutoffEvaluator {
         return cutoff_fn_(mono);
     }
 
+    // The decision from the (k, d) digest alone. The emit site gets both out of the partner merge, so
```
diagonal-hamiltonian: simplify or remove

cpp/monoprop/algebra/AlgebraCommon.h
```diff
@@ -243,6 +278,29 @@ class CutoffEvaluator {
...
+        if (length_cutoff_ != nullptr) {
+            return length_keeps(k, d, length_cutoff_->cutoff);
+        }
+        if (support_cutoff_ != nullptr) {
+            return support_keeps(k, d, support_cutoff_->cutoff);
+        }
+        return std::nullopt;
+    }
+
+    // The same decision when only the dense form is at hand, so d must be folded out of it.
```
diagonal-hamiltonian: remove

cpp/monoprop/core/SparseMonomial.h
```diff
@@ -0,0 +1,35 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+// The structural cutoffs over a monomial's (k, d) digest: k = popcount, d = modes carrying BOTH
```
diagonal-hamiltonian: remove cutoffs. Change digest to a better word everywhere or remove. remove caps words like BOTH

cpp/monoprop/detail/evolution/layer_build/Common.h
```diff
@@ -104,24 +104,8 @@ struct FusedContract {
...
-
-// The unsigned-int intermediate normalizes the ±1 sign bit into a fixed 32-bit pattern so the round-trip
-// is exact for any VecZ element width. Edit encode/decode as a pair.
-inline auto encode_phase(int phase) -> size_t {
-    return static_cast<size_t>(static_cast<unsigned int>(phase));
-}
-inline auto decode_phase(size_t word) -> int {
-    return static_cast<int>(static_cast<unsigned int>(word));
-}
+// Queries ride flat VecZ buffers in one VARIABLE-WIDTH format (SparseQuery): no stride exists, so every
```
diagonal-hamiltonian: remove

cpp/monoprop/detail/evolution/layer_build/Engine.h
```diff
@@ -70,7 +72,10 @@ inline auto append_inserted_endpoints(CosMask &cos_all, size_t combined_size, co
 template <size_t NumModes>
 struct GraphSink {
     static constexpr bool wants_values = false;
-    static constexpr size_t kStride = kQueryWords<NumModes>;
+    // Named apart: incoming_layout is what this rank RECEIVES, querier_layout its OWN send buffer. They
```
diagonal-hamiltonian: remove

cpp/monoprop/detail/evolution/layer_build/Engine.h
```diff
@@ -134,11 +139,16 @@ struct GraphSink {
         auto &out = acc[r].out_entries;
         const size_t base = out.size();
         const size_t nq = resp.size();
+        const QueryLayout layout = querier_layout();
         out.resize(base + nq);
+        // Forward walk, not indexing by q: a compact query's width depends on its own popcount.
```
diagonal-hamiltonian: remove

cpp/monoprop/detail/evolution/layer_build/Engine.h
```diff
@@ -186,7 +196,11 @@ struct GraphSink {
 template <size_t NumModes>
 struct ContractSink {
     static constexpr bool wants_values = true;
-    static constexpr size_t kStride = kQueryWordsFused<NumModes>;
+    // This rank RECEIVES fused (query+value) records, but the buffer on_response_block is handed is its
+    // own queries_r, which is PLAIN (build_fused writes the fused form into combined_qv_). Reading the
+    // phase with the wrong layout takes a neighbouring record's, which is a silent coefficient sign flip.
```
diagonal-hamiltonian: remove

cpp/monoprop/detail/evolution/layer_build/Engine.h
```diff
@@ -249,16 +263,19 @@ struct ContractSink {
             v_tgt = fused_scale ? op_coeffs[ip] * inv_cos : op_coeffs[ip];
         }
         else if (schrodinger) {
-            v_tgt =
-                is_paired<NumModes>(pr.mono[g]) ? algebra_state_phase<NumModes>(basis, pr.mono[g], state_mask_) : 0.0;
+            // Through the probe's accessors: it holds position lists, and mono_at builds a bitset only
```
diagonal-hamiltonian: remove or explain better if important

cpp/monoprop/detail/evolution/layer_build/Engine.h
```diff
@@ -249,16 +263,19 @@ struct ContractSink {
...
+            v_tgt = pr.is_paired_at(g) ? algebra_state_phase<NumModes>(basis, pr.mono_at(g), state_mask_) : 0.0;
         }
         else {
             v_tgt = 0.0; // Heisenberg fresh insert
         }
-        fc.cross_half[cross_base_ + g] = HalfRotationRec{ip,
-                                                         query_value<NumModes>(incoming[s], q),
-                                                         static_cast<int32_t>(pr.phase_of[g]),
-                                                         /*is_insert=*/ip >= pr.base};
+        // pr.off_of[g], not q: under the compact record a query ordinal does not name a buffer position.
```
diagonal-hamiltonian: remove

cpp/monoprop/detail/evolution/layer_build/Engine.h
```diff
@@ -297,8 +318,14 @@ struct ContractSink {
 // Owns build_layer's machinery over a compile-time Sink policy. combined_size = the pre-layer operator size.
 template <size_t NumModes, typename Sink>
 struct LayerBuildEngine {
+    // The store's position type, narrower than the wire's below 129 modes; decoded straight into.
```
diagonal-hamiltonian: remove comment

cpp/monoprop/detail/evolution/layer_build/PartnerMerge.h
```diff
@@ -0,0 +1,172 @@
+// Copyright 2026 Algorithmiq
```
diagonal-hamiltonian, line 1: Make the comments in this file clear and concise. Dont put optimisation comments. Just what the functions do.

cpp/monoprop/detail/evolution/layer_build/QueryCodec.h
```diff
@@ -0,0 +1,145 @@
+// Copyright 2026 Algorithmiq
```
diagonal-hamiltonian: Make the comments in this file clear and concise. Dont put optimisation comments. Just what the functions do and the big picture of why this is used. KISS

cpp/monoprop/detail/evolution/layer_build/Scan.h
```diff
@@ -157,23 +160,65 @@ inline auto rotation_dynamic_gate(std::optional<size_t> only_rotate_len_k,
     return true;
 }
 
+// M⊕G as the emit site needs it. The dense form is unavoidable -- the owner hash folds every word and
```
diagonal-hamiltonian: make more concise comment. Why unavoidable

cpp/monoprop/detail/evolution/layer_build/Scan.h
```diff
@@ -157,23 +160,65 @@ inline auto rotation_dynamic_gate(std::optional<size_t> only_rotate_len_k,
...
+    size_t k = 0;       // popcount(M⊕G)
+    size_t d = 0;       // modes of M⊕G carrying BOTH Majoranas
+    size_t overlap = 0; // slots in both M and G, which cancel
+    int phase_factor = 0;
+};
+
 // phase_factor is the basis-specific sign only: Majorana interleave_phase, still to be folded with
-// hermitian_phase at emit; Pauli pauli_rotation_sign, already rotation-ready.
-template <size_t NumModes, Algebra A>
+// hermitian_phase at emit; Pauli pauli_rotation_sign, already rotation-ready. `out_pos` receives the
```
diagonal-hamiltonian, line 174: make more concise dont mention profiling no caps

cpp/monoprop/detail/evolution/layer_build/Scan.h
```diff
@@ -184,6 +229,11 @@ struct FusedScanResult {
     // leader_src / follower_src. Empty when capture_values is false.
     std::vector<std::vector<double>> leader_val;
     std::vector<std::vector<double>> follower_val;
+    // Self-owned queries, staged as positions instead of encoded into leader_queries[my_rank]: they are
```
diagonal-hamiltonian: concise

cpp/monoprop/detail/evolution/layer_build/Scan.h
```diff
@@ -268,40 +318,66 @@ auto fused_find_and_collect(const MPOperator<NumModes> &op,
...
+        const OperatorIndex<NumModes> &ham = *op.store;
+        using RowPosT = typename OperatorIndex<NumModes>::PosT;
+
+        // The generator's positions, once per gate: the merge's second input.
+        std::vector<uint16_t> gen_pos;
+        gen_pos.reserve(gen_pop);
+        for (size_t b = gen.find_first(); b < gen.size(); b = gen.find_next(b)) {
+            gen_pos.push_back(static_cast<uint16_t>(b));
+        }
+        // 2*NumModes is the true bound: the partner's positions are distinct and below it. NOT
```
diagonal-hamiltonian: concise or remove

cpp/monoprop/detail/evolution/layer_build/SparseQuery.h
```diff
@@ -0,0 +1,325 @@
...
+#include <cassert>
+#include <cstddef>
+#include <cstdint>
+#include <vector>
+
+#include "monoprop/TypeAliases.h"
+
+namespace monoprop::detail {
+
+// A variable-width query record: one word-aligned record per term holding the term's ascending set-bit
```
diagonal-hamiltonian: concise or remove

cpp/monoprop/detail/evolution/layer_build/SparseQuery.h
```diff
@@ -0,0 +1,325 @@
...
+            uint64_t v = static_cast<uint64_t>(buf[base + word]) >> off;
+            if (off + width > 64) {
+                v |= static_cast<uint64_t>(buf[base + word + 1]) << (64U - off);
+            }
+            nbits += width;
+            return (width < 64) ? (v & ((uint64_t{1} << width) - 1U)) : v;
+        }
+    };
+
+    // ---- header -----------------------------------------------------------------------------------
```
diagonal-hamiltonian: remove this shit everywhere in this pr

cpp/monoprop/detail/evolution/layer_build/QueryWire.h
```diff
@@ -0,0 +1,325 @@
+// Copyright 2026 Algorithmiq
```
diagonal-hamiltonian, line 1: concise or remove comments on this. give general ide of why we need this and what its used for.

cpp/monoprop/detail/operator/OperatorIndex.h
```diff
@@ -251,12 +356,35 @@ class OperatorIndex {
     // Insert n distinct rows with consecutive indices [base, base+n). Rows must already be written.
     template <typename KeyFn>
     auto bulk_insert(size_t n, mapped_type base, KeyFn &&key_at) -> void {
+        if (n == 0) {
+            return;
+        }
+        // Delegating means both entry points share one insert loop, prefetch pipeline included.
+        bulk_insert_hashed(n, base, [&](size_t k) { return fold_hash(key_at(k)); });
```
diagonal-hamiltonian: Check failure on line R363
Explicitly capture the required scope variables.

cpp/monoprop/algebra/AlgebraCommon.h
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Gone — this file is no longer touched by the PR at all. The whole (k, d) cutoff fast path is deleted, so `AlgebraCommon.h` now diffs empty against `main`.

cpp/monoprop/algebra/AlgebraCommon.h
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Gone with the rest of the (k, d) path; the file is back to `main` byte for byte.

cpp/monoprop/algebra/AlgebraCommon.h
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Removed — the digest-form `length_cutoff` overload is deleted.

cpp/monoprop/algebra/AlgebraCommon.h
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Removed — the digest-form `support_cutoff` overload is deleted.

cpp/monoprop/algebra/AlgebraCommon.h
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Removed — `passes_from_digest` is deleted; the emit site is back to `passes_with_popcount`, which this PR had left dead.

cpp/monoprop/algebra/AlgebraCommon.h
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Removed with `passes_from_dense`.

cpp/monoprop/core/SparseMonomial.h
diagonal-hamiltonian: :robot: _AI text below_ :robot:

File deleted. "digest" no longer appears anywhere in `cpp/`, and neither do the shouted words.

cpp/monoprop/detail/evolution/layer_build/Common.h
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Removed.

cpp/monoprop/detail/evolution/layer_build/Engine.h
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Removed.

cpp/monoprop/detail/evolution/layer_build/Engine.h
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Removed.

cpp/monoprop/detail/evolution/layer_build/Engine.h
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Cut to two lines, and it is the one block I kept in this file: reading the wrong form takes a neighbouring record's phase, which is a silent coefficient sign flip rather than a crash.

cpp/monoprop/detail/evolution/layer_build/Engine.h
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Removed.

cpp/monoprop/detail/evolution/layer_build/Engine.h
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Removed.

cpp/monoprop/detail/evolution/layer_build/Engine.h
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Removed.

cpp/monoprop/detail/evolution/layer_build/PartnerMerge.h
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Rewritten: one line per function, saying what it does. The callgrind paragraph is gone. The triplicated loop body stays — it is measured — but nothing in the file says so any more.

cpp/monoprop/detail/evolution/layer_build/QueryCodec.h
diagonal-hamiltonian: :robot: _AI text below_ :robot:

The file is deleted. It was half pure forwarders, so it is merged into the record type, which is now `QueryWire<N>` in `QueryWire.h` — one header, one type, no `QueryRecord`/`CQ` aliases and no `QueryLayout` struct.

cpp/monoprop/detail/evolution/layer_build/Scan.h
diagonal-hamiltonian: :robot: _AI text below_ :robot:

One line now, and it answers the question: the owner hash folds every word and the basis sign reads the source bitset, so the dense form is built either way and the merge runs beside it.

cpp/monoprop/detail/evolution/layer_build/Scan.h
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Cut; no profiling, no caps.

cpp/monoprop/detail/evolution/layer_build/Scan.h
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Cut.

cpp/monoprop/detail/evolution/layer_build/Scan.h
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Removed.

cpp/monoprop/detail/evolution/layer_build/SparseQuery.h
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Header rewritten to a few lines: what the record is and why it exists. The entropy-bound and argmin material is gone.

cpp/monoprop/detail/evolution/layer_build/SparseQuery.h
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Gone everywhere — all four banner dividers and all five `//:` prefixes, which were a convention used nowhere else in `cpp/`.

cpp/monoprop/detail/evolution/layer_build/QueryWire.h
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Rewritten as above, and the file is now `QueryWire.h`.

cpp/monoprop/detail/operator/OperatorIndex.h
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Fixed: `[&]` -> `[this, &key_at]`. Swept the rest of the PR's diff for implicit captures; that was the only one.

diagonal-hamiltonian:

## Review pass: `4d9f477`

All 24 threads answered inline. The short version is that the review turned into a deletion.

### What went

| | |
|---|---|
| the `(k, d)` cutoff digest path | **~410 lines**, incl. `SparseMonomial.h` and two test files |
| `QueryCodec.h` | merged into the record, now `QueryWire<N>` in `QueryWire.h` |
| dead code | `push_mono`/`read_mono` (no production caller), `QueryRecord`, `CQ`, `kReserveWordsPerQuery` |

`AlgebraCommon.h` and `cpp/monoprop/core/CMakeLists.txt` now diff **empty** against `main` — this PR
no longer touches them at all. The diff is 23 files / +3080 down to 19 / +2270, and comment density
on the added lines is **22.7% → 8.9%**, against 13.1% measured on three untouched neighbours.

Three things fell out that were not the plan:

- Deleting `d` **dissolved the triplicated merge loop by itself** — the duplicated body *was* the `d`
  bookkeeping. The measured code path is unchanged; the callgrind comment justifying it is gone
  because there is nothing left to justify.
- Removing the dead dense-encode pair removed `QueryWire.h`'s last use of `Monomial`, which it named
  without including. It only ever compiled because callers included `Monomial.h` first.
- `passes_with_popcount` is live again. This stack had left it dead in production.

### What it cost: nothing, and one claim was wrong

Re-measured against `999caf18` — this branch's actual merge-base. The old tables were against
`48cadcb`, which the August rebase moved off.

| cell | port/main | previously |
|---|---|---|
| `build_graph[hubbard]` B_8x16 | **0.70x** (1.43x faster) | 0.702 — unchanged |
| `build_graph[hubbard]` X_1x1 | **0.76x** (1.32x faster) | 0.899 — *better* |
| `propagate[hubbard]` X_1x1 | 1.04x | 1.214 |
| `build_graph[pauli]` X_1x1 | 1.13x slower | 1.433 |
| `propagate[pauli]` X_1x1 | 1.20x slower | 1.542 |

The description used to credit the X_1x1 `build_graph` win to the digest. This branch deletes the
digest and that cell got **faster**, so the attribution was wrong. Every re-measured cell is the same
or better than what the PR previously claimed.

I also corrected the description's claim that the one-rank regression was removed. It is *reduced*,
by about a quarter, and pauli still costs 1.13x/1.20x at one rank x one partition.

### Gates

`ctest` **unit 271/272, serial 270/270**; Python MPI suite **600 passed** at each of 2x1, 2x16, 4x8
and 16x16 (world 2 to 256), plus the C++ MPI ctest 1/1.

### Two things for someone else

1. **`monoprop_link_export_probe` is red on `main`.** It fails identically on untouched `999caf18`
   (243/244) with `MPI_Comm_size() called before MPI_INIT` — #313's own regression test failing on
   the commit that added it. Not this branch's, but it means `-L unit` cannot go green on `main`
   here. Worth its own issue.
2. **#296 needs a rebase.** It is based on this branch and was already orphaned by the 2026-08-26
   rebase; this push moves the base again.

diagonal-hamiltonian:
@robertodr your back is scratched. Chop chop

robertodr:
I'm chopping wood like never before

cpp/monoprop/detail/evolution/layer_build/QueryWire.h
```diff
@@ -0,0 +1,332 @@
...
+#include "monoprop/TypeAliases.h"
+#include "monoprop/detail/evolution/layer_build/Common.h"
+
+namespace monoprop::detail {
+
+// One term's wire record for a cross-rank query: its ascending set-bit positions, gap-coded, plus a
+// phase. It replaces a fixed dense stride, which would spend one word per 64 modes no matter how few
+// bits a term actually sets.
+// Layout, LSB-first in word 0 onward: [2b phase][5b k, 31 escapes to a wider k][4b gap width gw]
+// [kPosBits first position][k-1 gaps of gw bits].
```
robertodr: ehm, what is the stuff in square brackets?

cpp/monoprop/detail/evolution/layer_build/QueryWire.h
```diff
@@ -0,0 +1,332 @@
...
+
+// `fused` records carry a trailing value word after the positions; `plain` ones do not. A named enum,
+// not a bare bool, so a form argument cannot read as plausibly correct either way.
```
robertodr:
```suggestion
// `fused` records carry a trailing value word after the positions; `plain` ones do not.
```

cpp/monoprop/detail/evolution/layer_build/QueryWire.h
```diff
@@ -0,0 +1,332 @@
...
+    static constexpr size_t kHeaderBits = kPhaseBits + kKBits + kGwBits;
+    static_assert(kPosBits <= (1U << kGwBits) - 1U, "gw <= kPosBits must fit the header's gap-width field");
+    static_assert(kHeaderBits + kLongKBits <= 64, "the widest header must be readable from word 0 alone");
+
+    static constexpr size_t kMaxPositions = kBits;
+
+    // Reserve hint only, for a caller batching many records into one flat position buffer.
+    static constexpr size_t kReservePositionsPerQuery = 6;
+
+    // Bit-packs fields into `buf`, one word at a time.
```
robertodr: Do I get it correctly that `Writer` is the C-ism equivalent to `bitset_to_indices`?

robertodr: if yes, I don't understand why we need it?

robertodr (changes requested):
I would ask two things:

1. It is not straightforward to get how the encoding for `QueryWire` works in practice. Can you add an explicit example using a Doxygen docstring?
2. Could you maybe try to tone down the C-isms? I'm not sure all of them are warranted, even in performance-sensitive contexts.

Also a concern, which might be me overthinking: there are a lot of `assert`s but these do nothing in `Release` so I don't know if we'll ever catch anything with them... Raising exceptions/returning `std::expected` seems also a bad idea though 🤷

cpp/monoprop/detail/evolution/layer_build/QueryWire.h
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Field widths in a packed record, LSB-first — `[2b phase]` meant "two bits of phase, then the
next field starts". Which is only legible if you already know that is what you are reading, so
it is gone.

`QueryWire` now carries a docstring that names each field with its actual bit range and works
one term end to end: positions `{3, 7, 8, 40}` at 128 modes, phase +1, giving `k = 4`, gaps
`{3, 0, 31}` and so `gw = bit_width(31) = 5`, laid out as

```
bits  0..1   phase   2   +1, biased by 1
bits  2..6   k       4   below kKEscape, so no wide-k field follows
bits  7..10  gw      5
bits 11..18  first   3
bits 19..23  gap     3   7 - 3 - 1
bits 24..28  gap     0   8 - 7 - 1
bits 29..33  gap     31  40 - 8 - 1
```

34 bits, so one word, against the four a 256-bit dense stride spends on that same term. That
contrast is the whole reason the type exists, and the escape and the per-record `gw` are what
keep it true for a two-bit term and a full-support one alike.

cpp/monoprop/detail/evolution/layer_build/QueryWire.h
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Taken — the "named enum, not a bare bool" sentence is gone and your line is the whole comment.

One deviation worth flagging: it is now `/*! @brief ... */` rather than `//`, because this pass
put Doxygen on the declarations in this header per `AGENTS.md`. The wording is yours verbatim.

cpp/monoprop/detail/evolution/layer_build/QueryWire.h
diagonal-hamiltonian: :robot: _AI text below_ :robot:

No — different job, and I should have made that visible in the header rather than leaving you to
infer it.

`bitset_to_indices` (`cpp/monoprop/algebra/AlgebraCommon.h:69`) turns a bitset into a heap
`VecZ` of **logical, MSb0** indices — `bs.size() - 1 - pos`, eight bytes each. `Writer` is a
variable-width **bit-field packer**: it appends 2-, 5-, 4-, `kPosBits`- and `gw`-bit fields into
64-bit words, so the whole record above is 34 bits in one word. Different output, different
index convention, and nothing to reuse — I checked, and the repo's other bit-packing
(`MPGraphEncodingStorage`'s packed phases, `InvertedIndex`'s row bitmaps, `CosMask`) is all
fixed-width one-bit-per-index, never variable-width fields. `Reader` is its inverse.

So `Writer`/`Reader` stay. What I did do everywhere else is drop the C-isms that were not
load-bearing — see the summary comment: pointer+length pairs and reference out-params are now
`std::span` and returned aggregates, which is already the idiom in `MPFunctions.h`,
`Exchange.h`, `InvertedIndex.h` and `Scan.h` itself.

One thing you have put a finger on indirectly: `Scan.h:311` open-codes the bitset→positions walk
instead of calling `bitset_to_indices`, and that *is* duplication. It is deliberate — the wire
path wants ascending **physical** positions as `uint16_t` with no allocation, and
`bitset_to_indices` gives logical MSb0 `size_t` in a fresh vector — but the reason was nowhere in
the code. Happy to add a helper for the physical/narrow variant if you would rather have one
spelling.

diagonal-hamiltonian:

## Review pass: `6c606f18`

All three asks addressed, plus the two open threads and the `Writer` question you self-resolved
without an answer. No wire-format change and no change to what any of this computes.

### 1. An explicit example, in a docstring

`QueryWire` gets a Doxygen block naming each field with its bit range and working one term end to
end — `{3, 7, 8, 40}` at 128 modes packs into 34 bits, one word, against the four a 256-bit dense
stride spends. The bracketed field list is gone. `AGENTS.md:28` already asked for Doxygen on
header declarations, which the file was not following, so this header now uses `/*! @brief */`
and `//!<` throughout. That does raise comment density here, against the previous round's
direction — deliberate, and only in this file.

### 2. C-isms

Pointer+length pairs and reference out-params become `std::span` / ranges and returned
aggregates. This is not new ground: `std::span` is already the idiom in `MPFunctions.h:47`,
`detail/mpi/Exchange.h:29`, `InvertedIndex.h:242` and `Scan.h:88,280`.

| was | is |
|---|---|
| `push(buf, const PosU *pos, size_t k, int phase)` | `push(buf, pos, phase)` |
| `read_query(buf, form, off, OutT *out, int &phase_out)` | `read_query(buf, form, off, out) -> Decoded{next, phase}` |
| `merge_partner_positions(a, ka, b, kb, out, size_t &overlap_out)` | `merge_partner_positions(a, b, out) -> MergedPartner{count, overlap}` |
| `find_batch_positions(pos_flat, pos_off, k_of, n, out, hash_out = nullptr)` | five spans; `n` is `pos_off.size()`, the optional tail is an empty span |
| `RowPositions{const PosT *pos; size_t count;}` | holds a `std::span` |
| `pos_flat.data() + pos_off[g]`, at three sites | `IncomingProbe::positions_at(g)` |

`gap_width`, `pair_count`, `read_positions`, `set_positions`, `row_eq_positions`,
`find_positions_`, `fold_hash_positions` and `SelfQueryStage::push` go the same way. The element
type stays deduced (a `contiguous_range`) rather than fixed, because `OperatorIndex::PosT` is
`uint8_t`/`uint16_t`/`uint32_t` by `NumModes` while the wire's is always `uint16_t` — that
genericity is real, not decoration. Every call site got shorter; they all passed
`.data(), .size()`.

Two things I kept, with reasons rather than by default: **`Writer`/`Reader`** (answered on the
thread — a variable-width bit-field packer, which `bitset_to_indices` is not and the repo has no
other instance of) and **`[[gnu::always_inline]]`**, which has seven prior uses on `main`.
`Resolve.h`'s `std::vector<double> *lv` is still a nullable-pointer-as-optional; I left it, as
it is an owning container rather than a view, but say the word.

### 3. The asserts — you were right, and it was worse than it looked

`main` carries 9 runtime asserts, all in `detail/`. This stack had added **24** on top. And
`pyproject.toml:131` pins `cmake.build-type = "Release"`, `CMakeLists.txt:41` defaults to it, and
every `cmake/compiler_flags/*.cmake` puts `-DNDEBUG` in `RELEASE` *and* `RELWITHDEBINFO` — while
`test.yml:239,248` and the justfile only ever run `ctest --test-dir build/editable/Release`. So
none of them executed anywhere, including CI. Only the coverage job keeps them live. clang-tidy
also runs against the Release `compile_commands.json`, so their bodies are not even linted.

Rule applied: **keep only what prevents undefined behaviour or a silently wrong encode at an
entry point; anything restating the code's own arithmetic or walking a stream becomes a test.**
24 added asserts → **7**; the branch total goes 33 → 16 against `main`'s 9, and `main`'s own 9 are
untouched. The 26 `static_assert`s all stay — compile-time, always on, and they are what pin the
header layout.

The one that mattered: `check_header` existed *only* to be called from an assert inside
`read_positions`, so it had never run. It moves to `cpp/tests/dense_query_reference.h` as
`wire_header_is_consistent` and is asserted from the differential — which also lets it check `k`
against the positions handed back, where the library version took `k` from the header it was
validating. 13 dead asserts became one check that runs in CI. The popcount identity dropped from
the emit site is already asserted in `partner_merge_tests.cpp:74`.

### Gates

`ctest -L serial` on a build made from this source: **`100% tests passed, 0 tests failed out of
270`**, 538 s. `clang-format --dry-run --Werror` clean on every touched file.

`monoprop_link_export_probe` remains the one red unit case, and it is red on `a24143fb` itself
(#313's own regression test, `MPI_Comm_size` before `MPI_Init`) — unchanged by this branch.

### Performance

You asked whether this slowed anything down, so I measured it rather than asserting it from
"semantically identical". It does not.

Interleaved A/B in one allocation, arms alternating with the order flipped every rep, 10 reps,
ratio taken as the **median of per-rep paired ratios**. `main` = `8697f6d9` (the branch as you
reviewed it), `port` = `6c606f18`. The arms are provably different binaries: `_core.so`
`e39ec0db…` vs `c27cb883…`, and `.text` 7,144,041 → 7,147,113 bytes, so this is not two copies
of one build being compared.

| layout | operation | main med | port med | port/main | raw p | Holm p | |
|---|---|---:|---:|---:|---:|---:|---|
| X_1x1 | `propagate[hubbard]` | 2440.9 ms | 2415.2 ms | **0.99x** | .0020 | .0117 | resolved, faster |
| X_1x1 | `propagate[pauli]` | 500.4 ms | 495.0 ms | **0.99x** | .0020 | .0117 | resolved, faster |
| X_1x1 | `build_graph[pauli]` | 627.0 ms | 618.8 ms | 0.99x | .0215 | .086 | unresolved |
| B_8x16 | `build_graph[pauli]` | 789.4 ms | 763.8 ms | 0.98x | .34 | .69 | unresolved |
| B_8x16 | `propagate[hubbard]` | 1949.1 ms | 1962.6 ms | 1.01x | .11 | .33 | unresolved |
| B_8x16 | `propagate[pauli]` | 699.9 ms | 700.4 ms | 1.01x | .75 | .75 | unresolved |

Holm across the family of six. **No cell resolves as slower.** Two resolve as ~1% faster, and I
am not going to claim that as a win: I cannot name the line that would cause it, the two B_8x16
`propagate` cells point the other way while unresolved, and a 3 KB shift in `.text` moves inlining
decisions on its own. The honest reading is "no detectable cost, with a hint of a small
improvement I can't attribute".

Peak RSS, kernel truth from `/usr/bin/time -v` around every rank: **1.00x** at B_8x16 and 1.01x at
X_1x1, both flat.

Provenance held on both cells, including the placement gate, which is a property of the layout and
was read off the baseline arm rather than declared: B_8x16 pins 16 NUMA-aligned threads per rank on
*both* arms, X_1x1 pins nothing on either.

Two limits worth stating rather than burying. The rung is **small** — hubbard 1.17M terms, pauli
223K, ~250 MB/rank — so this resolves a regression of a few percent, not a fraction of one; and
**`build_graph[hubbard]` is not covered**, because the suite skips it by its own guard (29
successive calls retain 29 layer-sets, over 242 GiB) unless `monoprop_BENCH_ALLOW_BIG_GRAPH=1`,
which at the default config would OOM the node.

### One for someone else

`harness/build.sh` broke on #314: passing `monoprop_ENABLE_MPI` as a config-setting gets the cmake
define but not the `build.requires` injection, because a `[[tool.scikit-build.overrides]]` `if.`
clause matches the environment only. Configure then dies in `src/monoprop/bindings/CMakeLists.txt`
while `mpi4py` sits in the project venv — the missing copy is the one in the isolated build
environment. Fixed on my side by exporting the variable; flagging it in case other tooling drives
the build the same way. The new `FATAL_ERROR` message is what made it a five-minute diagnosis
rather than an afternoon, so that part of #314 earned its keep.

robertodr (approved):

PR #265, opened 2026-08-21
feat(partition): 🧬 let derived propagators build correctly-typed partition children

ludmilaasb:

## Summary

`MonomialPropagator`'s constructor always built its partition children (the `S` single-partition propagators a partition facade owns) as the base `MonomialPropagator<NumModes>` type, never whatever derived class actually triggered the construction. On top of that, once the constructor decides to become a multi-partition facade it returns early, before a derived class's own constructor body/member-initializers would run — so a derived class's own state was never correctly set up on a facade in the first place.

Together these meant a derived subclass of `MonomialPropagator` could never safely support `partitions > 1`: its own added methods would see plain `MonomialPropagator` partition children instead of instances of its own type, and a copy of a partitioned facade would silently slice its children down to the base type too.

This PR adds two small, purely additive extension points so a derived class can opt into correct behavior under partitioning:

- An injected `PartitionChildFactory` constructor parameter (defaulted to `nullptr`, so every existing call site is unaffected) that a derived class can supply to build its own kind of partition child. This is deliberately a constructor parameter rather than a virtual method: a virtual "build my own child" hook called from inside the base constructor would always resolve to the base's own override, since a subobject's dynamic type is the base until its own constructor returns.
- A virtual `clone_()` hook, invoked only on fully-constructed objects (never during construction, so it doesn't hit the same trap), used by `PartitionGroup`'s copy constructor so copying a partitioned facade produces correctly-typed clones too.

Also widens the partition fan-out helpers (`for_each_partition_`, `map_partitions_`, `concat_partitions_`, `fold_partitions_`, `sum_partitions_`, `first_partition_`, `resolve_partition_count_`) from `private` to `protected`, and adds `is_partition_facade()` and `map_partitions_indexed_()`, so a derived class's own added methods can detect a facade and fan out through it using the same vocabulary the base class uses internally.

No existing behavior changes: every new parameter is defaulted, every widened/added member is additive, and the whole existing test suite passes unchanged.

## Changes

- `cpp/include/monoprop/MonomialPropagator.h`: add the `PartitionChildFactory` alias and trailing constructor parameter, add the virtual `clone_()` hook, widen the partition fan-out helpers to `protected`, add `is_partition_facade()` and `map_partitions_indexed_()`.
- `cpp/monoprop/detail/monomial_propagator/MonomialPropagator.inl`: thread the new constructor parameter through, falling back to the existing default factory when none is supplied; add `map_partitions_indexed_()`'s definition.
- `cpp/monoprop/detail/partition/PartitionGroup.h`: its copy constructor now clones each partition via the new virtual `clone_()` hook instead of directly naming `MonomialPropagator<NumModes>`.
- `cpp/tests/partition_group_clone_tests.cpp` (new): a toy derived `MonomialPropagator` subclass that supplies its own factory and `clone_()` override, confirming (via `dynamic_cast`) that both its partition children and a copy of its partitioned facade come out as the derived type.

## Checklist

- [ ] Documentation updated (docstrings, `docs/`, `CONTRIBUTING.md`) if needed

No public Python-facing API or behavior changes here (C++-only, additive extensibility hooks), so no `docs/` or Python docstring updates apply.

## AI/LLM disclosure

> [!IMPORTANT]


cpp/include/monoprop/MonomialPropagator.h
```diff
@@ -70,6 +70,17 @@ class MultiPartitionUnsupported : public std::runtime_error {
...
 public:
+    // Builds one partition child from a partition-local comm. When supplied, this is used instead
+    // of the base's own factory (which always builds a plain MonomialPropagator<NumModes>) -- a
+    // derived class supplies its own factory, by name, in its own constructor's mem-initializer
+    // list. This is deliberately NOT a virtual method called from inside this constructor: virtual
+    // dispatch during construction always resolves to this class's own override (a base subobject's
+    // dynamic type is this class until this constructor returns), so a virtual "make my own kind of
+    // child" hook would silently keep building plain MonomialPropagator objects no matter what
+    // derived constructor is running. Injecting the factory as an ordinary constructor parameter
+    // sidesteps that trap entirely.
```
ludmilaasb:
```suggestion
```

cpp/include/monoprop/MonomialPropagator.h
```diff
@@ -277,6 +289,14 @@ class MonomialPropagator {
     virtual auto update_initial_operator(const OperatorDict &op_dict) -> void { apply_initial_operator_(op_dict); }
 
 protected:
+    /// Deep-copy hook used when cloning a partition child (see PartitionGroup's copy constructor).
+    /// Unlike the child_factory constructor parameter above, this IS safe as a virtual call: it is
+    /// only ever invoked on an already fully-constructed, already-alive object, never from inside a
+    /// constructor body.
```
ludmilaasb:
```suggestion
```

cpp/include/monoprop/MonomialPropagator.h
```diff
@@ -380,6 +404,21 @@ class MonomialPropagator {
...
+    /// object's own (unpopulated) mp_op_/graph_/cutoff_fn_ directly.
+    auto is_partition_facade() const -> bool { return static_cast<bool>(partition_group_); }
+
+    /// Indexed variant of map_partitions_: `fn(r, partition)` runs once per partition, `r` in
+    /// [0, partition_count()), and the returned vector is indexed by `r`. Needed whenever a caller
+    /// must hand each partition a DIFFERENT pre-computed input (e.g. its own contiguous slice of a
+    /// term-indexed vector) rather than the same input to every partition -- map_partitions_'s
+    /// callback has no way to know which partition slot it is running in, and a shared mutable
+    /// counter would race across the partitions' concurrently-running master threads.
+    template <typename Fn, typename R = std::invoke_result_t<Fn &, int, MonomialPropagator &>>
```
ludmilaasb:
```suggestion
    auto is_partition_facade() const -> bool { return static_cast<bool>(partition_group_); }
    
    template <typename Fn, typename R = std::invoke_result_t<Fn &, int, MonomialPropagator &>>
```

robertodr (changes requested):
1. Would it be possible to consolidate the `private` and `protected` blocks? There's two of each now. Bonus points if you check whether such consolidation is needed also for `public`.
2. Why is `PartitionChildFactory` a new argument to the constructor? To ensure that the derived type resolves correctly?

matteoacrossi:
Can we remove the line breaks from the PR description?

ludmilaasb:
> 1. Would it be possible to consolidate the `private` and `protected` blocks? There's two of each now. Bonus points if you check whether such consolidation is needed also for `public`.

Sure!

>  2. Why is `PartitionChildFactory` a new argument to the constructor? To ensure that the derived type resolves correctly?

Yes, that's the idea!

robertodr (approved):

PR #264, opened 2026-08-21
test(cpp): ✅ migrate unit tests to Catch2

robertodr:
Since the unification of Python binding and C++ test compilation #152, the build times of the test step have increased. This is an experiment to check what's the impact of our unit test framework on compilation times in CI.

:robot: _AI text below_ :robot:

## Summary

Replace Boost.Test with Catch2 v3 for the C++ unit suite while preserving per-case CTest discovery and lockstep whole-suite MPI execution. This removes the Boost.Test binary dependency without removing Boost production utilities.

## Changes

- Port C++ test cases, fixtures, assertions, matchers, and generators to Catch2 v3.
- Use system-first Catch2 discovery with a pinned v3.15.3 CPM fallback.
- Preserve per-case serial CTest labels and whole-suite MPI rank variants, with deterministic ordering across ranks.
- Add Catch2 installation support and remove Boost.Test provisioning from development and CI environments.
- Update contributor and testing documentation.

## Testing

- Standard Python/workspace suite: 591 passed.
- Serial C++ suite: 208 passed, 2 expected single-rank skips.
- Two-rank MPI C++ suite: passed.
- Wide `TermIndex` configuration: 618 Python tests and 209 C++ tests passed, with 2 expected skips.
- Commit hooks: passed.
- Documentation build was not run because `npm` is unavailable in the development container.

## Checklist

- [x] Tests added or updated to cover the changes
- [x] Documentation updated (docstrings, `docs/`, `CONTRIBUTING.md`) if needed
- [x] `CHANGELOG` / release notes updated if applicable (not applicable: test infrastructure only)

## AI/LLM disclosure

- [ ] I did not use LLM tooling, or used it only privately for ideation
- [x] I used GitHub Copilot to help write this PR description
- [x] I used GitHub Copilot to generate or modify code

> [!IMPORTANT]
> By opening this PR I confirm that I have read [CONTRIBUTING.md](https://github.com/Algorithmiq/monoprop/blob/fa820c3e5a90f773417c367fedbf2bd50683b496/CONTRIBUTING.md) and I agree to the terms of the [Contributor License Agreement](https://github.com/Algorithmiq/monoprop/blob/fa820c3e5a90f773417c367fedbf2bd50683b496/CLA.md).

> [!WARNING]
> If you're contributing on behalf of your employer, contact [cla@algorithmiq.fi](mailto:cla@algorithmiq.fi) to arrange a Corporate CLA.

PR #269, opened 2026-08-21
feat(partition): 📈 report the launcher's placement, and delete the pinning knob

diagonal-hamiltonian:

## Summary

Adds a report-only diagnostic for the CPU placement the launcher actually gave each rank, and
deletes the knob that made pinning optional. `feat/commplace-placement-report` `7e376be`, 3 commits on
`cb9a033`, 7 files +218/−51.

1. **`monoprop_COMMPLACE` reports the placement the launcher actually gave each rank.** One
   `COMMPLACE` line per rank on stderr at propagator construction, naming the CPUs the launcher gave
   that rank and whether co-located ranks got disjoint masks (`masks=private` vs `masks=shared`);
   default off. The flag is a **noise gate, not a cost gate**, and it is worth being exact
   about that: the `MPI_Allgather` of the masks runs unconditionally — it is how `NodeMask` is
   picked, and it predates this PR — so what the flag avoids is one hwloc bitmap pass over the
   gathered rows plus a `std::format`, once per propagator construction. Measured volume is one
   line per rank per process (2 lines for a 2-rank bench), and `PartitionGroup` is built only
   under `n_partitions > 1`, so serial runs are silent either way — the whole Python suite emits
   nothing with the knob on. It earns its place by keeping a library from writing to stderr
   unbidden, not by saving work. The diagnostic earns a knob
   because `--cpu-bind=none` and `--cpu-bind=cores` give the same thread count and the same partition
   count, differing only in the mask the kernel enforces — a difference that has already cost a 1.45x
   tax nobody could see. A rank holding 16 of a host's 128 CPUs is equally "Slurm gave me my own 16"
   and "eight of us share these 16", and sampling `/proc` cannot separate them: it describes the
   process you are in, never a peer rank on the same host. `PartitionGroup` already exchanges the
   masks to pick a placement, so this reports what that exchange saw — no new collective. The flag
   gates printing only: the exchange is deliberately not gated on it, since the environment is
   per-rank, so a predicate over it is not rank-uniform, and one rank skipping a collective its peers
   entered is a hang, not a missing diagnostic. `format_place_line` returns the line rather than
   writing it, which is what keeps the formatting reachable in a test binary launched without the
   knob set.

   **Known gap, deliberately not closed here.** A flag cannot help someone who does not know to
   set it, and that is the person who loses the 1.45x. The case worth catching unprompted is
   `node_size > 1` with `masks=shared`; `partition_cpusets` three functions away already warns
   unconditionally via `std::call_once` when pinning is impossible, so both the precedent and the
   pattern exist. A follow-up should warn there. This PR ships the recording half only.

   Consequence of adding a boolean knob: **`off` now means off.** The shared `parse_flag` matched only
   the first *character* against `{0,f,F,n,N}`, so `monoprop_COMMPLACE=off` would have switched the
   diagnostic **on**. It now compares whole words, case-insensitively, against `0|false|no|off`,
   keeping its `(value, default_value)` signature so a default-ON flag stays expressible. Deliberate
   consequence: bare `f` and `n` were falsey and are now truthy; a test pins that, and is
   mutation-verified — restoring first-character parsing fails it.

2. **`monoprop_PARTITION_PINNING` is deleted; pinning is now unconditional.** Not because its parser
   was wrong, but because the experiment that would have used it is refuted: see below. No campaign
   ran on the wrong arm and none could have — no harness script ever set the variable, verified across
   the whole harness repo, so nothing needs re-auditing.

**Not shipped:** the placement deletion this was salvaged from. Its pre-registered falsifier was
*abandon if >5% cost anywhere*, and `propagate[hubbard]` at layout A/N=1 came in **2.90x slower, 10
of 10**, with 11 of 12 layout-A tests slower. Refuted by its own criterion — which is why the knob
can be deleted rather than fixed. (Correcting an earlier draft of that work: `main` places at
**both** layouts, 128 at A and 16 at B.)

**Two corrections to an earlier revision of this PR:**

- It listed `disabled` among the values the old parser read as ON *and the new one fixes*. `disabled`
  parses as ON under **both** rules, so it is not evidence of anything. Only `off`/`OFF`/`Off` change
  behaviour. With that removed, the parser change stops being "we found and fixed a bug" — the knob
  whose parser had the bug is deleted by this same PR, and nothing ever set it — and becomes what it
  is: a footgun removed from the knob being added. It is presented that way above.
- It claimed **zero** overlapping files with the open PRs. That holds against #259 and #263 only.
  **#264 rewrites both test files this PR touches** (`cpu_topology_tests.cpp`,
  `env_config_tests.cpp`) to Catch2, and **#265 touches `PartitionGroup.h`**. No semantic
  interaction in either case, but both are textual conflicts in one order or the other.

## Changes

- `monoprop_COMMPLACE`: one `COMMPLACE` line per rank on stderr at propagator construction,
  reporting the launcher's actual CPU mask and `masks=private`/`shared`/`alone`/`unknown`. Off by
  default; nothing is computed when off. `alone` is not `private` — nothing to be disjoint from is
  not evidence a multi-rank launcher did the right thing — and `unknown` is what a mask too wide for
  the exchanged window reports, rather than a plausible zero.
- `summarize_masks` is the one exported helper behind the line: `hwloc_bitmap_weight`/`_or`/
  `_list_snprintf` do the popcount, union and range formatting, in a TU that already includes
  `<hwloc.h>`. Truncation is stated as a trailing `,+` and cuts back to the last whole range, so the
  marker never follows a half-written CPU id.
- `monoprop_PARTITION_PINNING` deleted; pinning is unconditional, and `parse_flag` matches whole
  words instead of first characters.
- `docs/features/parallelism.mdx` swaps the row, states the boolean vocabulary once for the whole
  table instead of per row, and states that pinning is not configurable, with the reason.

## Measurement

Neither piece can move a number: the diagnostic is report-only and off by default, and the deleted
knob was never set by any harness script. No campaign ran, and none is presented here.

Gates, all on arm md5 `76eacbae`: `ctest -L unit` 228/228, `-L serial` 227/227, `-L mpi` 1/1, and
592 Python tests passed at each of four MPI layouts (2x1, 2x16, 4x8, 16x16 — worlds of 2, 32, 32 and
256). Counts are not comparable with the 222/221 an earlier revision quoted, nor with 225/224 from
before this branch merged `main` again: #259 added three cases.

Every new case asserts something, checked rather than assumed (`--report_level=detailed`):
`summarize_masks` 11, `place_line_reports_every_field` 1, `place_line_unknown_is_not_a_verdict` 1,
`parse_flag_falsey_words_in_any_case` 10, `parse_flag_truthy_words` 7,
`parse_flag_compares_whole_words` 9.

The `COMMPLACE` line is checked against ground truth rather than against a previous build, because
ground truth is the stronger check: under `srun -n2 --cpu-bind=cores` the two co-located ranks must
report `masks=private` with `node_cpus` = 2x `cpus`, and under `--cpu-bind=none` `masks=shared` with
`node_cpus` == `cpus`. Both arms give the same thread count and the same partition count, which is
the whole reason the line exists.

Measured, same binary as the gates:

```
--cpu-bind=cores  COMMPLACE rank=0 node_rank=0 node_size=2 masks=private cpus=64  node_cpus=128 cpu_list=0-127
                  COMMPLACE rank=1 node_rank=1 node_size=2 masks=private cpus=64  node_cpus=128 cpu_list=0-127
--cpu-bind=none   COMMPLACE rank=0 node_rank=0 node_size=2 masks=shared  cpus=128 node_cpus=128 cpu_list=0-127
                  COMMPLACE rank=1 node_rank=1 node_size=2 masks=shared  cpus=128 node_cpus=128 cpu_list=0-127
```

Note that `cpu_list` is identical in both arms — the job owns the whole node either way. `masks` and
`cpus` are what separate them, which is the case for reporting the peers' masks rather than sampling
`/proc`.

The parser change is likewise verified exhaustively rather than argued: of 18 representative values,
**7 change behaviour** — `off`/`OFF`/`Off` ON→off (the point), `f`/`n`/`0abc`/`nope` off→ON (bare
initials lost, prefix accidents tightened). `disabled` and `offbeat` are ON under **both** rules.

## Relationship to the open PRs

No semantic interaction with any open PR. Textual overlap: #264 rewrites both test files here to
Catch2, and #265 touches `PartitionGroup.h`; zero overlapping files with #259 or #263. The refuted
placement deletion this is salvaged from is unrelated to the placement fix that merged as #249.

## Checklist

## AI/LLM disclosure

> [!IMPORTANT]

> [!WARNING]


docs/content/docs/features/parallelism.mdx
```diff
@@ -29,7 +29,13 @@ whose partner lives in another partition are resolved through a per-gate exchang
...
 | `monoprop_NUM_THREADS` | one partition per physical core | Caps the number of partitions. Set it to run fewer partitions than cores. |
 | `monoprop_PARTITIONS` | `auto` | `auto` = one partition per core (capped by `monoprop_NUM_THREADS`); an integer `N` = exactly `N` partitions; `off` = one partition holding the whole operator. |
-| `monoprop_PARTITION_PINNING` | `on` | `0`/`false`/`no` disables pinning each partition to a core. Supported on platforms where hwloc can bind threads. |
+| `monoprop_COMMPLACE` | `off` | Report-only. One `COMMPLACE` line per rank on stderr when the propagator is built, naming the CPUs the launcher gave that rank and whether co-located ranks got disjoint masks. Changes no placement decision. |
+
+Boolean variables are off for `0`, `false`, `no` or `off` (whole word, any case)
+and on for any other non-empty value.
+
+Pinning each partition to a core is not configurable: leaving placement to the
+launcher measured `propagate[hubbard]` 2.90x slower, so the disabled arm is gone.
```
robertodr:
```suggestion
```

docs/content/docs/features/parallelism.mdx
```diff
@@ -29,7 +29,10 @@ whose partner lives in another partition are resolved through a per-gate exchang
 | --- | --- | --- |
 | `monoprop_NUM_THREADS` | one partition per physical core | Caps the number of partitions. Set it to run fewer partitions than cores. |
 | `monoprop_PARTITIONS` | `auto` | `auto` = one partition per core (capped by `monoprop_NUM_THREADS`); an integer `N` = exactly `N` partitions; `off` = one partition holding the whole operator. |
-| `monoprop_PARTITION_PINNING` | `on` | `0`/`false`/`no` disables pinning each partition to a core. Supported on platforms where hwloc can bind threads. |
+| `monoprop_COMMPLACE` | `off` | Report-only. One `COMMPLACE` line per rank on stderr when the propagator is built, naming the CPUs the launcher gave that rank and whether co-located ranks got disjoint masks. Changes no placement decision. |
```
robertodr: since it goes to `stderr`, I'd let it report always instead of adding a new env-var

robertodr (changes requested):
this looks good to me, it's just probing the hwloc masks to get a human-readable report for thread placement.

Changes:

1. The report is printed to `stderr` so there's no need to hide them behind a new env-var. The user can decide how to redirect the stream.
2. `clang-tidy` and SonarQube had some lints on `CpuTopology.cpp`. Some predate this PR, but we should fix them.

docs/content/docs/features/parallelism.mdx
diagonal-hamiltonian: agree

robertodr (approved):

robertodr:
I'll be very glad when we figure out how to _document_ how to do this from outside, instead of programmatically (which we seem to patch up for every new machine that pops up) :sweat:

docs/content/docs/features/parallelism.mdx
diagonal-hamiltonian: Both done in `928d31c`.

**1. `monoprop_COMMPLACE` is gone; the line is written unconditionally.** Verified as ground truth rather than as a diff — the probe now sets no knob at all, and a run that prints nothing is a failure:

```
--cpu-bind=cores  COMMPLACE rank=0 node_rank=0 node_size=2 masks=private cpus=64  node_cpus=128 cpu_list=0-127
                  COMMPLACE rank=1 node_rank=1 node_size=2 masks=private cpus=64  node_cpus=128 cpu_list=0-127
--cpu-bind=none   COMMPLACE rank=0 node_rank=0 node_size=2 masks=shared  cpus=128 node_cpus=128 cpu_list=0-127
                  COMMPLACE rank=1 node_rank=1 node_size=2 masks=shared  cpus=128 node_cpus=128 cpu_list=0-127
```

Two consequences worth naming rather than burying:

- **That was the last boolean env var**, so `parse_flag` and `iequals` had no caller left. Deleted, along with their four tests; `Settings` is now just `num_threads`. The whole-word `off` fix goes with them — it only ever existed to serve this knob, and the PR body's argument for it no longer applies. If you would rather keep the parser staged for the next bool, say so and I will restore it with its tests.
- **Volume.** `PartitionGroup` is built only under `n_partitions > 1`, so serial and `monoprop_PARTITIONS=off` runs stay silent — but a partitioned run now emits one line per propagator *construction*, not one per process. The C++ suite emits 19. `partition_cpusets` three functions away already uses `std::call_once` for its unconditional "pinning not possible" warning, so collapsing the repeats is the same idiom and one line; I left it out because it would lose fidelity when a process builds an `alone` propagator before an MPI one. Your call.

**2. Lints on `CpuTopology.cpp`: 10 → 0 clang-tidy, 1 → 0 SonarQube.** Pre-existing ones included, as you asked.

| Check | Sites |
| --- | --- |
| `modernize-return-braced-init-list` | `placement_order`'s return |
| `readability-qualified-auto` | 3x `auto *const topo` |
| `misc-misplaced-const` | `core`, `l3`, and the two `summarize_masks` bitmaps — `const hwloc_obj_t` is `hwloc_obj *const`, not a pointer to const |
| `modernize-use-designated-initializers` | `PhysicalCore{.cpu =, .l3_domain =}` |
| `modernize-use-integer-sign-comparison` | the truncation test, now `std::cmp_greater_equal` |
| `cpp:S886` | the `summarize_masks` row loop |

`cpp:S886` was the only one with any substance: the stop condition read `ok`, which the loop body writes. It now tests only the counter and `break`s, which also drops a duplicated `hwloc_bitmap_weight` call. Behaviour is unchanged — a zero row still discards the summary and reports `masks=unknown`.

The clang-tidy job is `continue-on-error: true`, so its green tick was never evidence either way; the number above is from the log — `CpuTopology.cpp`'s TU now reports `Suppressed 40547 warnings (40547 in non-user code)` with nothing unsuppressed.

Gates on arm md5 `edb00c44`: `ctest -L unit` 238/238, `-L serial` 237/237. MPI sweep running; I will update the PR body with it and re-request review.

PR #270, opened 2026-08-21
perf(graph,mpi): ⚡ derive the exchange layouts and store only occupied world slots

diagonal-hamiltonian:

## Summary

Per-layer graph memory no longer grows with the flat world size `P = ranks x partitions x nodes`.

1. **Only occupied world slots are stored.** `CrossRankPartnerRange` was a dense array over `P`
   (32 B narrow / 40 B wide). `CrossRankOccupiedSlot` is 12 B / 24 B and exists only where there is
   traffic, with offsets from a running prefix. One such array per layer made an `O(P)` cost `O(P^2)`
   per job.
2. **Exchange layouts are derived, not kept** — `O(P + occupied)` into per-thread scratch.
   `RecvLayout.h` and `resolve_recv` are gone: the count matrix is symmetric, so the recv layout *is*
   the send layout. That also removes a hazard — `resolve_recv` could skip a *collective* on a cache
   hit, letting ranks disagree about whether it ran. Nothing in the replacement can split ranks. Its
   width precondition is not lost: it survives as `mpi::check_exchange_layout_width`, on the path
   every exchange takes.
3. **The `HybridComm` staging matrix is re-indexed** peer-outermost, two owner-written tables publish
   before barrier B1, and `scatter_off_` becomes a cursor. Still 4 barriers per payload verb, and the
   offset tables are elementwise equal — integer sums reassociated, not an exchange reordered.
   Partition 0 stays `Theta(R*S^2)`: this is not a fix for the funnel.

`ContractSink::finalize` returns `nullptr`, so `propagate` builds no `LayerCore` — its win comes
entirely from mechanism 3, while `build_graph` / `energy` / `gradient` are where mechanism 1 shows up.

Also: `graph_memory_breakdown` in `binder.h` (five counters, each outside `total_bytes()` and inside
`operator+=`), and `ExchangeLayoutOracle.h` kept outside the library so it cannot drift into agreeing
with the code it checks.

## Reviewing this PR

Where the risk is, for anyone coming to this cold:

- **The symmetry argument in mechanism 2 is the load-bearing claim.** Deleting `resolve_recv` is only
  safe if the exchange count matrix is symmetric, so a rank can derive what it will receive from what
  its peers send. If that is wrong anywhere, the failure is a hang or silent corruption, not a failed
  assert. `cpp/tests/ExchangeLayoutOracle.h` checks the derivation independently, from outside the
  library.
- **`HybridComm.h` (mechanism 3) is the hardest file to review**, and the place where a reordering
  that looks equivalent may not be. The claim is unchanged barrier count and elementwise-equal offset
  tables, with reassociation of integer sums only.
- **The occupied-slot `find()`** is now `std::ranges::lower_bound` over a sorted array. It requires
  `occupied` to stay sorted by `slot`; a builder that appends out of order breaks lookup silently.
- **Everything else is mechanical**: a struct shrunk, an allocation removed, comments and docs cut.

## Measurement

**These figures supersede the tables posted earlier.** Those were measured at `e25bbfb`, on a binary
that still carried the since-dead PR 5 and PR 7. What follows is a fresh 12-cell campaign on `main`.

Setup: 10 interleaved reps, one Slurm allocation per cell holding both arms with the order flipped per
`(rep, cell)`. Ratios are formed per rep then medianed, judged by a two-sided sign test with Holm
step-down within each family. Peak RSS is `VmHWM` from `/usr/bin/time -v`, summed over nodes. Layout
A = 1 rank x 128 partitions, B = 8 ranks x 16 partitions, each at 1 and 2 nodes. Harness `ce49174`,
campaign row `pr6v2`.

- **main** `1250a27eded39f1639a6913ca3d8e101` @ `668c90b` — 216/216 unit, 215/215 serial (job 1840045),
  592 passed x 4 MPI layouts (1840046)
- **port** `97b9a114a6f0cac97eded7fe5c0d7b97` @ `446d60ca` — 230/230, 229/229, `-L mpi` 1/1 (1839668),
  593 passed x 4 layouts (1839669)

Checked before any timing was read: term counts identical to the term in every cell
(278,722,051 / 96,981,051 / 91,273,861, zero spread), thread placement symmetric across arms in all 12
cells (128/rank at A, 16/rank at B, zero mismatches), arms provably distinct binaries, and no MISSING
or FAILED rep anywhere.

**Wall time.** 13 of 24 tests resolve at the 10-rep sign-test floor, all of them improvements and none
a regression, split A=9 / B=4 so this is not one layout.

| cell | N | operation | port/main |
|---|---:|---|---|
| `grid-pauli-A` | 2 | build_graph[pauli] | **0.50x** |
| `grid-pauli-A` | 2 | gradient[pauli] | **0.52x** |
| `grid-pauli-A` | 2 | energy[pauli] | **0.52x** |
| `grid-hubbard-graph-A` | 2 | build_graph[hubbard] | **0.53x** |
| `grid-hubbard-A` | 2 | propagate[hubbard] | **0.57x** |
| `grid-pauli-A` | 2 | propagate[pauli] | **0.65x** |
| `grid-pauli-B` | 2 | gradient[pauli] | 0.80x |
| `grid-pauli-B` | 2 | energy[pauli] | 0.84x |
| `grid-pauli-B` | 2 | build_graph[pauli] | 0.84x |
| `grid-pauli-A` | 1 | build_graph[pauli] | 0.87x |
| `grid-pauli-A` | 1 | gradient[pauli] | 0.91x |
| `grid-hubbard-graph-A` | 1 | build_graph[hubbard] | 0.91x |
| `grid-pauli-B` 2 propagate[pauli] / `grid-hubbard-B` 2 propagate | 2 | propagate | 0.95x |

The remaining tests do not resolve at 10 reps, which means indistinguishable rather than a small win
in either direction.

**Peak RSS, node sum.** The graph cells are the headline:

| cell | N | port/main |
|---|---:|---|
| `grid-pauli-A` | 2 | **0.40x** (2.53x smaller) |
| `grid-pauli-B` | 2 | **0.46x** (2.18x smaller) |
| `grid-pauli-B` | 1 | 0.73x |
| `grid-pauli-A` | 1 | 0.71x |

`hubbard-graph` moves less (0.89x and 0.91x at N=2, ~0.97x at N=1) and plain `hubbard` is flat
(1.00x), as expected: `propagate` releases each layer as it contracts, so there is little retained
graph to shrink.

**Why every large win is at N=2.** The flat world is `ranks x partitions x nodes`, so 1 node to 2
doubles `P`, and what this PR removes is `P^2`-scaled waste in the per-layer arrays. Layout A at N=2
(`P=256`) is where the graph work is most `P^2`-dominated and it carries the biggest ratios; layout B
at the same `P` shows the same signs, smaller. The N=1 cells landing at 0.91-0.97x are the same effect
at `P=128`. This is an explanation consistent with all 24 points, not an isolated mechanism.

Direction and rough magnitude match the superseded tables — `build_graph[hubbard]` A/N=2 at 0.53x here
against 0.43x there, pauli graph peak RSS 0.40x against 0.39x. The timing absolutes are somewhat
smaller, which is what you would expect once the two dead PRs leave the baseline.

**Caveats.**

- Four cells carry an **unresolved** flag (8/10 or 7/10 agreement): `propagate[hubbard]` at A/N=1 and
  B/N=1, `build_graph[hubbard]` at graph-B/N=2, `propagate[pauli]` at A/N=1. The fix is more reps, not
  a re-read of the ratio.
- The pre-registered **commit-level bisect over 5 perf commits was never completed**, so this is
  attributed to the stack as a whole and to no named commit.
- The cache-coherence story for the layout dependence remains **inference, never isolated**.
- `staging_s` is not on this branch, so the funnel cannot be read as `staging_s + mpi_s` and the
  COMMPROF residual is invalid here — it goes negative at layout B. These are wall time and peak RSS.
- The benchmark harness does not ship, so this is not reproducible from the diff.

## Gate on the current head

Re-run on `3a87c66`, after the `main` merges and after the review fixes, because #267 and #268 both
touch files this PR changes:

- `ctest -L unit` and `-L serial`, and `-L mpi`
- the Python suite at four MPI layouts (worlds 2, 32, 32, 256)

All green. Counts and the tested binary's md5 are in a comment on this PR rather than inlined, so the
body cannot drift from what was run.

Next, not here: most world slots carry no traffic (28.0% occupancy at `P=16`, 16.6% at `P=128`), so a
neighbourhood collective would retire `MPI_Alltoallv`, `world_size` and the dense prefix sum, and
`pack_off_` is now the dominant serial term. Both false-share below `S=16` and need their own
campaigns.

## Checklist

## AI/LLM disclosure

> [!IMPORTANT]

> [!WARNING]


diagonal-hamiltonian:
## Campaign — re-measured on this cut

Fresh 12-cell campaign, **10 interleaved reps**, one allocation per cell holding both arms with the
order flipped per `(rep, cell)`. Harness `ce49174`, campaign row `pr6v2`.

- **main** `1250a27eded39f1639a6913ca3d8e101` @ `668c90b` — gated 216/216 unit, 215/215 serial (job 1840045), 592 passed × 4 MPI layouts (1840046)
- **port** `97b9a114a6f0cac97eded7fe5c0d7b97` @ `446d60ca` — gated 230/230, 229/229, `-L mpi` 1/1 (1839668), 593 passed × 4 layouts (1839669)

Preconditions checked *before* any timing was read: term counts identical to the term within every
cell (278,722,051 / 96,981,051 / 91,273,861, zero spread), thread placement symmetric across arms in
all 12 cells (128/rank at layout A, 16/rank at B), zero placement mismatches, arms provably distinct
binaries, and **no MISSING or FAILED rep anywhere**.

### Timing — 13 of 24 tests at the sign-test floor, and not one regression

`P(X >= 13)` = **1.47e-29** against 0.05 expected by chance. Split A=9 / B=4, so no single-layout
segregation. **Zero of the 24 timing tests resolved against the port**; every point is a win or flat.

| cell | N | operation | port/main |
|---|---|---|---|
| `grid-pauli-A` | 2 | build_graph[pauli] | **2.01x faster** (0.50x) |
| `grid-pauli-A` | 2 | gradient[pauli] | **1.93x faster** (0.52x) |
| `grid-pauli-A` | 2 | energy[pauli] | **1.92x faster** (0.52x) |
| `grid-hubbard-graph-A` | 2 | build_graph[hubbard] | **1.88x faster** (0.53x) |
| `grid-hubbard-A` | 2 | propagate[hubbard] | **1.75x faster** (0.57x) |
| `grid-pauli-A` | 2 | propagate[pauli] | **1.54x faster** (0.65x) |
| `grid-pauli-B` | 2 | gradient[pauli] | 1.26x faster (0.80x) |
| `grid-pauli-B` | 2 | energy[pauli] | 1.19x faster (0.84x) |
| `grid-pauli-B` | 2 | build_graph[pauli] | 1.18x faster (0.84x) |
| `grid-pauli-A` | 1 | build_graph[pauli] | 1.15x faster (0.87x) |
| `grid-pauli-A` | 1 | gradient[pauli] | 0.91x (resolved) |
| `grid-hubbard-graph-A` | 1 | build_graph[hubbard] | 0.91x (resolved) |
| `grid-pauli-B` | 2 | propagate[pauli] / `grid-hubbard-B` 2 propagate | 0.95x (resolved) |

Read the labels carefully: a bare `0.91x` is still a ~9% win for the port — the "Nx faster" label
only attaches above a threshold. Nothing in this table points the other way.

### Memory — peak RSS, kernel truth

The graph cells are the headline, and they replicate the withdrawn figure almost exactly:

| cell | N | port/main (peak RSS) |
|---|---|---|
| `grid-pauli-A` | 2 | **0.40x** (2.53x smaller) |
| `grid-pauli-B` | 2 | **0.46x** (2.18x smaller) |
| `grid-pauli-B` | 1 | 0.73x |
| `grid-pauli-A` | 1 | 0.71x |

The `hubbard-graph` cells move less — 0.89x and 0.91x at N=2, ~0.97x at N=1 — and the plain
`hubbard` cells are flat (1.00x), which is expected: `propagate` releases each layer as it contracts,
so there is little retained graph to shrink.

### The N=2 pattern is the mechanism, not a coincidence

Every large win is at **N=2**. That is exactly what this change predicts: the flat world is
`ranks x partitions x nodes`, so going from 1 node to 2 doubles `P`, and what this PR removes is
`P²`-scaled waste in the per-layer arrays. Layout A at N=2 (`P=256`) is where the graph work is most
`P²`-dominated and it carries the biggest ratios; layout B at the same `P` shows the same signs at
smaller magnitude. The N=1 cells resolving at 0.91–0.97x rather than flat is the same effect at
`P=128`.

I am labelling that an explanation consistent with all 24 points, not an isolated mechanism — see the
carried-forward caveats below.

### Versus the withdrawn numbers

The old writeup claimed 16 of 24 resolved with `build_graph[hubbard]` A/N=2 at 2.34x and pauli graph
peak RSS 0.39x, measured on a binary carrying the dead PR 5 and PR 7. This cut, on `main`:
**13 of 24 at the floor**, `build_graph[hubbard]` A/N=2 at **1.88x**, pauli graph peak RSS **0.40x**.
Direction and rough magnitude survived; the timing absolutes are somewhat smaller, which is what you
would expect once the two dead PRs are no longer in the baseline. **These are this cut's numbers and
supersede the old ones.**

### Caveats, unchanged

- Four cells carry an **unresolved** flag (8/10 or 7/10 agreement): `propagate[hubbard]` at A/N=1 and
  B/N=1, `build_graph[hubbard]` at graph-B/N=2, `propagate[pauli]` at A/N=1. The fix is more reps, not
  a re-read of the ratio.
- The pre-registered **commit-level bisect over 5 perf commits was never completed**, so this is
  attributed to the stack as a whole and to no named commit.
- The cache-coherence story for the layout dependence remains **inference, never isolated**.
- `staging_s` is not on this branch, so the funnel cannot be read as `staging_s + mpi_s` and the
  COMMPROF residual is invalid here — it goes negative at layout B. These figures are wall time and
  peak RSS.

Not reproducible from the diff: the benchmark harness does not ship.

cpp/monoprop/detail/graph/MPGraphViews.h
```diff
@@ -40,6 +40,16 @@ struct GraphMemoryBreakdown final {
...
+    // Diagnostics, deliberately EXCLUDED from total_bytes(): folding any in would silently redefine
+    // graph_memory_bytes(). The split matters because slot_record_bytes is the part sized by the flat
+    // world (see CrossRankOccupiedSlot) and the endpoint count below is the part sized by traffic.
+    size_t slot_record_bytes = 0; // one record per STORED world slot -- occupied only, once sparse
+    size_t layer_cores = 0;       // distinct LayerCores walked (shared cores counted once)
+    size_t slot_records = 0;      // the flat world P per core, so slot_records / layer_cores == P
+    size_t occupied_slots = 0;    // slots carrying any traffic: occupancy = occupied_slots / slot_records
+    // Cross-rank endpoints: the traffic itself, and the ceiling on occupied_slots.
+    size_t cross_rank_endpoints = 0;
+
```
diagonal-hamiltonian, line 49: reduce comments, no caps

cpp/monoprop/detail/graph_encoding/MPGraphEncoding.cpp
```diff
@@ -100,21 +84,45 @@ auto packed_phase_storage_bytes(const PackedPhaseStorage &storage) -> size_t {
...
-        range.sin_recv_count = static_cast<TermIndex>(partner.sin_recv_entries.size());
-        range.in_count = static_cast<TermIndex>(partner.in_count);
+        // The record stores one count and one offset for both B and D, so their equal length is a
+        // precondition: a skew would not throw, it would mis-derive Q and read the wrong endpoint.
+        if (partner.sin_send_indices.size() != partner.sin_recv_entries.size()) {
+            throw std::logic_error(std::format(
+                "Cross-rank slot {} has {} send endpoints against {} recv endpoints; B and D are the same set.",
+                rank,
+                partner.sin_send_indices.size(),
+                partner.sin_recv_entries.size()));
```
diagonal-hamiltonian, line 99: fix sonar Check warning on line R99
Define and throw a dedicated exception instead of using a generic one.
Check warning: Define and throw a dedicated exception instead of using a generic one.

See more on https://sonarcloud.io/project/issues?id=Algorithmiq_monoprop&issues=AaAyuNRba8KXQfRGAUNu&open=AaAyuNRba8KXQfRGAUNu&pullRequest=270
/ SonarCloud Code Analysis

cpp/monoprop/detail/graph_encoding/MPGraphEncoding.cpp
```diff
@@ -100,21 +84,45 @@ auto packed_phase_storage_bytes(const PackedPhaseStorage &storage) -> size_t {
...
+        // P (the in-block) is a boundary WITHIN B. Q = P+Q - P is computed in unsigned width by
+        // slot_sin_recv_index, so a P past the end yields not a negative Q but one near 2^64, which
+        // every index compares less than -- and every D read then addresses B off the end.
+        if (partner.in_count > partner.sin_send_indices.size()) {
+            throw std::logic_error(
+                std::format("Cross-rank slot {} declares an in-block of {} inside {} endpoints; the in-block is a "
+                            "boundary within the endpoint list, not an addition to it.",
+                            rank,
+                            partner.in_count,
+                            partner.sin_send_indices.size()));
```
diagonal-hamiltonian, line 109: sonar Define and throw a dedicated exception instead of using a generic one.
Check warning: Define and throw a dedicated exception instead of using a generic one.

See more on https://sonarcloud.io/project/issues?id=Algorithmiq_monoprop&issues=AaAyuNRba8KXQfRGAUNv&open=AaAyuNRba8KXQfRGAUNv&pullRequest=270

cpp/monoprop/detail/graph_encoding/MPGraphEncoding.cpp
```diff
@@ -142,61 +152,104 @@ auto build_packed_cross_rank_storage(const std::vector<CrossRankPartnerData> &da
...
+                            LayerExchangeLayout &out,
+                            const char *what) -> void {
+    const size_t num_ranks = cross_rank.rank_count();
+    // assign(), not resize(): `out` is reused across layers and an empty slot must read zero.
+    out.counts.assign(num_ranks, 0);
+    out.displs.resize(num_ranks);
+
+    // Scatter over the slots that carry traffic: a dense probe would binary-search every possible
+    // partner, at O(P log occupied) per exchange, to fill an array that is mostly zeros.
+    for_each_occupied_slot(cross_rank, [&](size_t slot, const CrossRankSlotView &view) {
```
diagonal-hamiltonian: sonar

Check failure on line R221
Explicitly capture the required scope variables.
Check failure: Explicitly capture the required scope variables.

See more on https://sonarcloud.io/project/issues?id=Algorithmiq_monoprop&issues=AaAyuNRba8KXQfRGAUNw&open=AaAyuNRba8KXQfRGAUNw&pullRequest=270

cpp/monoprop/detail/graph_encoding/MPGraphEncoding.cpp
```diff
@@ -142,61 +152,104 @@ auto build_packed_cross_rank_storage(const std::vector<CrossRankPartnerData> &da
...
+    for (size_t r = 0; r < num_ranks; ++r) {
+        out.displs[r] = checked_exchange_int(total, what, "displacement");
+        total += static_cast<size_t>(out.counts[r]);
     }
-    return storage;
+    out.total_count = total;
 }
 
-} // namespace monoprop::detail
+auto build_layer_storage_unified(std::vector<CrossRankPartnerData> all_partners, size_t my_rank)
```
diagonal-hamiltonian: sonar Check warning on line R237
Pass expensive to copy object "all_partners" by reference to const.
Check warning: Pass expensive to copy object "all_partners" by reference to const.

See more on https://sonarcloud.io/project/issues?id=Algorithmiq_monoprop&issues=AaAyuNRba8KXQfRGAUNx&open=AaAyuNRba8KXQfRGAUNx&pullRequest=270

cpp/monoprop/detail/graph_encoding/MPGraphEncoding.cpp
```diff
@@ -142,61 +152,104 @@ auto build_packed_cross_rank_storage(const std::vector<CrossRankPartnerData> &da
...
+    auto storage = std::make_shared<LayerCore>();
 
-namespace monoprop {
+    storage->cross_rank = build_packed_cross_rank_storage(std::move(all_partners));
+    resolve_self_slot(storage->cross_rank, my_rank);
 
-auto LayerCore::derivative_exchange_layout() const -> const LayerExchangeLayout & {
-    if (!derivative_exchange_layout_cache_) {
-        derivative_exchange_layout_cache_ = detail::build_derivative_exchange_layout(evolution_exchange_layout);
+    {
```
diagonal-hamiltonian, line 242: remove comment

cpp/monoprop/detail/graph_encoding/MPGraphEncodingStorage.h
```diff
@@ -77,28 +73,124 @@ inline auto store_packed_phase(PackedPhaseStorage &storage, size_t idx, int phas
 
 auto build_packed_cross_rank_storage(const std::vector<CrossRankPartnerData> &data) -> PackedCrossRankStorage;
 
-inline auto cross_rank_sin_send_index(const PackedCrossRankStorage &storage, size_t rank, size_t idx) -> size_t {
-    const size_t offset = storage.ranges[rank].sin_send_offset + idx;
-    return static_cast<size_t>(storage.sin_send_indices[offset]);
+// Record where my_rank's own slot sits, so the gradient's self-slot reads are O(1). Once per layer.
+auto resolve_self_slot(PackedCrossRankStorage &storage, size_t my_rank) -> void;
+
+// One world slot's position in the flat B/D arrays, resolved once. The per-element accessors below
```
diagonal-hamiltonian: dont reference old code in this comment, make short

cpp/monoprop/detail/graph_encoding/MPGraphEncodingTypes.h
```diff
@@ -116,38 +107,75 @@ struct CrossRankPartnerData {
...
-    size_t sin_send_offset = 0; // into sin_send_indices; cumulative across ranks, so size_t (may exceed 2^32)
-    TermIndex sin_send_count =
-        0;                      // == sin_recv_count (both endpoints); TermIndex-wide so one rank/layer can exceed 2^32
-    size_t sin_recv_offset = 0; // into sin_recv_phases; cumulative across ranks, so size_t (see sin_send_offset)
-    TermIndex sin_recv_count = 0;
+// One world slot that carries traffic. The dense array this replaces held a record per POSSIBLE
+// partner, so it was sized by the flat world P and the graph carried P-squared records however little
+// was ever sent; what is stored here is bounded by the traffic instead. Neither the D range (it IS the
+// B range, permuted; enforced in build_packed_cross_rank_storage) nor the B/D offset (the running
+// prefix over stored entries) is a field: a size_t offset would pad the record from 12 B to 24.
```
diagonal-hamiltonian: remove

cpp/monoprop/detail/graph_encoding/MPGraphEncodingTypes.h
```diff
@@ -116,38 +107,75 @@ struct CrossRankPartnerData {
...
+    size_t self_offset = 0;
+
+    auto rank_count() const -> size_t { return world_size; }
+
+    // O(log occupied); a sweep over every partner should use for_each_occupied_slot instead.
+    auto find(size_t rank) const -> const CrossRankOccupiedSlot * {
+        const auto it =
+            std::lower_bound(occupied.begin(), occupied.end(), rank, [](const CrossRankOccupiedSlot &e, size_t r) {
+                return e.slot < r;
+            });
```
diagonal-hamiltonian: Replace with the version of "std::ranges::lower_bound" that takes a range.
Check warning: Replace with the version of "std::ranges::lower_bound" that takes a range.

See more on https://sonarcloud.io/project/issues?id=Algorithmiq_monoprop&issues=AaAyuNI6a8KXQfRGAUNs&open=AaAyuNI6a8KXQfRGAUNs&pullRequest=270

cpp/monoprop/detail/graph_encoding/MPGraphEncodingTypes.h
```diff
@@ -116,38 +107,75 @@ struct CrossRankPartnerData {
...
+        return (it == occupied.end() || it->slot != rank) ? nullptr : &*it;
```
diagonal-hamiltonian: Use "std::to_address" to convert iterator to pointer.
Check warning: Use "std::to_address" to convert iterator to pointer.

See more on https://sonarcloud.io/project/issues?id=Algorithmiq_monoprop&issues=AaAyuNI6a8KXQfRGAUNt&open=AaAyuNI6a8KXQfRGAUNt&pullRequest=270

cpp/monoprop/detail/graph_encoding/MPGraphEncodingTypes.h
```diff
@@ -116,38 +107,75 @@ struct CrossRankPartnerData {
...
 
 struct LayerCore final {
     PackedCrossRankStorage cross_rank;
-    LayerExchangeLayout evolution_exchange_layout;
-
-    auto derivative_exchange_layout() const -> const LayerExchangeLayout &;
 
-    // A copied core must not inherit the source's cache: it is eval-time state, not data.
-    auto reset_derivative_exchange_layout() -> void { derivative_exchange_layout_cache_.reset(); }
+    // NOTHING about the exchange is retained here: counts[r] is cross_rank.sin_send_size(r) except at
```
diagonal-hamiltonian: remove comment

cpp/monoprop/detail/monomial_propagator/MonomialPropagator.inl
```diff
@@ -819,8 +811,7 @@ auto MonomialPropagator<NumModes>::set_parameter_mapping(const VecZ &parameter_m
     auto relabel = [this](size_t layer, size_t new_param_index) {
         auto &target = graph_.get_layer(layer);
         auto new_core = std::make_shared<LayerCore>(target.core());
-        // Drop the inherited eval-time derivative layout: it must not depend on a prior gradient run.
-        new_core->reset_derivative_exchange_layout();
+        // A plain copy, with nothing to invalidate: the core retains no exchange state.
```
diagonal-hamiltonian: remove

cpp/monoprop/detail/mpi/Exchange.h
```diff
@@ -19,16 +19,29 @@
...
+//
+// 2. OPT-IN, and opted into at BUILD time: an audit of the invariant the exchange now rests on -- a
+//    layer's recv counts equal its send counts, so there is no transpose to compute or store (see
+//    Evolution.cpp's derive_layer_exchange). Compiled out unless the CMake option
+//    monoprop_CHECK_EXCHANGE_SYMMETRY is ON, because enabled it costs exactly the collective the
+//    change exists to remove. Not an environment variable, and deliberately not one: the audit IS a
+//    collective, so ranks that disagree about whether to run it hang the job rather than
+//    misreporting, and a per-rank variable is exactly how they come to disagree. A binary cannot.
+//    Worth having because the unguarded failure is a peer blocked in MPI_Alltoallv against a size
+//    nobody sends: a hang with no line number, which this turns into an exception naming the slot.
```
diagonal-hamiltonian: shorten this comment and remove the exchange symmetry flag. It is correct mathamathically

cpp/monoprop/detail/mpi/Exchange.h
diagonal-hamiltonian: 'auto request' can be declared as 'auto *request' [readability-qualified-auto]

cpp/monoprop/detail/pare/PareGraph.cpp
```diff
@@ -84,17 +75,13 @@ auto mark_cross_rank_endpoints_kept(const LayerTraversal &layer, size_t my_rank,
...
-                                                 0,
-                                                 layer.cross_rank_sin_recv_size(rank),
-                                                 [&mark](size_t, size_t tgt_idx, int) { mark(tgt_idx); });
-    }
-    for_each_remote_rank(layer, my_rank, [&layer, &mark](size_t rank) {
-        layer.for_each_cross_rank_sin_send_range(rank,
-                                                 0,
-                                                 layer.cross_rank_sin_send_size(rank),
-                                                 [&mark](size_t, size_t src_idx) { mark(src_idx); });
+    layer.for_each_occupied_slot([&](size_t rank, const detail::CrossRankSlotView &slot) {
```
diagonal-hamiltonian: Explicitly capture the required scope variables.
Check failure: Explicitly capture the required scope variables.

See more on https://sonarcloud.io/project/issues?id=Algorithmiq_monoprop&issues=AaAyuNRwa8KXQfRGAUNy&open=AaAyuNRwa8KXQfRGAUNy&pullRequest=270

cpp/monoprop/Evolution.cpp
```diff
@@ -60,8 +60,10 @@ auto combine_endpoint_contrib(const EndpointContrib &a, const EndpointContrib &b
 struct FlatExchangeBuffers {
     VecD send_buffer;
     VecD recv_buffer;
-    std::vector<int> recv_counts;
-    std::vector<int> recv_displs;
+    // Derived per layer for the exchange being posted, and reused, so it allocates once per thread per
```
diagonal-hamiltonian: shorten or remove

cpp/monoprop/Evolution.cpp
```diff
@@ -73,21 +75,27 @@ auto &acquire_flat_exchange_buffers() {
...
+auto layer_exchange_participates(const mpi::Comm &comm) -> bool {
+    return mpi::size(comm) != 1;
+}
+
+// Derive this layer's exchange layout into `buffers.layout` at `scale`; it describes BOTH sides. The
+// count matrix is symmetric -- rank m's slot for r holds the same endpoint pair as rank r's slot for m,
+// swapped -- so counts are equal, displacements are their prefix sums, and the recv layout IS the send
+// layout. Scaling by the same literal on every rank preserves that. The symmetry itself is auditable
+// under monoprop_CHECK_EXCHANGE_SYMMETRY; the call below is unconditional because it also enforces the
+// layout's width against the communicator, which is a precondition rather than an audit.
```
diagonal-hamiltonian: shorten massively

cpp/monoprop/Evolution.cpp
```diff
@@ -136,19 +142,19 @@ struct InFlightExchange {
     bool active = false;
 };
 
-// Callers run the participation guard first (see active_evolution_exchange_layout) so the layout is
-// never materialized at a single rank.
+// Callers run layer_exchange_participates first, so no layout is derived for a transfer that never posts.
```
diagonal-hamiltonian: remove

cpp/monoprop/Evolution.cpp
```diff
@@ -167,13 +173,14 @@ inline auto finish_layer_exchange(InFlightExchange &in_flight, Apply apply)
...
     return apply(ExchangePayload{.recv_buffer = in_flight.handle.buffers->recv_buffer,
-                                 .recv_displs = in_flight.handle.buffers->recv_displs,
+                                 .recv_displs = in_flight.handle.buffers->layout.displs,
                                  .my_rank = in_flight.my_rank});
 }
 
 // Per-thread pre-cos snapshot buffers, reused across layers to avoid a malloc/free per layer. Passed whole
 // to the pack and apply passes: each reads two of the four vectors, and re-splitting them at every hand-off
 // is what lets a send buffer be mistaken for a recv one.
+// Indexed by occupied position, not world slot, so nothing here grows with P.
```
diagonal-hamiltonian: remove

cpp/monoprop/Evolution.cpp
```diff
@@ -192,23 +199,19 @@ void pack_cross_rank_derivative_payload_impl(const DerivativeSnapshotScratch &sn
                                              int my_rank,
                                              const LayerExchangeLayout &layout,
                                              VecD &send_buffer) {
-    const size_t num_ranks = layer.cross_rank_rank_count();
-    for (size_t rank = 0; rank < num_ranks; ++rank) {
+    layer.for_each_occupied_slot([&](size_t pos, size_t rank, const detail::CrossRankSlotView &slot) {
```
diagonal-hamiltonian: Explicitly capture the required scope variables.
Check failure: Explicitly capture the required scope variables.

See more on https://sonarcloud.io/project/issues?id=Algorithmiq_monoprop&issues=AaAyuNYTa8KXQfRGAUN3&open=AaAyuNYTa8KXQfRGAUN3&pullRequest=270

cpp/monoprop/Evolution.cpp
```diff
@@ -219,37 +222,29 @@ auto apply_cross_rank_derivative_exchange_impl(VecD &state,
                                                const DerivativeSnapshotScratch &snap,
                                                const TrigValues &trig,
                                                const ExchangePayload &payload) -> EndpointContrib {
-    const size_t num_ranks = layer.cross_rank_rank_count();
     EndpointContrib local{};
-    for (size_t rank = 0; rank < num_ranks; ++rank) {
+    layer.for_each_occupied_slot([&](size_t pos, size_t rank, const detail::CrossRankSlotView &slot) {
```
diagonal-hamiltonian: Explicitly capture the required scope variables.
Check failure: Explicitly capture the required scope variables.

See more on https://sonarcloud.io/project/issues?id=Algorithmiq_monoprop&issues=AaAyuNYTa8KXQfRGAUN4&open=AaAyuNYTa8KXQfRGAUN4&pullRequest=270

cpp/monoprop/Evolution.cpp
```diff
@@ -286,54 +283,42 @@ void pack_cross_rank_evolution_payload_impl(VecD &op,
                                             int my_rank,
                                             const LayerExchangeLayout &layout,
                                             VecD &send_buffer) {
-    const size_t num_ranks = layer.cross_rank_rank_count();
-    for (size_t rank = 0; rank < num_ranks; ++rank) {
+    layer.for_each_occupied_slot([&](size_t rank, const detail::CrossRankSlotView &slot) {
```
diagonal-hamiltonian: Explicitly capture the required scope variables.
Check failure: Explicitly capture the required scope variables.

See more on https://sonarcloud.io/project/issues?id=Algorithmiq_monoprop&issues=AaAyuNYTa8KXQfRGAUN6&open=AaAyuNYTa8KXQfRGAUN6&pullRequest=270

cpp/monoprop/Evolution.cpp
```diff
@@ -286,54 +283,42 @@ void pack_cross_rank_evolution_payload_impl(VecD &op,
...
 }
 
 void apply_cross_rank_evolution_exchange_impl(VecD &op,
                                               const LayerTraversal &layer,
                                               double sin_val,
                                               const ExchangePayload &payload) {
     // op[i] is already cos-scaled, so only the sine term is added; rv[k] is the partner's pre-cos value.
-    const size_t num_ranks = layer.cross_rank_rank_count();
-    for (size_t rank = 0; rank < num_ranks; ++rank) {
+    layer.for_each_occupied_slot([&](size_t rank, const detail::CrossRankSlotView &slot) {
```
diagonal-hamiltonian: Explicitly capture the required scope variables.
Check failure: Explicitly capture the required scope variables.

See more on https://sonarcloud.io/project/issues?id=Algorithmiq_monoprop&issues=AaAyuNYTa8KXQfRGAUN7&open=AaAyuNYTa8KXQfRGAUN7&pullRequest=270

cpp/monoprop/Evolution.cpp
```diff
@@ -352,22 +337,21 @@ inline auto finish_cross_rank_evolution_exchange(VecD &op,
...
     }
     const auto pairs = self_d_count / 2;
     EndpointContrib local{};
     for (size_t k = 0; k < pairs; ++k) {
-        const size_t i1 = layer.cross_rank_sin_recv_index_at(my_rank, k);
-        const double phi1 = static_cast<double>(layer.cross_rank_sin_recv_phase_at(my_rank, k));
-        const size_t i2 = layer.cross_rank_sin_recv_index_at(my_rank, k + pairs);
-        const auto phi2 = static_cast<double>(layer.cross_rank_sin_recv_phase_at(my_rank, k + pairs));
+        const size_t i1 = detail::slot_sin_recv_index(slot, k);
+        const double phi1 = static_cast<double>(detail::slot_sin_recv_phase(slot, k));
```
diagonal-hamiltonian: Replace the redundant type with "auto".
Check warning: Replace the redundant type with "auto".

See more on https://sonarcloud.io/project/issues?id=Algorithmiq_monoprop&issues=AaAyuNYTa8KXQfRGAUN8&open=AaAyuNYTa8KXQfRGAUN8&pullRequest=270

cpp/monoprop/Evolution.cpp
```diff
@@ -392,46 +376,46 @@ void snapshot_remote_endpoints(const VecD &state,
...
-            layer.for_each_cross_rank_sin_send_range(r, 0, bc, [&bs, &bh, &state, &op](size_t k, size_t i) {
-                bs[k] = state[i];
-                bh[k] = op[i];
-            });
+    };
+    grow(snap.sin_send_state);
+    grow(snap.sin_send_op);
+    grow(snap.sin_recv_state);
+    grow(snap.sin_recv_op);
+    layer.for_each_occupied_slot([&](size_t pos, size_t r, const detail::CrossRankSlotView &slot) {
```
diagonal-hamiltonian: Explicitly capture the required scope variables.
Check failure: Explicitly capture the required scope variables.

See more on https://sonarcloud.io/project/issues?id=Algorithmiq_monoprop&issues=AaAyuNYTa8KXQfRGAUN9&open=AaAyuNYTa8KXQfRGAUN9&pullRequest=270

diagonal-hamiltonian: Check warning on line R392
This lambda has 26 lines, which is greater than the 20 lines authorized. Split it into several lambdas or functions, or make it a named function.
Check warning: This lambda has 26 lines, which is greater than the 20 lines authorized. Split it into several lambdas or functions, or make it a named function.

See more on https://sonarcloud.io/project/issues?id=Algorithmiq_monoprop&issues=AaAyuNYTa8KXQfRGAUN-&open=AaAyuNYTa8KXQfRGAUN-&pullRequest=270

cpp/tests/boost-test.cmake
```diff
@@ -5,6 +5,25 @@ set(
...
+  if(NOT _monoprop_have_multirank)
+    message(
+      FATAL_ERROR
+      "monoprop_MPI_TEST_PROCS is \"${monoprop_MPI_TEST_PROCS}\", which has no entry >= 2. "
+      "The multi-rank cases open with `if (world_size() < 2) return;`, so a world-1-only rank list "
+      "reports them Passed having executed zero assertions."
+    )
+  endif()
+endif()
+
```
diagonal-hamiltonian: do we need this? delete if not

cpp/tests/README.md
```diff
@@ -93,8 +96,9 @@ name and cannot address suite-nested cases, tests use flat
   (recompute equivalence + snapshot invariance), `exact_upper_atol_rescue.cpp`,
   `large_cosine_storage_tests.cpp`, `gate_boundaries.cpp`.
 - **Graph encoding / packing**: `graph_encoding_tests.cpp` (CosineWordBuilder
-  coalescer, checked_* overflow guards, packed-phase storage + int8 read,
-  build_layer_exchange_layout, and both arms of the D-from-B derivation).
+  coalescer, checked_* overflow guards, packed-phase storage + int8 read, the
```
diagonal-hamiltonian: remove changes to readme

docs/content/docs/features/parallelism.mdx
```diff
@@ -67,6 +67,27 @@ A pure-MPI rank keeps its whole share in one partition unless
...
+  its cost tracks the terms crossing partition boundaries rather than the number
+  of partitions that exist.
+- The all-to-all counts and displacements are **derived** when an exchange is
+  posted and dropped afterwards, instead of being kept on the layer. They are a
+  prefix sum over what the layer already holds, and the recv side equals the send
+  side, so neither is stored per layer nor resolved over the wire.
+
+A build with `monoprop_CHECK_EXCHANGE_SYMMETRY=ON` audits that derivation against
+a real all-to-all on every exchange; see [Building from source](/building).
+
```
diagonal-hamiltonian: remove

docs/content/docs/building.mdx
```diff
@@ -85,6 +85,28 @@ mpiexec -n 8 uv run python your_script.py
 See [Parallelism and distribution](/features/parallelism) for the communicator options and the
 operator-partitioning controls.
 
+## Build options
```
diagonal-hamiltonian: remove this

docs/content/docs/testing.mdx
```diff
@@ -85,12 +85,39 @@ A 64-bit `TermIndex` build is the only configuration that compiles the
 just test-wide   # rebuilds via uv sync with monoprop_WIDE_TERM_INDEX=ON, then runs CTest
 ```
 
+Run it whenever you change a stored layout. Continuous integration compiles the
+wide build in a single matrix cell and every other build — local, packaged, and
+on the clusters — is narrow, so a wide-only regression has exactly one place left
```
diagonal-hamiltonian: remove this whole change

docs/content/docs/testing.mdx
```diff
@@ -85,12 +85,39 @@ A 64-bit `TermIndex` build is the only configuration that compiles the
...
+```bash
+SKBUILD_CMAKE_ARGS="-Dmonoprop_ENABLE_MPI=ON;-Dmonoprop_CHECK_EXCHANGE_SYMMETRY=ON" \
+    uv sync --all-extras --reinstall-package monoprop --no-cache
+mpiexec -n 4 build/editable/Release/bin/monoprop_unit_tests.x
+```
+
+The option table, and why it is a build option rather than an environment
+variable, are in [Building from source](/building). Leave it off for timing runs —
+it reinstates the per-exchange collective the derivation removes.
+
```
diagonal-hamiltonian: remove

AGENTS.md
```diff
@@ -104,6 +104,26 @@ Key files:
   for the mutating/collecting paths, which run on the partitions' own pinned masters; `sum_partitions_`,
   `fold_partitions_`, `first_partition_` for reads off quiescent partitions) rather than hand-rolling a
   `run_on_all` loop — the declarations record which helper is legal where.
+- **The distributed graph is indexed by the FLAT world**: a `LayerCore`
```
diagonal-hamiltonian: remove changes to agents file

CMakeLists.txt
```diff
@@ -55,6 +55,14 @@ option(
   OFF
 )
 
+# Build-time and not an environment variable on purpose; see check_exchange_symmetry
```
diagonal-hamiltonian: remove as we are removing this flag

cpp/monoprop/detail/mpi/Exchange.h
diagonal-hamiltonian: Applied as `MPI_Request request = MPI_REQUEST_NULL;` rather than `auto *request`.

`auto *` compiles only where `MPI_Request` is a pointer typedef, which is true for Open MPI (what CI and the clusters use) but not for MPICH, where it is an `int` handle — there `auto *` fails to compile outright. The explicit type satisfies the check by leaving no `auto` at all, and matches what the rest of `detail/mpi` already does: `Exchange.h:83` and `MPICompat.h:128` both spell it `MPI_Request`.

cpp/tests/boost-test.cmake
diagonal-hamiltonian: Deleted.

Checked before removing it: no configuration in the repo can trip it. The cache default is `"2"`, and the other places that set `monoprop_MPI_TEST_PROCS` — `justfile:37`, `tools/vscode-pytest-wrapper.sh`, `.vscode/pytest.env` — set it as a shell variable for `mpiexec -n`, which never reaches CMake. There is no `$ENV{...}` read of it anywhere in the CMake. The only documented non-default, `1;2;4`, passes.

To be accurate rather than convenient: it is not dead code. It is declared `CACHE STRING`, so a hand-passed `-Dmonoprop_MPI_TEST_PROCS=1` does override the default and does fire. It guarded a value someone types, not one we ship, which is why it goes.

The hole it aimed at is real and stays open: the cases that open with `if (world_size() < 2) return;` report Passed having asserted nothing on a world-1-only list. Deleting this declines to catch that rather than fixing it. Its comment also said "Nine cases" where there are eleven.

cpp/monoprop/detail/graph_encoding/MPGraphEncoding.cpp
diagonal-hamiltonian: Fixed, and it turned out to be more than a lint.

The parameter was by value, but the `std::move` into `build_packed_cross_rank_storage` was a no-op: that function only ever takes `const std::vector<CrossRankPartnerData> &`, so the rvalue bound to a const reference and nothing was consumed. The by-value bought one vector move-construct and no ownership.

Now `const &`, with the inert `std::move` dropped at both call sites (`Engine.h:178` and the one in `large_cosine_storage_tests.cpp`). So this removes a move rather than adding a copy.

docs/content/docs/features/parallelism.mdx
diagonal-hamiltonian: Cut the two lines about the audit, which is what this comment sits on, but kept the `### Graph memory at large world sizes` section above it.

Flagging that as a deviation rather than doing it quietly: those 17 lines are the `O(P^2)` rule this PR exists to fix, and they say nothing about the build flag. If you meant the whole section, say so and it goes.

cpp/tests/README.md
diagonal-hamiltonian: Reverted the paragraph this PR rewrote, but kept two bullets. Also a deviation, so: the kept bullets are the only documentation of `ExchangeLayoutOracle.h` and `exchange_layout_precondition_tests.cpp`, both files this PR adds.

A full revert has a second problem: the base sentence names `build_layer_exchange_layout` in the graph-encoding coverage list, and this PR deletes that library function — it now exists only as the test oracle. So the clean revert is the version that ships a stale statement, unless the bullet naming the oracle stays to explain where the name went.

Say the word and I will take both out.

cpp/monoprop/detail/mpi/Exchange.h
diagonal-hamiltonian: Flag and audit both gone — `monoprop_CHECK_EXCHANGE_SYMMETRY`, the CMake option, the compile definition, the audit body, its test, and the docs for it. The comment is now two lines.

One thing worth keeping in view, because "remove the function" would have been a regression rather than a simplification: `check_exchange_symmetry` was doing two jobs. The width check above the `#ifndef` is not this branch inventing a diagnostic — on `main` it lives inside `resolve_recv` (`MPICompat.cpp:122`, same message), which this branch deletes, so commit 6d6da04 relocated it rather than added it. It is unconditional, does no collective, and stops `MPI_Alltoallv` reading a counts array shorter than the rank count.

So the audit is deleted and the width check survives, renamed to `check_exchange_layout_width` since "symmetry" no longer describes it. Its two test cases survive with it; only the one inside the `#ifdef` is gone.

cpp/monoprop/detail/graph_encoding/MPGraphEncoding.cpp
diagonal-hamiltonian: Both throws now use a dedicated `CrossRankSlotLayoutError`, declared in `MPGraphEncodingStorage.h` above `build_packed_cross_rank_storage`, following the shape of `CollectiveArgumentError` in `detail/mpi/CheckedCount.h`.

It derives from `std::logic_error`, so the existing `BOOST_CHECK_THROW(..., std::logic_error)` cases stay green without edits.

Worth noting these were the only generic throws in the library: the codebase throws its own type 36 times against six generic ones, and all six were in this file. The other four are `std::overflow_error`, which is already specific, so I left them.

cpp/monoprop/detail/graph/MPGraphViews.h
diagonal-hamiltonian: Reduced, and I took "no caps" as general rather than local: there were 33 ALL-CAPS emphasis words across 12 files in this diff, and they are all gone now. The only remaining all-caps in added comments is the acronym `MPI`.

On comment volume overall: the diff added 264 comment lines against 942 of code (21.8%). It is now 183 (16.3%). I stopped there rather than chasing a lower number — the bulk of what is left is `HybridComm.h`'s barrier and lifetime rules and the per-case test rationale, and cutting further would have removed reasoning rather than prose. Happy to keep going on specific blocks if you point at them.

cpp/monoprop/Evolution.cpp
diagonal-hamiltonian: Split. The per-slot body moved into two named functions in the same anonymous namespace, `clear_slot_snapshot` and `fill_slot_snapshot`, so the lambda is now a branch and two calls.

diagonal-hamiltonian:

## Correctness gate on `5edeb53`

Re-run from scratch on this head, on Deucalion (`dev-x86`), after merging `main` at `638ee6f`. A
re-gate was required rather than optional: #267 touches
`cpp/monoprop/detail/evolution/layer_build/Engine.h` and #268 touches `src/monoprop/bindings/binder.h`,
both of which this PR also changes.

**Tested binary:** `_core.abi3.so` md5 `a336940cbde224c12543abde96bc97b9`,
`libmonoprop.so` md5 `0036644a82f93d4f09f2b3957a9d7aec`. The md5 is the arm identity here —
`monoprop.__version__` reports HEAD for every editable install and the dist-info stamp is not rewritten
by a later cmake rebuild, so both version identifiers go stale together and neither identifies a build.

| Suite | Result |
|---|---:|
| `ctest -L unit` | **235 / 235** |
| `ctest -L serial` | **234 / 234** |
| `ctest -L serial` (MPI build) | **234 / 234** |
| `ctest -L mpi` / `mpi-2` | **1 / 1** |
| Python suite, 4 layouts | **593 passed × 24 runs** |

Python layouts: ranks=2/partitions=1 (world 2), ranks=2/partitions=16 (world 32),
ranks=4/partitions=8 (world 32), ranks=16/partitions=16 (world 256). Verdicts
`BOTH GATES PASSED` and `ALL SUITES PASSED`; zero `FAILED`, zero `BAD_COMMAND`, zero
`Errors while running CTest`.

The unit and serial counts moved 231/230 → 235/234 across the two `main` merges, which is #267 adding
one case and #268 adding four. They track what `main` added, which is the check that the merges brought
in test coverage rather than displacing it.

Slurm jobs 1846481 (build) → 1846482 (unit + serial) → 1846483 (MPI + Python), run strictly in
sequence. An earlier attempt ran the two suites concurrently against one worktree and produced 14
spurious `BAD_COMMAND` failures — `ctest-worktree.sh` defaults `SYNC=1`, so its `uv sync` rebuilt the
editable tree the other job was executing from. Those were not test failures and that run is not the
one reported here.

**What this gate does not cover.** It is correctness only. The timing and memory tables in the
description were measured at `e25bbfb` and have not been re-measured since; the benchmark harness
records time and memory and cannot detect a wrong answer, so it is not evidence either way. Nothing
here re-establishes bit-identity, which this PR does not claim.

cpp/monoprop/MPGraph.cpp
```diff
@@ -51,7 +51,16 @@ auto layer_storage_memory_usage(const LayerCore &storage) -> GraphMemoryBreakdow
     GraphMemoryBreakdown breakdown;
     breakdown.layer_storage_object_bytes = sizeof(LayerCore);
     breakdown.cross_rank_bytes = detail::cross_rank_storage_bytes(storage.cross_rank);
-    breakdown.exchange_layout_bytes = detail::layer_exchange_layout_storage_bytes(storage.evolution_exchange_layout);
+    // Nothing retained: derived into per-thread scratch. The field stays so the row does not vanish
+    // from a memory A/B.
+    breakdown.exchange_layout_bytes = 0;
+
+    // Diagnostics. The graph does not partition: its arrays are indexed by the flat world.
```
robertodr: I don't think we need to be so conservative and comment-y

```suggestion
```

cpp/monoprop/Evolution.cpp
```diff
@@ -73,21 +72,22 @@ auto &acquire_flat_exchange_buffers() {
...
-    }
-    // All ranks must participate even at local total_count 0, else MPI_Alltoallv deadlocks.
-    return &layer.evolution_exchange_layout();
+// A property of the communicator, not the layer: all ranks participate even at local total_count 0.
+auto layer_exchange_participates(const mpi::Comm &comm) -> bool {
+    return mpi::size(comm) != 1;
+}
+
+// Derives both sides at once: the count matrix is symmetric, so the recv layout is the send layout.
+auto derive_layer_exchange(const LayerTraversal &layer, const mpi::Comm &comm, int scale, FlatExchangeBuffers &buffers)
```
robertodr: why `FlatExchangeBuffers &buffers` (IO param) instead of returning a `FlatExchangeBuffers` object?

cpp/monoprop/Evolution.cpp
```diff
@@ -386,52 +361,64 @@ auto apply_self_slot_derivative_paired(VecD &state,
...
-            layer.for_each_cross_rank_sin_recv_range(r,
-                                                     0,
-                                                     dc,
-                                                     [&ds, &dh, &state, &op](size_t k, size_t i, int /*phi*/) {
-                                                         ds[k] = state[i];
-                                                         dh[k] = op[i];
-                                                     });
+    // Grow-only: shrinking would free the allocations this scratch exists to reuse.
+    const size_t occupied = layer.occupied_slot_count();
+    const auto grow = [occupied](std::vector<VecD> &v) {
```
robertodr: for readability, this could be an `inline` function outside of the body of this one, instead of a lambda.

robertodr: actually, this lambda is equivalent to:
```c++
v.resize(std::max(v.size(), occupied));
```

cpp/monoprop/detail/graph_encoding/MPGraphEncodingTypes.h
```diff
@@ -116,38 +107,65 @@ struct CrossRankPartnerData {
...
 
-struct CrossRankPartnerRange final {
-    size_t sin_send_offset = 0; // into sin_send_indices; cumulative across ranks, so size_t (may exceed 2^32)
-    TermIndex sin_send_count =
-        0;                      // == sin_recv_count (both endpoints); TermIndex-wide so one rank/layer can exceed 2^32
-    size_t sin_recv_offset = 0; // into sin_recv_phases; cumulative across ranks, so size_t (see sin_send_offset)
-    TermIndex sin_recv_count = 0;
+// One world slot that carries traffic, so the array is bounded by the traffic and not by the flat
+// world P. The D range is the B range permuted, and the B/D offset is a running prefix, so neither is
+// a field: a size_t offset would pad the record from 12 B to 24.
```
robertodr: please rephrase this comment (or remove it) Like it is, it doesn't help.

robertodr (dismissed):
Got some comments, but overall LGTM. This is a complex piece of code in need of complete rewrite though...

cpp/monoprop/MPGraph.cpp
diagonal-hamiltonian: Applied — the two comments and the `breakdown.exchange_layout_bytes = 0;` line your suggestion spanned.

Removing the assignment is a no-op: `exchange_layout_bytes` is `= 0` in `GraphMemoryBreakdown` and `breakdown` is default-initialised, so `total_bytes()` and `operator+=` see the same value either way.

cpp/monoprop/Evolution.cpp
diagonal-hamiltonian: Because `buffers` is `thread_local` scratch — `acquire_flat_exchange_buffers()` hands back the same object on every call, and the point is to keep its three allocations across layers and steps. Returning a `FlatExchangeBuffers` would reallocate `counts`/`displs` (sized by world) and both payload buffers on every layer of every verb.

That said, the parameter was wider than the function: it only ever writes `buffers.layout`. It now takes `LayerExchangeLayout &`.

While there, `resize_flat_exchange_buffers` is gone — it was a one-line wrapper around a single `resize`. The send-buffer sizing is now inline at the call site, the same shape `begin_flat_exchange` already uses for the recv buffer:

```c++
derive_layer_exchange(layer, comm, scale, buffers.layout);
buffers.send_buffer.resize(buffers.layout.total_count == 0 ? 1 : buffers.layout.total_count);
```

cpp/monoprop/Evolution.cpp
diagonal-hamiltonian: It is — `std::max` never shrinks, which is the whole content of the guard. Lambda gone, four plain calls in its place:

```c++
snap.sin_send_state.resize(std::max(snap.sin_send_state.size(), occupied));
snap.sin_send_op.resize(std::max(snap.sin_send_op.size(), occupied));
snap.sin_recv_state.resize(std::max(snap.sin_recv_state.size(), occupied));
snap.sin_recv_op.resize(std::max(snap.sin_recv_op.size(), occupied));
```

(`<algorithm>` added; the file did not include it.)

cpp/monoprop/detail/graph_encoding/MPGraphEncodingTypes.h
diagonal-hamiltonian: Cut to one line:

```c++
// One world slot carrying traffic. Offsets come from a running prefix, which keeps the record at 12 B.
```

The B/D-permutation detail it used to carry is already stated where it is load-bearing, on the naming legend above `CrossRankPartnerData`.

diagonal-hamiltonian:

## Gate on `3a87c66`

The run the body promises, on the current head — after the `main` merges and after the review fixes in `3a87c66`. Deucalion, `dev-x86`, Open MPI 5.0.8, `MONOPROP_MAX_NUM_MODES=1024`.

Tested binary: `_core.abi3.so` md5 **`5ad0b693127b552f5f704a6d2dd7c95b`**, built from this tree by the gate job itself and verified unchanged after both CTest labels ran.

| gate | result | job |
|---|---|---|
| `ctest -L unit` | 238 / 238 passed, 0 failed | 1850563 |
| `ctest -L serial` | 237 / 237 passed, 0 failed | 1850563 |
| `ctest -L serial` (2 nodes) | 237 / 237 passed, 0 failed | 1850573 |
| `ctest -L mpi` | 1 / 1 passed | 1850573 |
| `pytest tests --with-mpi`, ranks 2 x partitions 1 (world 2) | 593 passed | 1850573 |
| `pytest tests --with-mpi`, ranks 2 x partitions 16 (world 32) | 593 passed | 1850573 |
| `pytest tests --with-mpi`, ranks 4 x partitions 8 (world 32) | 593 passed | 1850573 |
| `pytest tests --with-mpi`, ranks 16 x partitions 16 (world 256) | 593 passed | 1850573 |

`ALL SUITES PASSED`, no failures at any layout.

The counts are higher than the ones in the body (230 / 229 / 593) because `main` merged in since: the C++ suite grew, the Python suite did not.

**Not re-measured.** The performance tables in the body stand on the `pr6v2` campaign at `446d60c`. `3a87c66` is comment removal plus three mechanical changes — a narrowed parameter, an inlined one-line wrapper, and a `std::max` in place of an `if` — so it does not touch what those tables measure, but it is also not the binary they were measured on.

cpp/monoprop/Evolution.cpp
robertodr: I'm glad I still can see these simplifications

robertodr (approved):

PR #271, opened 2026-08-21
fix: :bug: caching cosine values for gradient computation

### LINKED ISSUE #146: Numerical instability in gradient computation
### Bug description

Gradient values seems to blow-up with the circuit length.  While it is OK for them to have approximation error, such high values are unphysical 
The issues gets more serious the closer (in the script below) the parameters are to pi/2

### Steps to reproduce

```
from __future__ import annotations

import itertools

import numpy as np

from monoprop import Circuit, ExpGate, Pauli, PauliOperator, PauliPropagator


def finite_diff_gradient_from_value_and_grad(
    expval_fn, params: np.ndarray, eps: float = 1e-5
) -> np.ndarray:
    """Compute a finite-difference gradient using the objective's value output."""
    params = np.asarray(params, dtype=float)
    grad = np.zeros_like(params)

    for i in range(params.size):
        plus = params.copy()
        minus = params.copy()
        plus[i] += eps
        minus[i] -= eps

        value_plus = expval_fn(plus)
        value_minus = expval_fn(minus)
        grad[i] = (value_plus - value_minus) / (2 * eps)

    return grad


num_qubits = 10
print(f"######## Num qubits: {num_qubits} ########")
cutoff = 4
print("Cutoff:", cutoff)

observable = PauliOperator(
    {Pauli("ZZ", [i, j]): 0.5 for i, j in itertools.combinations(range(num_qubits), 2)},
    num_qubits=num_qubits,
)

for p in [2, 4, 6, 8, 10]:
    gates = []
    for k in range(p):
        gates += [
            ExpGate(PauliOperator({pauli: v}, num_qubits=num_qubits), index=k)
            for pauli, v in observable.terms.items()
        ] + [
            ExpGate(
                PauliOperator({Pauli("X", i): -1.0}, num_qubits=num_qubits),
                index=p + k,
            )
            for i in range(num_qubits)
        ]

    circuit = Circuit(gates=gates)
    propagator = PauliPropagator(observable, [], cutoff=cutoff)
    propagator.build_graph(circuit)

    params = [1.0] * 2 * p
    grad = propagator.gradient(params)

    grad_fd = finite_diff_gradient_from_value_and_grad(
        propagator.expectation_value, params
    )

    print(
        f"> {p} -> expval={propagator.expectation_value(params)}, max|analytic|={np.max(np.abs(grad))}, "
        f"max|fd|={np.abs(grad_fd).max()}"
    )
```
Obtained
```
######## Num qubits: 10 ########
Cutoff: 4
> 2 -> expval=2.3303049804304212, max|analytic|=6.261198674357776, max|fd|=6.261198670931555
> 4 -> expval=1.4092108937992016, max|analytic|=1.1440711810707906, max|fd|=1.1514755316022196
> 6 -> expval=1.165524812710813, max|analytic|=2255462442.110155, max|fd|=0.3925156424777753
> 8 -> expval=1.0773135241021183, max|analytic|=8.751457623744551e+19, max|fd|=0.1670660519925704
> 10 -> expval=1.0382744613742356, max|analytic|=1.0783257981744055e+32, max|fd|=0.08047828148338709
```

### Expected behavior

Values similar to grad_fd (at least in order of magnitude)

### Operating system

Linux (aarch64 / ARM64)

### monoprop version

1efc636a06a811dfe4077c2e66e51d8aeb24c524

### Python version

Python 3.11.15

### Additional context

_No response_

### LINKED ISSUE #190: Reverse-mode gradient silently returns NaN when cos(2*g*theta) approaches zero

## What

The reverse-mode gradient recovers pre-cosine coefficients by dividing by `cos(2·gen_coeff·θ)`. There
is no guard, so at angles where that cosine approaches zero the gradient silently becomes `NaN`.

`cpp/monoprop/Evolution.cpp:45-52`

```cpp
explicit TrigValues(double param, double gen_coeff = 1.0) {
    const double g = 2.0 * gen_coeff;
    cos_val = std::cos(g * param);
    sin_val = std::sin(g * param);
    sec_val = 1.0 / cos_val;      // <-- unguarded
    g_val = g;
    tan_val = sin_val * sec_val;
}
```

`sec_val` is then applied per coefficient in `cpp/monoprop/detail/evolution/CosineRecompute.h`
(`accumulate_cos_mask:233`, `accumulate_cos_lazy:202`):

```cpp
loc += state[i] * ham[i];
ham[i] *= sec_val;      // un-does this layer's cos scaling
state[i] *= cos_val;
```

and in the self-slot pass at `cpp/monoprop/Evolution.cpp:365-368`.

## Why this is a problem

`θ = π/(4·gen_coeff)` makes `cos(2·gen_coeff·θ) = cos(π/2)`. Because `π/2` is not exactly
representable, `std::cos` returns ~`6.1e-17` rather than `0.0`, so `sec_val` is ~`1.6e16`: the whole
operator vector is amplified by 16 orders of magnitude, and every subsequent reverse layer inherits
the damage. The result is a `NaN` or a meaningless gradient with no diagnostic.

This is not an exotic angle. `π/4` at unit generator coefficient is an ordinary VQE/Trotter parameter,
and the repository's own kicked-Ising notebook fixes the coupling at `J = π/4`
(`docs/notebooks/kicked_ising/kicked_ising.ipynb`). An optimizer that wanders into the neighbourhood
gets a corrupted gradient rather than an error.

**Scope:** gradient path only. The forward path (`evolve_step`, `evolve_operator`,
`contract_partially`, `expectation_value`) uses only `cos_val`/`sin_val` and never divides, so it is
unaffected.

## Suggested fix

Two related changes:

1. In `TrigValues`, test `std::abs(cos_val)` against a documented floor and throw a named exception
   identifying the layer index, the angle, and the generator coefficient, instead of letting `inf`
   propagate. An exact `cos_val == 0.0` test is not enough — as above, the pathological case returns a
   denormal-ish nonzero, not zero.
2. Apply the same floor in the fused build path,
   `cpp/monoprop/detail/evolution/layer_build/Engine.h:545-547`:
   ```cpp
   const bool fused_scale = use_fused && only_rotate_len_k == 0 && fused_scale_coeffs != nullptr
                            && param.has_value() && cos_build != 0.0;
   ```
   This already falls back to a two-pass path when the fused `1/cos` recovery is impossible, so
   widening the exact-zero test into a threshold reuses machinery that already exists and hardens the
   forward fused path against the same amplification.

`cpp/tests/mpi_pare.cpp:45` already asserts `std::isfinite` on an expectation value, so the project
already treats non-finite results as failures — this just makes the gradient path say so.

## Verification

Add a test that evaluates `expectation_value_and_gradient` at `θ = π/4` with a unit generator
coefficient and asserts either a finite gradient or the new exception. Today it yields `NaN`.

---

_Found by a code-reading review of the repository at `29a8050`. No build tree was available, so the
analysis is from source inspection and should be confirmed against a build._

Leander-Thiessen:
Fixes #190. Fixes #146

## The problem

The gradient could come back completely wrong, off by many orders of magnitude, while the energy stayed correct. Two ways in: angles landing near a bad spot, or simply a deep enough circuit — #146 hits it with every angle at a perfectly ordinary `|cos| ≈ 0.5`.

## The bug

`sec_val = 1.0 / cos_val`. Going forward, `scale_cos_lazy` / `scale_cos_mask` scale every anticommuting coefficient by `cos(φ)` (with `φ = 2·g·θ`), and a separate sine pass then adds the partner term for each pair that survives the cutoff. Where the cutoff pruned the partner, the coefficient only ever gets that bare `cos` factor. To walk the layer back, `accumulate_cos_lazy` / `accumulate_cos_mask` did `ham[i] *= sec_val`, which is the exact inverse on paper. The trouble is that `ham[i]` arrives carrying rounding error from the reverse steps already done — the endpoint inverse rotations subtract, so cancellation leaves absolute error — and dividing scales that error by `1/cos`. Each coefficient picks up one such factor per layer it is anticommuting at, so the error compounds with depth. It does not take a near-singular angle: at `|cos| ≈ 0.5` the factor is only 2 per layer, and #146's 550 layers turn that into 1e+31 against a true gradient of 0.08. The forward build already refuses this trick when it cannot invert it safely (`cos != 0`); the reverse pass had no equivalent guard.

## The fix

The reverse pass reads the forward pass's pre-layer coefficients instead of dividing them back out, so there is nothing left to amplify. The whole design question is how little has to be kept to do that.

**Only the cosine set is ever read back.** The forward and reverse sweeps walk one layer's cosine set in the same order off the same fold, so a flat buffer written during `scale_cos_*` replays during `accumulate_cos_*` with no index map at all. The self-slot endpoints, which need random access, get their own small buffer filled in `sin_recv` order beside the existing `self_b_snapshot`. Everything else in the operator is never consulted and is never copied — that is what takes the Schrödinger benchmark's gradient from +8519 MiB down to +969 MiB.

**Every layer is recorded, and it has to be.** A layer's record restores exactly the coefficients that layer's cosine sweep divided, and no others. Skip a layer and its coefficients stay divided unrecovered; a later layer's record only rescues the ones its own cosine set happens to cover. A coefficient anticommuting with few generators is therefore reset rarely while still being divided often, so its error compounds over the whole circuit rather than between consecutive records. An earlier version of this branch recorded selectively, keyed on the running product of `1/|cos|` between records — that looked sound, passed the entire existing suite, and #146 still failed at 1e+30. It is gone.

The division survives as the `Replay == false` instantiation, reached by the energy path's unrecorded sweep and by two direct unit tests, hoisted behind a template parameter so its inner loop keeps its plain body.

One knock-on: the cosine sweep covers *all* anticommuting coefficients, endpoints included, and `apply_self_slot_derivative_paired` used to recover its pre-cos values by multiplying `op[i]` back by `cos_val` — undoing the very division we removed. With a record in hand the pre-layer value is already known exactly, so it is written straight through instead of being inverse-rotated back onto. `snapshot_remote_endpoints` and the cross-rank path needed no change; they already read snapshots taken before the sweep. `ev` is untouched and still evolves in a single `evolve_operator` call.

## Does it work

Issue #146's script, 10 qubits, cutoff 4, `max|analytic − finite differences|`:

| repetitions | layers | main | this branch |
|---|---|---|---|
| 4 | 220 | 5.98e-03 | 4.45e-10 |
| 6 | 330 | 1.53e+08 | 6.61e-11 |
| 8 | 440 | 6.66e+20 | 7.33e-11 |
| 10 | 550 | 3.97e+31 | 6.06e-11 |

Near-singular angles (120 gates, 16 modes, cutoff 8, twenty layers at `cos ≈ 1.6e-9`) match finite differences to 4.5e-11, where main diverges to ~6.5e+90. Pared graphs match to 1.7e-09 relative. The expectation value is bit-identical to the energy functional in every case.

`tests/test_deep_circuit_gradient.py` pins the depth failure: 8 qubits, cutoff 4, 8 repetitions, 0.3 s to run, and it fails on `main` by 12 orders of magnitude. Worth knowing that the cutoff has to be low enough relative to the qubit count to actually prune, or there is no error for the division to amplify and the check passes vacuously — which is what the entire existing suite was doing.

Suites: 209 C++, 621 Python, MPI at 1 and 2 ranks (594 Python per rank, 217 C++ each), and `just test-wide`.

## Cost

`benches/bench_random.py`, default sizes, `main` @ `2953349` vs this branch, same machine, alternating runs.

| | main | this branch |
|---|---|---|
| `opmemdelta` gradient [heisenberg] | 0.0 MiB | 0.0 MiB |
| `opmemdelta` gradient [schrodinger] | 85.1 MiB | 1048.6 MiB |
| `opmempeak` gradient [heisenberg] | 146.0 MiB | 145.8 MiB |
| `opmempeak` gradient [schrodinger] | 1798.8 MiB | 2767.8 MiB |
| gradient time [heisenberg] | 0.948 ms | 0.846 ms |
| gradient time [schrodinger] | 400.87 ms | 376.94 ms |

The Heisenberg gradient is unchanged on both axes. The Schrödinger gradient carries +969 MiB of records, a 1.54× peak — the real price of the fix, and Bencher will flag it. Every timing lands below 1.0×, including the energy path this branch does not touch, so that is drift between builds and not a speedup; there is no runtime difference to measure on this box.

## Notes for reviewers

- **changes done by AI, check if it makes sense**
- The load-bearing claim is that the forward and reverse cosine sweeps visit one layer's indices in the same order, which is what lets the record be a flat buffer with no index map. Both callbacks close over the same per-layer `LayerCos` entry in `build_cos_callbacks`, so they fold identically. `recorded_cosine_sweep_replays_pre_layer_coefficients` asserts the round-trip bit-exactly, driven with a `sec_val` that is deliberately not `1/cos_val` so the division path could not reproduce it by accident.
- Getting the Schrödinger +969 MiB down further means not holding every layer's record at once, i.e. checkpointing full operators every k layers and replaying segments forward. That trades memory for a second forward pass and is a much larger change; not attempted here.
- This branch was **re-written**, not rebased. The original commit patched code that no longer exists, so the fix was re-implemented against current `main`. The diff looks nothing like the old one.
- The branch's original second commit (a standalone benchmark script) was dropped. It imported helpers that have since been deleted, and `benches/bench_random.py::test_random_gradient` already covers the same thing.
- `detail::LayerCosScale` gains a fourth argument and `detail::CosCallbacks` a `cos_counts` member, so out-of-tree code constructing either needs updating. `evolve_step` and `state_operator_derivative_local` take their record defaulted and stay source-compatible.


robertodr:
Can we see a benchmark report for memory use on `main` vs this branch?

diagonal-hamiltonian:
Time and memory benchmarks is essential for this pls

Leander-Thiessen:

Benchmark report, `main` (668c90b) vs this branch, `benches/bench_random.py`
at default sizes (128 modes, 10k observable terms, 100 gates, cutoff 6),
`--bench-rounds 5`, serial.

### Memory — peak RSS over the timed call

| benchmark | main | branch | change |
|---|---|---|---|
| energy [heisenberg] | 147.1 MiB | 148.2 MiB | +1.1 MiB |
| gradient [heisenberg] | 146.6 MiB | 201.5 MiB | **+54.9 MiB (1.4×)** |
| energy [schrodinger] | 1714.4 MiB | 1714.3 MiB | −0.1 MiB |
| gradient [schrodinger] | 1801.1 MiB | 10320.1 MiB | **+8519 MiB (5.7×)** |

### Runtime — mean

| benchmark | main | branch | change |
|---|---|---|---|
| energy [heisenberg] | 0.56 ms | 0.53 ms | 0.95× |
| gradient [heisenberg] | 1.12 ms | 2.63 ms | **2.36×** |
| energy [schrodinger] | 143.2 ms | 155.5 ms | 1.09× |
| gradient [schrodinger] | 415.3 ms | 3195.0 ms | **7.69×** |

### Reading it

The energy path is unaffected on both axes, as intended.

The gradient cost is exactly `layers × operator_terms × 8` bytes, confirmed
against the measurement: Heisenberg has 70,336 terms over 100 layers → 53.7 MiB
predicted, 53.9 MiB measured. Schrödinger has **11,167,248** terms → 8.3 GiB
predicted and measured. The runtime cost is the same copy traffic.

**This is too expensive to merge as-is**, and the Schrödinger number is a
practical blocker.

### The fix 

The cache currently stores the whole operator per layer. Only the coefficients
in each layer's cosine set are ever read back:

| | full operator (current) | cosine set only |
|---|---|---|
| heisenberg | 53.7 MiB | 2.9 MiB (**18× less**) |
| schrodinger | 8519.9 MiB | 962.4 MiB (**8.9× less**) |

That brings the Schrödinger gradient peak from 5.7× down to roughly 1.5×, and
should remove most of the runtime regression along with the copy traffic.

Holding the PR for that change rather than merging the current version.

adamglos92:
@diagonal-hamiltonian @robertodr note that this is just a quick AI-cooked solution we needed quickly. 
* possibly there is some inefficiency in copying. In particular, perhaps we could only store the part which actually would go bad, e.g. those from unpaired cos branches? (I might be wrong but I think AI text above suggests that as well)
* if there would be still a big memory consumption, perhaps we could have two variants of gradient computation separated by mode?

robertodr:
@Leander-Thiessen thanks for posting the benchmarks. As-is, I cannot merge this. One of us can however prompt an LLM to apply the fix it suggested itself.

Leander-Thiessen:
> @Leander-Thiessen thanks for posting the benchmarks. As-is, I cannot merge this. One of us can however prompt an LLM to apply the fix it suggested itself.

I have one running already to test whether the fix actually works

matteoacrossi:
May I also ask to not have line breaks in the PR descriptions? For some reason Claude started doing that recently. You can ask it to fix it for you.

diagonal-hamiltonian:
Cheers leander, are the benchmarks done on Leonardo

Leander-Thiessen:
> Cheers leander, are the benchmarks done on Leonardo

no, I don't have access to leonardo atm - these are done on my laptop

adamglos92:
@Leander-Thiessen can you check if the instability issue from #146 is resolved?

Leander-Thiessen:
Claude thinks there might be a bug in the last change, will double check

adamglos92:
@Leander-Thiessen why closing?

Leander-Thiessen:
I just want to make sure it works before anyone starts reviewing

adamglos92:
@Leander-Thiessen better then convert to draft

Leander-Thiessen:

## Updated benchmark report — memory and runtime

`benches/bench_random.py -k "random_gradient or random_energy"`, `--bench-rounds 5`, default sizes (128 modes, 10 000 observable terms, 100 gates, cutoff 6, seed 0). `main` @ `2953349` vs this branch, same machine, runs alternating. Operator sizes identical on both: 70 336 terms (Heisenberg), 11 167 248 (Schrödinger).

### Memory

| | main | this branch | delta |
|---|---|---|---|
| `opmemdelta` energy [heisenberg] | 1.0 MiB | 0.9 MiB | −0.1 |
| `opmemdelta` **gradient** [heisenberg] | 0.0 MiB | 0.0 MiB | +0.0 |
| `opmemdelta` energy [schrodinger] | 86.7 MiB | 86.1 MiB | −0.7 |
| `opmemdelta` **gradient** [schrodinger] | 85.1 MiB | 1048.6 MiB | **+963.5** |
| `opmempeak` energy [heisenberg] | 146.4 MiB | 146.2 MiB | −0.2 |
| `opmempeak` **gradient** [heisenberg] | 146.0 MiB | 145.8 MiB | −0.2 |
| `opmempeak` energy [schrodinger] | 1714.6 MiB | 1719.5 MiB | +4.9 |
| `opmempeak` **gradient** [schrodinger] | 1798.8 MiB | 2767.8 MiB | **+969.0** |

The Heisenberg gradient is unchanged. The Schrödinger gradient carries the records: +969 MiB, a 1.54× peak. That is the honest cost of the fix and Bencher will flag it. For scale, caching the whole operator at every layer — the first version of this branch — cost +8519 MiB on the same benchmark, so this is 8.8× better but not free.

### Runtime

Best-of-2 minimum, 1000 repetitions (Heisenberg) / 12 (Schrödinger), builds measured alternating:

| | main | this branch | ratio |
|---|---|---|---|
| energy [heisenberg] | 0.447 ms | 0.407 ms | 0.91× |
| **gradient** [heisenberg] | 0.948 ms | 0.846 ms | 0.89× |
| energy [schrodinger] | 140.62 ms | 130.17 ms | 0.93× |
| **gradient** [schrodinger] | 400.87 ms | 376.94 ms | 0.94× |

All four land below 1.0×, including the energy path this branch does not touch, so this is the machine's drift between builds rather than a speedup. There is no runtime difference to measure here. A bare-metal number from the `bench_main.yml` testbed would be more decisive than this box.

### Where the memory goes

The reverse pass needs each layer's pre-layer operator. Only the cosine set is ever read back, and the forward and reverse sweeps walk it in the same order off the same fold, so a flat per-layer buffer written during `scale_cos_*` replays during `accumulate_cos_*` with no index map. The self-slot endpoints, which need random access, get their own small buffer in `sin_recv` order. Everything else in the operator is no longer copied at all — that is the difference between +969 MiB and +8519 MiB.

Every layer is recorded, and it has to be: a layer's record restores exactly the coefficients that layer's cosine sweep divided, so skipping a layer leaves those divided unrecovered, and a later record only rescues the ones its own cosine set covers.

| per-layer record | Heisenberg | Schrödinger |
|---|---|---|
| cosine set + endpoints (this branch) | 2.9 MiB | 962 MiB |
| whole operator (first version) | 53.7 MiB | 8520 MiB |

### Accuracy

Issue #146's script, 10 qubits, cutoff 4, `max\|analytic − finite differences\|`:

| repetitions | layers | main | this branch |
|---|---|---|---|
| 4 | 220 | 5.98e-03 | 4.45e-10 |
| 6 | 330 | 1.53e+08 | 6.61e-11 |
| 8 | 440 | 6.66e+20 | 7.33e-11 |
| 10 | 550 | 3.97e+31 | 6.06e-11 |

Also checked: near-singular angles (120 gates, cutoff 8, twenty layers at `cos ≈ 1.6e-9`) match to 4.5e-11; pared graphs match to 1.7e-09 relative; the expectation value stays bit-identical to the energy functional throughout.

Suites: 209 C++, 621 Python, MPI at 1 and 2 ranks (594 Python per rank, 217 C++ each), and `just test-wide` (210 C++, 621 Python).

adamglos92:
To me benchmarks look quite convincing. I can't tell much about the code of course.
@robertodr @diagonal-hamiltonian let me recall; if we are unhappy with the current implementation, I suggest for gradient functional to have mode specifying between memory hungry but precise, and memory-cheap and oftenly good implementation

diagonal-hamiltonian:
I would feel more comfortable if the memory regressions in the Schrödinger picture were resolved and the benchmarks were done on a cluster.

adamglos92:
@diagonal-hamiltonian BTW why there is a difference between Schrodinger and Heisenberg?

diagonal-hamiltonian:
I didn't write this pr no idea

Leander-Thiessen:
same

Leander-Thiessen:

Good question — it isn't the picture as such, it's the size and the weight profile of whatever vector is being evolved. Both arms of the benchmark run 100 layers, so depth isn't the variable. I instrumented the record layout to get the exact numbers (per hash partition; the propagator is a facade over 8 on this box, and each holds its own record):

| | Heisenberg | Schrödinger |
|---|---|---|
| evolved vector | 70 336 terms | 11 167 248 terms |
| cutoff | 6 | 8 (`schrodinger_cutoff = cutoff + 2`) |
| cosine set per layer, per partition | 469 | 157 700 |
| …as a fraction of that partition's terms | 5.3 % | 11.3 % |
| **total record** | **3.1 MiB** | **963 MiB** |

Two factors multiply:

**The vector is 159× bigger.** The Schrödinger arm evolves the state, whose support is far denser than the 10 000-term observable the Heisenberg arm evolves. That alone accounts for most of the gap.

**A higher fraction of it anticommutes.** The record holds each layer's cosine set, i.e. the terms anticommuting with that layer's generator. The generators are random weight-4 Majorana monomials in both arms, but the Schrödinger vector is capped at weight 8 rather than 6, and heavier monomials overlap a weight-4 generator more often — so the anticommuting fraction roughly doubles, 5.3 % → 11.3 %.

159 × 2.1 ≈ 330, and 963 MiB / 3.1 MiB ≈ 310. The same arithmetic checks against the version of this branch that cached the whole operator: 100 layers × terms × 8 B is 53.7 MiB / 8520 MiB, and the record is exactly the cosine fraction of that — 5.3 % × 53.7 = 2.8 MiB, 11.3 % × 8520 = 963 MiB, matching the measured `opmemdelta` of +963.5 MiB.

One reporting artifact on top of that: the Heisenberg row shows `opmemdelta` +0.0 MiB rather than +3.1. The metric is a `HighWaterMark`, so it only moves when the peak moves, and 3.1 MiB fits inside allocator headroom the propagator is already holding against a 146 MiB peak. 963 MiB against a 1799 MiB peak has nowhere to hide. So the Heisenberg record isn't free, it's just too small to register.

The practical read: the cost tracks the evolved vector's size and cutoff, not the picture label. A Heisenberg run at cutoff 8 on an 11 M-term operator would pay the same 963 MiB.

**Can it be improved?**
Truncation makes the forward step non-invertible (a term whose sine partner was pruned was multiplied by cos and nothing else, so the information to undo it is gone), which means reverse mode has to either store the pre-layer values or recompute them, and there is no third option. The 963 MiB is 11.3 full operators' worth, since the cosine set is 11.3 % of the terms across 100 layers. Checkpointing every ~30 layers would bring it to ~573 MiB for about +35 % gradient time, and a sound whole-circuit ∏1/|cos| bound would put the benchmark at exact parity with main — but only because its angles are small (product 6.56; triple the angles and it is already 1.7e+08), so that buys a green metric rather than user relief. The one avenue that could genuinely shrink it for free is recording only the pruned-partner terms, since surviving pairs are real rotations invertible by transpose, but that needs the pruned fraction measured first and is a rework of the evolution kernels. My recommendation is to ship as-is: 54 % over baseline, 8.8× better than the first attempt, and correct everywhere.

adamglos92:
@diagonal-hamiltonian based on AI analysis, it is what it is.
I still think we should have two options to choose; in many applications the old formula was just fine

Leander-Thiessen:
at least with this approach some increase in memory due to caching seems necessary, but no idea how far it can be minimized

cpp/monoprop/detail/monomial_propagator/MonomialPropagator.inl
```diff
@@ -896,23 +896,46 @@ auto build_cos_callbacks(const detail::InvertedIndex<NumModes> &inverted_index,
...
+                }
+                return detail::accumulate_cos_mask(s, h, *e.filtered, record, v, sec);
+            }
+            if (record != nullptr) {
+                return detail::accumulate_cos_lazy<NumModes, true>(*sc, e.recipe, s, h, record, v, sec);
+            }
+            return detail::accumulate_cos_lazy<NumModes>(*sc, e.recipe, s, h, record, v, sec);
+        };
+    // Sized once here rather than per call: the gradient records every layer, so nothing would be saved by
+    // deferring, and a recompute layer's count costs a fold.
```
robertodr:
```suggestion
```

diagonal-hamiltonian:
So if your going the caching route, caching every layers anticommuting set is not a good way to go. Do you only cache the angles which lead to an unstable inversion?

Leander-Thiessen:
I told the AI to run some more of the benchmarks and summarize them:

## Schrödinger picture: extra data points for the default-mode decision

`main` @ `638ee6f` vs this branch @ `4709acc`, same machine (8 cores, 19 GB, 1 rank, 8 partitions), builds alternating. Memory from `benches/bench_random.py -k "schrodinger and (random_gradient or random_energy)"` at `--bench-rounds 1`, timings at `--bench-rounds 5`. All memory in MiB.

**Summary of what the numbers say:** the memory cost is `layers × state terms` and scales linearly with depth against a flat baseline. But on this benchmark family the fix buys **nothing** — main is already correct at every depth, angle scale and cutoff tested, agreeing to ≤1.1e-15. The problems that do need it fail by 40 to 150 orders of magnitude, and they are also the expensive ones.

### 1. Memory cost across the Schrödinger space

`x base` = the increase as a fraction of main's energy-benchmark peak for the same configuration.

| config | layers | state terms | energy peak (main) | grad peak main | grad peak branch | delta | x base | grad `opmemdelta` main | grad `opmemdelta` branch |
|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| depth 25 | 25 | 11,057,843 | 1710.3 | 1795.1 | 2034.5 | **+239.4** | 0.14 | 85.1 | 324.1 |
| depth 50 | 50 | 11,083,463 | 1712.4 | 1796.0 | 2280.0 | **+484.1** | 0.28 | 85.0 | 569.9 |
| **depth 100 (default)** | 100 | 11,167,248 | 1715.4 | 1801.4 | 2766.2 | **+964.8** | 0.56 | 86.0 | 1049.0 |
| depth 200 | 200 | 11,311,942 | 1735.9 | 1823.8 | 3767.3 | **+1943.4** | 1.12 | 88.4 | 2028.0 |
| depth 400 | 400 | 11,625,172 | 1752.5 | 1845.7 | 5775.5 | **+3929.8** | 2.24 | 93.6 | 4023.5 |
| depth 800 | 800 | 12,514,604 | 1924.6 | 2027.6 | 10094.6 | **+8067.1** | 4.19 | 103.7 | 8171.1 |
| cutoff 4 | 100 | 351,178 | 196.4 | 200.1 | 215.6 | **+15.5** | 0.08 | 4.1 | 20.5 |
| cutoff 5 | 100 | 11,019,178 | 1709.2 | 1794.4 | 2749.4 | **+955.0** | 0.56 | 88.3 | 1044.7 |
| 64 modes | 100 | 767,791 | 235.7 | 242.2 | 353.3 | **+111.0** | 0.47 | 6.4 | 118.1 |
| 160 modes | 100 | 27,153,259 | 4892.1 | 5103.1 | 7022.2 | **+1919.1** | 0.39 | 209.1 | 2130.5 |
| 1k obs terms | 100 | 11,168,141 | 1718.4 | 1797.5 | 2758.3 | **+960.8** | 0.56 | 84.7 | 1047.5 |
| 100k obs terms | 100 | 11,230,979 | 1755.3 | 1841.3 | 2788.2 | **+946.9** | 0.54 | 86.8 | 1033.1 |

Notes:
- **The state saturates immediately** — 11.06 M terms at 25 layers, 12.51 M at 800. So the record grows linearly with depth while the baseline stays flat: +14 % of the resting peak at 25 layers, +56 % at 100, +112 % at 200, +224 % at 400, +419 % at 800.
- **Observable size is irrelevant.** 1k and 100k terms both land at 0.55×.
- Cutoff and mode count matter only through the term count they produce.
- 200 modes exceeded this box (OOM), so the modes axis stops at 160.

### 2. Accuracy on the same family — main and this branch agree

Same random problem, both builds, gradients compared component by component. `angles` multiplies the benchmark's `0.1·randn` parameters. Expectation values are identical between builds in every row.

| config | layers | angles | ev | main \|grad\|max | branch \|grad\|max | rel dev |
|---|---:|---:|---:|---:|---:|---:|
| L25 | 25 | 1x | +1.355767e+00 | 9.6855e-01 | 9.6855e-01 | 4.5e-19 |
| L25 | 25 | 3x | +1.086025e+00 | 2.6777e+00 | 2.6777e+00 | 4.1e-17 |
| L25 | 25 | 10x | -6.091082e-01 | 2.9503e+00 | 2.9503e+00 | 0.0e+00 |
| L50 | 50 | 1x | +1.325392e+00 | 6.8204e-01 | 6.8204e-01 | 4.1e-19 |
| L50 | 50 | 3x | +8.709490e-01 | 1.5725e+00 | 1.5725e+00 | 1.5e-18 |
| L50 | 50 | 10x | +4.451461e-01 | 1.7243e+00 | 1.7243e+00 | 1.3e-16 |
| L100 | 100 | 1x | +1.315344e+00 | 5.0617e-01 | 5.0617e-01 | 3.3e-16 |
| L100 | 100 | 3x | +8.169894e-01 | 1.0088e+00 | 1.0088e+00 | 2.5e-16 |
| L100 | 100 | 10x | +5.088169e-03 | 4.6273e-01 | 4.6273e-01 | 4.6e-18 |
| L200 | 200 | 1x | +1.089890e+00 | 1.2630e+00 | 1.2630e+00 | 2.0e-17 |
| L200 | 200 | 3x | +9.674841e-02 | 7.4430e-01 | 7.4430e-01 | 4.6e-16 |
| L200 | 200 | 10x | -4.706244e-03 | 1.4429e-01 | 1.4429e-01 | 3.8e-16 |
| L400 | 400 | 1x | +9.776365e-01 | 2.0654e+00 | 2.0654e+00 | 5.4e-17 |
| L400 | 400 | 3x | +1.926210e-02 | 2.6129e-01 | 2.6129e-01 | 1.1e-15 |
| L400 | 400 | 10x | +1.778528e-03 | 4.7560e-02 | 4.7560e-02 | 1.5e-16 |

At 10× angles the whole-circuit amplification bound is already 8.1e+21 and nothing degrades, which means there is essentially no error present for the division to amplify.

### 3. Same, sweeping the cutoff — truncation pressure alone does not create the instability

| cutoff | layers | angles | state terms | ev | main \|grad\|max | branch \|grad\|max | rel dev |
|---:|---:|---:|---:|---:|---:|---:|---:|
| 2 | 100 | 1x | 8,263 | +1.3153e+00 | 5.0617e-01 | 5.0617e-01 | 1.1e-16 |
| 2 | 100 | 10x | 8,263 | +5.0882e-03 | 4.6273e-01 | 4.6273e-01 | 1.5e-17 |
| 2 | 200 | 1x | 8,267 | +1.0899e+00 | 1.2630e+00 | 1.2630e+00 | 8.8e-17 |
| 2 | 200 | 10x | 8,267 | -4.7062e-03 | 1.4429e-01 | 1.4429e-01 | 4.8e-17 |
| 2 | 400 | 1x | 8,273 | +9.7764e-01 | 2.0654e+00 | 2.0654e+00 | 4.3e-16 |
| 2 | 400 | 10x | 8,273 | +1.7785e-03 | 4.7560e-02 | 4.7560e-02 | 7.3e-17 |
| 3 | 100 | 1x | 349,639 | +1.3153e+00 | 5.0617e-01 | 5.0617e-01 | 1.1e-16 |
| 3 | 100 | 10x | 349,639 | +5.0882e-03 | 4.6273e-01 | 4.6273e-01 | 1.5e-17 |
| 3 | 200 | 1x | 349,643 | +1.0899e+00 | 1.2630e+00 | 1.2630e+00 | 8.8e-17 |
| 3 | 200 | 10x | 349,643 | -4.7062e-03 | 1.4429e-01 | 1.4429e-01 | 4.8e-17 |
| 3 | 400 | 1x | 349,649 | +9.7764e-01 | 2.0654e+00 | 2.0654e+00 | 4.3e-16 |
| 3 | 400 | 10x | 349,649 | +1.7785e-03 | 4.7560e-02 | 4.7560e-02 | 7.3e-17 |
| 4 | 100 | 1x | 351,178 | +1.3153e+00 | 5.0617e-01 | 5.0617e-01 | 1.1e-16 |
| 4 | 100 | 10x | 351,178 | +5.0882e-03 | 4.6273e-01 | 4.6273e-01 | 1.5e-17 |
| 4 | 200 | 1x | 352,490 | +1.0899e+00 | 1.2630e+00 | 1.2630e+00 | 8.8e-17 |
| 4 | 200 | 10x | 352,490 | -4.7062e-03 | 1.4429e-01 | 1.4429e-01 | 4.8e-17 |
| 4 | 400 | 1x | 354,911 | +9.7764e-01 | 2.0654e+00 | 2.0654e+00 | 4.3e-16 |
| 4 | 400 | 10x | 354,911 | +1.7785e-03 | 4.7560e-02 | 4.7560e-02 | 7.3e-17 |
| 6 | 100 | 1x | 11,167,248 | +1.3153e+00 | 5.0617e-01 | 5.0617e-01 | 3.3e-16 |
| 6 | 100 | 10x | 11,167,248 | +5.0882e-03 | 4.6273e-01 | 4.6273e-01 | 4.6e-18 |
| 6 | 200 | 1x | 11,311,942 | +1.0899e+00 | 1.2630e+00 | 1.2630e+00 | 2.0e-17 |
| 6 | 200 | 10x | 11,311,942 | -4.7062e-03 | 1.4429e-01 | 1.4429e-01 | 3.8e-16 |
| 6 | 400 | 1x | 11,625,172 | +9.7764e-01 | 2.0654e+00 | 2.0654e+00 | 5.4e-17 |
| 6 | 400 | 10x | 11,625,172 | +1.7785e-03 | 4.7560e-02 | 4.7560e-02 | 1.5e-16 |

Dropping the cutoff from 6 to 2 takes the state from 11,167,248 terms to 8,263 and leaves the expectation value and gradient unchanged to six digits. This problem discards almost nothing that carries weight — which is exactly why main survives it.

### 4. Schrödinger problems where the fix *is* needed

ZZ brickwork: every qubit pair, then an X on every qubit, repeated, with `schrodinger_cutoff` equal to the operator cutoff. Every angle is ordinary (`|cos| ≈ 0.5`, nothing near singular). Memory is process high-water mark around the energy-only call versus after the gradient, one configuration per process (these do not run through the bench harness).

| problem | layers | state terms | main \|grad\|max | branch \|grad\|max | resting | with record | x base |
|---|---:|---:|---:|---:|---:|---:|---:|
| 12 qubits, cutoff 4, 10 reps | 780 | 46,171 | **4.9074e+43** | 5.4117e-01 | 127 | 208 | 0.64 |
| 16 qubits, cutoff 5, 12 reps | 1632 | 1,220,725 | **1.0054e+126** | 1.4889e+00 | 850 | 5345 | 5.29 |
| 18 qubits, cutoff 5, 12 reps | 2052 | 2,344,780 | **4.1091e+154** | 1.4915e+00 | 1534 | 11421 | 6.44 |

The branch's values were verified against central finite differences on the same shape at smaller scale (`tests/test_deep_circuit_gradient.py`).

### 5. Timing — not resolvable on this machine

Best-of-5 minimum, ms. The **energy path is untouched by this branch**, so its column is the control.

| config | gradient main | gradient branch | ratio | energy main | energy branch | energy ratio |
|---|---:|---:|---:|---:|---:|---:|
| depth 25 | 106.92 | 115.46 | 1.08 | 37.70 | 37.36 | 0.99 |
| depth 50 | 200.57 | 226.20 | 1.13 | 69.27 | 78.20 | 1.13 |
| depth 100 | 521.27 | 452.35 | 0.87 | 139.88 | 141.95 | 1.01 |
| depth 200 | 809.65 | 919.87 | 1.14 | 285.64 | 278.50 | 0.98 |
| cutoff 4 | 61.21 | 3.34 | 0.05 | 2.05 | 1.38 | 0.67 |
| cutoff 5 | 564.46 | 448.45 | 0.79 | 191.41 | 135.51 | 0.71 |
| 64 modes | 12.72 | 15.57 | 1.22 | 4.19 | 5.40 | 1.29 |
| 1k obs terms | 576.25 | 428.71 | 0.74 | 203.28 | 127.78 | 0.63 |
| 100k obs terms | 549.95 | 430.52 | 0.78 | 205.74 | 126.32 | 0.61 |

The control column spans 0.61×–1.29×, so the gradient's 0.74×–1.22× carries no signal. The `cutoff 4` row (0.05×) is an obvious warm-cache artifact, reported as measured. A bare-metal run from the `bench_main.yml` testbed would be needed to say anything about runtime.

### 6. A cheap discriminator between the two regimes

Each coefficient's amplification is a product of `1/|cos|` over the layers it anticommutes at, and every factor is ≥ 1, so the product over *all* layers is a sound upper bound. It depends only on the parameters and generator coefficients, is known before the forward pass, and is identical on every rank and partition with no communication.

| problem | layers | ∏ 1/\|cos\| | main correct? |
|---|---:|---:|---|
| random bench, 1× angles | 100 | 6.6e+00 | yes |
| random bench, 3× angles | 100 | 1.7e+08 | yes |
| random bench, 10× angles | 100 | 8.1e+21 | yes |
| 12 qubits, cutoff 4 | 780 | ~1e+222 | **no** |
| 16 qubits, cutoff 5 | 1632 | ~1e+458 | **no** |
| 18 qubits, cutoff 5 | 2052 | ~1e+573 | **no** |

Conservative rather than tight — it clears the benchmark at its default angles and would trigger at 3×, where main is still fine. But it never misses a genuinely unstable circuit, and applied **all-or-nothing per call** (record every layer, or none) it carries no partial-reset hazard. Applied per-interval it does, which is the trap the earlier revision of this branch fell into.

### Reading this for the default

- **Stable as the default.** The failure mode is silent and enormous — 1e+126 where the answer is 1.49 — and nothing about the picture, the cutoff or the angle scale predicts it: a 780-layer brickwork breaks while an 800-layer random circuit is fine. Memory is predictable and can be planned around; a plausible-looking wrong gradient cannot.
- **The mode keyword still earns its place** as an opt-out for users who have checked their own problem, and as the honest answer for anyone already at 400+ layers in the Schrödinger picture who wants main's footprint back.
- **Worth pricing before building a user-facing switch:** the amplification bound above would give benign problems main's exact cost automatically, including the tracked benchmark, with no flag for anyone to get wrong.

Panadestein:
Closing as superseded by #288.

PR #283, opened 2026-08-25
docs(benchmarks): 📝 restructure the benchmarks page and refresh every figure

Panadestein:

## Summary

The [benchmarks page](https://docs.monoprop.algorithmiq.tech/benchmarks) had grown to 532 lines, roughly 70% of it reproduction instructions, and its numbers no longer matched anything in the repository. This PR re-measures every sweep on Leonardo, restructures the page around what it reports, and adds a guard so the specific way it went stale cannot recur silently.

Memory is now reported as peak RSS from the kernel's `VmHWM`, stated once with the reasons it is trustworthy (exact rather than sampled, window settled and reset per step, same instrument applied to every engine from outside). The three scattered hedges that used to qualify it are gone.

## Changes

- **Restructure** `benchmarks.mdx`: 532 -> ~340 lines, no duplicate headings. Was 3x `SPECS`, 2x `Results`, 2x `Set up the Python environment` with identical bodies, and heading levels that drifted. Internal benchmarks now lead; the Pauli sweeps are `Fixed lattice, increasing depth` and `Increasing lattice size` rather than two sections with the same name. Nine numbered how-to subsections collapse into one `Reproducing` block, with schemas and field glossaries moved to `packages/bench-third-party/README.md`.
- **Re-measure everything** on Leonardo with `monoprop 0.8.1.dev61+gcb9a033aa`, which picks up the memory work in `638ee6f` and `cb9a033`. All records now carry the host RSS metric.
- **Add `check_memory_metric()`** to `plot_scaling.py`, refusing to label an axis "host RSS" when the records say otherwise.
- **Stamp provenance** into `run_model.py`'s output; move `num_threads()` into `backends.py` so both runners share one implementation.
- **Handle the two-node run**: the GPU could not share the CPU node, so the ladder comes back as `scaling_cpu.jsonl` + `scaling_gpu.jsonl`, combined at plot time with the CPU file first so `monoprop` stays on the CPU node. Each engine's own curve is single-host, which `warn_mixed_hosts` confirms.
- **Publish the scaling memory figure as `working_set_MB`.** For CPU engines that is the peak RSS; it differs only for the GPU, whose host RSS is a flat 530 MB that describes nothing while its device footprint reaches 34.4 GB. The host-RSS version drew that as a flat line an eighth of monoprop's, which reads as the opposite of the truth.
- **Fix stale claims**: the fixed-lattice sweep is 12x12 and not 6x6; `run_model.py` merges rather than overwrites; the broken `#3-optional-install-julia-and-pauliPropagationjl` anchor and the `### 5.` heading levels are gone.
- Replace `scaling_results.jsonl` with the two new files; gitignore the regenerated figures and sidecars.

## Results

The re-run is a materially harder problem than the last one, 28 Trotter points rather than 21, so ~55M terms rather than ~5.8M.

At the fixed 12x12 lattice `monoprop` is now the fastest engine including the GPU:

| engine | final step | vs monoprop | peak RSS |
| --- | ---: | ---: | ---: |
| `monoprop` | 1.56 s | - | 4275 MB |
| `cuPauliProp (GPU)` | 12.23 s | 7.8x | 11652 MB (device) |
| `QuEra ppvm` | 417.29 s | 267x | 6956 MB |
| `PauliPropagation.jl` | 433.16 s | 277x | 11102 MB |
| `Qiskit pauli-prop` | 507.74 s | 325x | 34270 MB |

On the ladder `monoprop` holds 6.35 s -> 14.1 s and 5045 -> 7835 MB across a 9x larger lattice. Majorana: 57.1 s against 1583.7 s (27.7x) at 8.8 GB against 17.5 GB, half the memory while retaining 1.38x more terms.

Two honest caveats, both stated on the page:

- **The 300 s per-point budget is now binding for the single-core engines.** `QuEra ppvm` and `Qiskit pauli-prop` do not finish 6x6, and `PauliPropagation.jl` stops after 10x10. Their curves end at the budget, not at the engines' limits. A longer timeout would extend them; worth deciding before the next run.
- **The "ppvm is leaner than monoprop" caveat has been dropped.** It was true of the old smaller operator and is not true here, where ppvm uses 1.6x the RSS.

## Verification

- `npm run build` passes on the rebased base with the bumped docs dependencies; all 12 heading anchors are unique, all six image references resolve, no orphaned assets.
- All six PNGs regenerate **bit-for-bit** from the committed JSON, so they can be redrawn without a Leonardo allocation.
- `ruff check` and `ruff format` clean relative to `main`; all `prek` hooks pass.
- `check_memory_metric()` verified to fire on the old data and stay silent on the new.

Note: `cf84009 perf(graph,mpi)` landed after these measurements were taken. It touches the distributed graph path rather than the single-node Pauli and Majorana runs, so the numbers should be unaffected, but they are one perf commit behind `main`.

## Checklist

## AI/LLM disclosure

- [x] I used the following tool to help write this PR description: pi with claude-opus-5
- [x] I used the following tool to generate or modify code: pi with claude-opus-5


docs/content/docs/benchmarks.mdx
```diff
@@ -1,396 +1,230 @@
 ---
 title: Benchmarks
-description: External and internal benchmarks for `monoprop`.
+description: Internal regression benchmarks and comparisons against other propagation engines.
```
matteoacrossi:
```suggestion
description: Comparisons against other propagation engines and internal regression benchmarks.
```

docs/content/docs/benchmarks.mdx
```diff
@@ -1,396 +1,230 @@
...
+than its 552 MB host RSS.
 
-Number of terms, final overlap, runtime, and memory vs. circuit layers for the 1D Fermi-Hubbard
-model, comparing `monoprop` against `MajoranaPropagation.jl`. See the
-section below for details on the setup and how to reproduce these results.
+![Runtime and peak RSS per Trotter step for five Pauli propagation engines](/benchmarks/pauli_results.png)
 
-![Runtime for the 1D Hubbard model in MajoranaPropagation.jl vs monoprop: seconds vs circuit layers, linear axes](/benchmarks/majorana_runtime.png)
+Solid lines denote peak RSS. Faint lines denote implementation-specific memory estimates and are
+not comparable across implementations.
```
matteoacrossi: Here we can use the caption block. 
Also, I don't understand the in-figure caption for the dashed lines. Consider removing them entirely. In any case, the caption should be in text outside the figure.

docs/content/docs/benchmarks.mdx
```diff
@@ -1,396 +1,230 @@
...
+| `cuPauliProp (GPU)` | 12.23 s | 7.8× | 11652 MB *(device)* | 2.7× |
+| `QuEra ppvm` | 417.29 s | 267× | 6956 MB | 1.6× |
+| `PauliPropagation.jl` | 433.16 s | 277× | 11102 MB | 2.6× |
+| `Qiskit pauli-prop` | 507.74 s | 325× | 34270 MB | 8.0× |
 
-![Benchmark results for the 1D Hubbard model in MajoranaPropagation.jl vs monoprop: number of terms, final overlap, runtime, and memory vs layers](/benchmarks/majorana_results.png)
+The observables produced by the coefficient-threshold implementations differ by at most 3.8e-4.
+`Qiskit pauli-prop`, which truncates by term count instead, differs by 2.8e-3. The `cuPauliProp`
+operator resides in GPU memory; consequently, the table reports its peak device allocation rather
+than its 552 MB host RSS.
```
matteoacrossi: We could explain somewhere what RSS stands for?

matteoacrossi:
```suggestion
operator resides in GPU memory; consequently, the table reports its peak GPU allocation rather
than the host RSS (which we measured at 552 MB).
```

docs/content/docs/benchmarks.mdx
```diff
@@ -1,396 +1,230 @@
...
-band, since flat here means small, not zero. For `monoprop`'s own growth with depth, read the
-term-count and memory panels of the combined figure, or the numbers below.
+#### Increasing lattice size
 
+This experiment holds the circuit depth and truncation threshold fixed while increasing the lattice
+from 6×6 (36 qubits) to 18×18 (324 qubits). Each model size and implementation runs in a separate
+subprocess with a 300 s limit.
 
-#### SPECS
+![Total runtime versus lattice size, monoprop, PauliPropagation.jl and cuPauliProp](/benchmarks/pauli_scaling_runtime.png)
```
matteoacrossi, line 67: Bonus: we could add a dashed line with "Timeout limit" annotation

docs/content/docs/benchmarks.mdx
```diff
@@ -1,396 +1,230 @@
...
+At the selected coefficient threshold, the operator size saturates at approximately 55.0
+million terms above 100 qubits. Beyond this point, the experiment primarily measures the cost per
+retained term. Across the full range, `monoprop` runtime increases from 6.35 s to 14.1 s (2.2×),
+while the qubit count increases 9×.
 
-- CPU: 2× Intel Xeon Platinum 8480+ (112 cores total, Sapphire Rapids)
-- RAM: ~480 GB
+The 300 s limit excludes `QuEra ppvm` and `Qiskit pauli-prop` at all lattice sizes, and excludes
+`PauliPropagation.jl` above 10×10. The endpoints of these curves therefore indicate timeouts, not
+implementation limits.
```
matteoacrossi:
```suggestion
`PauliPropagation.jl` above 10×10. We remark that this is not an implementation limit for the packages but a consequence of the timeout limit.
```

docs/content/docs/benchmarks.mdx
```diff
@@ -1,396 +1,230 @@
...
 
-`monoprop` is 22-142× faster over the grid, and the gap widens with both depth and system size:
-21.9× at 20 sites / 10 layers, 141.6× at 60 sites / 18 layers (4.2 s versus 592 s). Its runtime
-is nearly flat in site count while the Julia engine's grows roughly linearly. Dividing out the
-cores each engine actually kept busy leaves 8.5-52×, so the majority of the gap is algorithmic
-rather than a thread-count artifact. Memory is comparable at low depth and diverges in
-`monoprop`'s favour as the operator grows (2834 MB versus 3390 MB at 60 sites / 18 layers).
+| | node | hardware | engines |
+| --- | --- | --- | --- |
+| CPU | `lrdn4635` (DCGP) | 2× Intel Xeon Platinum 8480+, 112 cores, ~480 GB RAM | `monoprop` (56 threads), `PauliPropagation.jl` (28), `QuEra ppvm`, `Qiskit pauli-prop` |
```
matteoacrossi: Why is there a difference in the number of threads between monoprop (56) and PauliPropagation.jl (28)? Should we put (1) next to ppvm and qiskit?

docs/content/docs/benchmarks.mdx
```diff
@@ -1,396 +1,230 @@
...
-### Pauli Propagation
+### How memory is measured
 
-The chosen problem is the Trotterized time evolution of a 2D transverse-field Ising model (TFIM), tracking the runtime,
-expectation value, and operator size at each Trotter step.
+CPU memory denotes **peak resident set size (RSS)**, obtained from the Linux kernel's `VmHWM`
+counter. This definition is independent of the implementation language. For `cuPauliProp`, whose
+operator resides on the GPU, working memory instead denotes peak device allocation.
 
+- **The measurement is exact rather than sampled.** The kernel updates `VmHWM` whenever RSS
```
matteoacrossi:
```suggestion
- **The measurement is exact.** The kernel updates `VmHWM` whenever RSS
```

docs/content/docs/benchmarks.mdx
```diff
@@ -1,396 +1,230 @@
...
-Majorana and Pauli propagation engines. The benchmark suite includes memory and time
-performance metrics, providing insights into the efficiency of different
-propagation engines.
+The figure compares wall-clock time per propagation step on a logarithmic scale. At the final
+Pauli step, the runtime ratios relative to `monoprop` are 267-325× for the other CPU
+implementations and 7.8× for the GPU implementation. Over the complete Majorana evolution, the
+runtime ratio for `MajoranaPropagation.jl` is 27.7×. `monoprop` also has the lowest measured
+working-memory value in both comparisons. All results were obtained on
+[Leonardo (CINECA)](https://docs.hpc.cineca.it/hpc/leonardo.html); the experimental conditions are
+specified under [Methods](#methods).
```
matteoacrossi, line 24: This could be the figure caption (see #284). There could be a more generic paragraph saying:

We provide standalone benchmark scripts comparing `monoprop` against other free Pauli and Majorana propagation engines. We choose two paradigmatic models: a 2D TFIM and a 1D Fermi-Hubbard model respectively. Monoprop consistently outperforms the other packages in terms of runtime bla bla. some summary of the contents of the benchmark.

matteoacrossi (changes requested):
It's looking pretty good already. I left some comments. I'd like @robertodr and @diagonal-hamiltonian to review too. 

Remember to update the README with the new figures too.

docs/content/docs/benchmarks.mdx
```diff
@@ -1,396 +1,230 @@
...
+Working memory denotes peak host RSS for CPU implementations and peak device allocation for
+`cuPauliProp`. Across the full range, `monoprop` increases from 5045 MB to 7835 MB (1.55×), while
+`cuPauliProp` increases from 7.6 GB to 34.4 GB. At 324 qubits, the latter is 4.5× the
+`monoprop` host footprint and occupies more than half of the A100's 64 GB device memory.
 
-1. [Set up the Python environment](#1-set-up-the-python-environment).
-2. [Install Julia and `MajoranaPropagation.jl`](#2-install-julia-and-majoranapropagationjl).
-3. [Run the benchmarks](#3-run-the-benchmarks) to produce the JSONL result files.
-4. [Plot the results](#4-plot-the-results) to produce `majorana_results.png` and `majorana_runtime.png`.
+![Bar chart of monoprop's speed-up over each engine, grouped by qubit count](/benchmarks/pauli_speedup.png)
```
matteoacrossi, line 92: Consider removing the caption from the image and add it to the text.

docs/content/docs/benchmarks.mdx
robertodr: Each package has some API to query the internal memory accounting. E.g. the number of bytes taken by an operator. This should track closely what the OS reports, which is the trustworthy metric, but it's really up to the programmer to keep such internals up to date.

The above said: I agree with removing the dashed lines

docs/content/docs/benchmarks.mdx
```diff
@@ -1,396 +1,230 @@
...
+  `indexmap` variant rather than the Rayon-parallel `config::dashmap` variant; Qiskit's
+  `_accelerate` extension has no parallel propagation path. Runtime ratios are therefore
+  wall-clock ratios, not per-core ratios.
+- **The Julia benchmarks use the fastest documented containers.** These are `VectorPauliSum`
+  with `Performance.propagate!` (development branch 0.8.0 or later) and `VectorMajoranaSum` with
+  AcceleratedKernels. Their dictionary-backed alternatives are serial. The vector Pauli path does
+  not merge duplicate strings, so its reported term count is a storage count rather than the
+  number of distinct operator terms.
+- **The CPU and GPU measurements use different node types.** Each scaling curve is obtained on a
+  single node, and the plotting scripts reject curves whose records contain multiple hosts. The
```
```suggestion
  single node. The
```

docs/content/docs/benchmarks.mdx
robertodr: Or, we could keep the dashed lines, but I would make them report number of terms in the operator. This is a metric that differs slightly from package to package, because the truncations are not exactly the same. It would need a mirror y-axis.

docs/content/docs/benchmarks.mdx
robertodr: we measure at each engine's best. 56 is the number of cores on a single socket:

* Beyond 28 PP.jl performance degrades. Would be interesting to report how Julia binds threads to the NUMA domains on the single socket, or if the 28 threads are evenly distributed across sockets and NUMA domains.
* I'm guessing (@diagonal-hamiltonian confirm or deny) that we're binding to a single socket. To use the whole socket efficiently we'd need to use MPI (1 process per socket)

robertodr (changes requested):
The docs mention [RSS](https://en.wikipedia.org/wiki/Resident_set_size) (resident set size) but aren't we measuring [PSS](https://en.wikipedia.org/wiki/Proportional_set_size) (proportional set size)?

docs/content/docs/benchmarks.mdx
Panadestein: I removed it since I agree it wasn't adding much.

docs/content/docs/benchmarks.mdx
Panadestein: Thanks! I have removed all rasterized captions and made them text instead.

Panadestein:
> The docs mention [RSS](https://en.wikipedia.org/wiki/Resident_set_size) (resident set size) but aren't we measuring [PSS](https://en.wikipedia.org/wiki/Proportional_set_size) (proportional set size)?

Good catch: the benchmarks measure RSS from the kernel’s VmHWM, not PSS; the PSS sampler exists in the tooling but is not currently used by the benchmark harness. I clarified this in the MPI section of the docs, also defined RSS.

robertodr:
Should the PSS sampler then be removed?

docs/content/docs/benchmarks.mdx
matteoacrossi: In that case we should make it explicit that we are doing things at each engine's best. Otherwise the first impression is we are running PauliPropagation.jl with half the resources than monoprop.

robertodr: there used to be a footnote in this regard.

Anyway, @Panadestein could you check how those 28 threads are spread across the two sockets? Even looking at `htop` on the code might be enough.

docs/content/docs/benchmarks.mdx
```diff
@@ -1,418 +1,259 @@
...
 
-- CPU: 2× Intel Xeon Platinum 8480+ (112 cores total, Sapphire Rapids)
-- RAM: ~480 GB
+This experiment uses a 12×12 lattice (144 qubits), 28 Trotter points, `dt = 0.05`, a coefficient
+threshold of `1e-6`, and no weight cutoff. The final operator contains approximately 33 million
+terms.
 
-Each engine runs at the thread count it is fastest at — `monoprop` at 56 threads,
-`MajoranaPropagation.jl` at 28 — since they do not peak at the same count.
+| engine | final step | vs `monoprop` | peak RSS | vs `monoprop` |
```
matteoacrossi:
```suggestion
| Engine | Final step | VS `monoprop` | Peak RSS | VS `monoprop` |
```

docs/content/docs/benchmarks.mdx
```diff
@@ -1,418 +1,259 @@
...
-| `dt`         | Trotter step size.                                                                    |
-| `step_min`   | First Trotter step to simulate (inclusive).                                           |
-| `step_max`   | Last Trotter step to simulate (inclusive).                                            |
-| `lower_atol` | Coefficient magnitude below which Pauli terms are truncated.                          |
-| `cutoff`     | Maximum Pauli weight to keep during propagation; `null` disables weight truncation.   |
-| `obs_qubits` | The qubit pair (row-major grid index) whose `ZZ` correlator is measured.              |
+| | node | hardware | engines |
+| --- | --- | --- | --- |
+| CPU | `lrdn4635` (DCGP) | 2× Intel Xeon Platinum 8480+, 112 cores, ~480 GB RAM | `monoprop` (56 threads), `PauliPropagation.jl` (28), `QuEra ppvm`, `Qiskit pauli-prop` |
+| GPU | `lrdn0219` (Booster) | NVIDIA A100-SXM-64GB, 8 host cores | `cuPauliProp` |
```
matteoacrossi:
```suggestion
| Partition | Hardware | Engines |
| --- | --- | --- |
`DCGP (CPU) | 2× Intel Xeon Platinum 8480+, 112 cores, ~480 GB RAM | `monoprop` (56 threads), `PauliPropagation.jl` (28), `QuEra ppvm`, `Qiskit pauli-prop` |
| Booster (GPU) | NVIDIA A100-SXM-64GB, 8 host cores | `cuPauliProp` |
```

I was thinking that the node id is excessive, just the partition is enough.

robertodr: agree

docs/content/docs/benchmarks.mdx
Panadestein: I did a dry run to understand the NUMA configuration and how the threads are spread. So we have in the DCGP nodes 2 sockets × 4 NUMA domains × 14 cores. Then `JULIA_NUM_THREADS=28` puts 14 threads in each socket.

Panadestein: And then ~3–4 threads per NUMA node across all 8 domains.

robertodr: thank you Ramon!

matteoacrossi (dismissed):
Thanks, I don't have more comments, except that I meant to the versions of all packages benchmarked (so also MajoranaPropagation.jl, ppvm etc.) maybe in a table. It's not blocking though.

docs/content/docs/benchmarks.mdx
```diff
@@ -1,418 +1,261 @@
...
+![Time per step for monoprop against other Pauli and Majorana propagation engines](/benchmarks/headline_runtime.png)
 
 <Caption>
-  Number of terms, final overlap, runtime, and memory vs. circuit layers for the 1D Fermi-Hubbard
-  model, comparing `monoprop` against `MajoranaPropagation.jl`. See the
-  section below for details on the setup and how to reproduce these results.
+  Wall-clock time per propagation step on a logarithmic scale. At the final Pauli step, the
+  runtime ratios relative to `monoprop` are 267-325× for the other CPU implementations and 7.8×
+  for the GPU implementation. Over the complete Majorana evolution, the runtime ratio for
+  `MajoranaPropagation.jl` is 27.7×.
```
matteoacrossi:
```suggestion
  `MajoranaPropagation.jl 0.4.1` is 27.7×.
```

robertodr:
nitpick: the "Increasing lattice size" plots use a different color for the packages

matteoacrossi (dismissed):

robertodr (approved):
Thank you!

PR #288, opened 2026-08-25
fix(gradient): a stable gradient at the same time and memory

diagonal-hamiltonian:
## What goes wrong today

`gradient()` can return a number with no correct digits. It does not warn, slow down, or return a NaN — it just returns noise. At 10 qubits, cutoff 4, θ = 1, that starts at 12 circuit layers.

The reverse pass walks the circuit backwards and undoes one layer at a time. Undoing a layer means undoing its cosine scaling, and today the code does that by dividing by the cosine. Two separate things go wrong, and they multiply together.

**1. Rotations lose digits.** Most coefficients are only *scaled*: `h → h·cos` forwards, `h·cos → h` back. Multiply then divide by the same number is harmless however small the cosine is — about 1 ulp per round trip, and it does not grow. But a pair whose sine partner survives the cutoff is *rotated* rather than scaled:

```math
\begin{pmatrix} h_1'\\ h_2' \end{pmatrix} = \begin{pmatrix}c & -\varphi s \\ \varphi s & c\end{pmatrix} \begin{pmatrix}h_1\\ h_2\end{pmatrix}, \qquad c=\cos 2g\theta,\ s=\sin 2g\theta
```

Undoing that is a subtraction, `h₁ = c·h₁' + φs·h₂'`. When `c` is small, `h₁'` is tiny while `h₂'` is order 1, so the recovered `h₁` comes back with an **absolute** error of roughly `ε·|h₂|` — relative size `ε·|h₂|/|h₁|`. That is ordinary catastrophic cancellation, and it happens at every rotation.

**2. The division then amplifies it.** Once an error exists it is multiplied by `sec = 1/cos` at every layer the reverse pass has yet to walk. After descending past layers `M > j`, an error injected at the top has been amplified by

```math
S(j)=\sum_{M>j}\log_2\lvert\sec_M\rvert \quad\text{bits.}
```

`S` grows without bound in circuit depth: at 550 layers with θ = 1 it is in the hundreds of bits, so the leading bit of the answer is noise. This is not a "be more careful" problem. Inverting a contraction is ill-conditioned, and no rearrangement of the arithmetic changes the condition number. Something has to be remembered.

Measured at 10 qubits, cutoff 4, θ = 1 — `main` at `e64757e`, this branch at `32520de`:

| reps | layers | terms | `main` err | `main` \|grad\| | branch err | branch \|grad\| |
|---:|---:|---:|---:|---:|---:|---:|
| 2 | 4 | 145,656 | 7.6e-09 | 6.26 | 7.5e-09 | 6.26 |
| 4 | 8 | 407,736 | 4.4e-03 | 1.15 | 4.5e-10 | 1.15 |
| 6 | 12 | 669,816 | **6.6e+08** | 6.6e+08 | 6.6e-11 | 0.39 |
| 8 | 16 | 931,896 | **5.7e+20** | 5.7e+20 | 7.3e-11 | 0.17 |
| 10 | 20 | 1,193,976 | **3.7e+31** | 3.7e+31 | 6.1e-11 | 0.08 |

`err` is `max |analytic − central difference|`; `|grad|` is `max |analytic|`. From 12 layers on, `main`'s error and its own gradient magnitude are the same number to every digit shown — the returned gradient is entirely noise. The branch's magnitude decays smoothly and its error stays at the finite-difference floor.

## The fix: don't divide, remember

The forward pass already walks the circuit layer by layer with the pre-layer coefficients in hand. Before stepping through a layer it copies the coefficients the reverse pass will need, and the reverse pass puts those back instead of reconstructing them.

**Layer `j` records what layer `j−1` rotates.** That is the whole trick. At reverse step `j`, once layer `j` is undone, the operator is handed to step `j−1` — which is about to do the cancellation-prone inverse rotation. Resetting exactly those coefficients to their recorded forward values cuts the chain at every point where new error is injected, so there is nothing left to amplify. Recording at the rotating layer itself would restore one layer too late.

**Almost no circuit pays for it.** `S(j)` depends only on the parameters, so it is computed before a single coefficient is touched. If `S(j) < 53`, no error injected above layer `j` can have reached the leading bit and the record cannot change any answer, so it is skipped. The shipped benchmark circuits (θ ≈ 0.1) accumulate 2.7 bits over 100 layers: they record nothing at all and run `main`'s code path verbatim.

**A vanishing cosine is a separate failure needing a separate record.** When $|\cos| < 2^{-20}$, the sum `A = Σ_{i∈C_j} s_i h_i` is formed as `(Σ s_i h_i')·sec` out of terms scaled very differently from each other — rotation endpoints are order 1 while purely scaled coefficients are order `cos`. The small terms fall below the sum's own rounding, and multiplying by `sec` afterwards restores their magnitude but not their lost bits. Such a layer therefore records its **whole cosine set** — every coefficient anticommuting with its generator, and nothing else. Pre-dividing then makes the scaling uniform, and a single `×sec` undoes it exactly.

**One rearrangement.** The per-layer derivative is assembled in pre-layer units:

```math
\frac{\partial E}{\partial\theta}=-g\bigl(\sin\cdot(A-\text{ep}_{\cos})-\text{ep}_{\sin}\bigr)
```

`main` used `tan·(A − ep_cos)` on post-layer sums. Same value, but `tan = sin·sec`, so `main` formed the difference of two nearly equal sums and *then* multiplied the cancellation by `sec`. Here the subtraction happens before any amplification.

## Why the hot kernels are untouched

The obvious way to use a recorded value is to teach the reverse kernel to skip the division at those indices — a membership test in the hottest loop in the library.

Instead, `predivide_cos_record` writes `value × cos` at the recorded indices before the kernel runs. The unchanged kernel multiplies by `sec` only what is in the cosine set, so a recorded index comes back as `v·cos·sec` — the recorded value to within 2 ulp. An index *outside* the set is never visited and keeps the pre-divided value only until `restore_cos_record` overwrites it. Nothing has to be inferred, and duplicate indices are harmless.

So `accumulate_cos_mask` / `accumulate_cos_lazy` are **byte-identical to main**, `LayerCosAccumulate` keeps its signature, the `cos_acc` lambda is unchanged, the forward path of `Evolution.cpp` is unchanged, and `large_cosine_storage_tests.cpp` needed no edit.

## Cost

10 reps, arms interleaved and flipped per rep, ratios paired per rep. `energy` is the null control: it takes `evolve_operator` verbatim.

| test | `main` (s) | branch (s) | ratio | agree | p |
|---|---:|---:|---:|---:|---:|
| energy[heisenberg] *(null)* | 1.0484 | 1.0482 | 0.9994x | 6/10 | .75 |
| energy[schrodinger] *(null)* | 1.5058 | 1.5124 | 1.0027x | 10/10 | .0020 |
| gradient[heisenberg] | 2.2973 | 2.3175 | 1.0109x | 10/10 | .0020 |
| gradient[schrodinger] | 3.9862 | 3.9932 | 1.0021x | 6/10 | .75 |

Read each gradient against its own cell's null: schrodinger shows nothing above the null, heisenberg costs ~1.1%. Zero coefficients are recorded in either cell, so that residue is the layer-by-layer stepping the gradient path now does, not the record.

With the record *active* (46 of 100 layers recording) the gradient held 3.23 MiB above the energy path against a 3.3 GiB operator, and the wall was indistinguishable from the free cell. Rotation endpoints are a small subset, so a live record is close to free. Full numbers, and the bench knobs used to reach that regime, are in a comment below — they are not part of this PR.

## Tests

- `tests/test_deep_circuit_gradient.py` — deep circuits, a vanishing cosine, and a circuit that cannot amplify anything and so must record nothing.
- `cpp/tests/combined_recompute_equivalence.cpp` — pre-divide → unchanged kernel → restore must leave the operator bitwise pre-layer and the sum equal to an undivided reference, driven with a `sec_val` that is not `1/cos_val` so the check cannot pass vacuously.
- 600 Python tests across 4 rank/partition layouts (1×1, 1×16, 2×8, 8×16), so the record path is covered multi-rank.

The record indices are `TermIndex`, so their width tracks `monoprop_WIDE_TERM_INDEX`.


cpp/monoprop/detail/evolution/CosineRecompute.h
```diff
@@ -181,6 +181,37 @@ template <size_t NumModes>
     return apply_fold_mask(blk[wi - bb], wi, r.fold, row_parity);
 }
 
+// Take the recorded coefficients out of the sweep's way, keeping the state each carried in. A zeroed ham is
+// never divided, which is what subtracting the division back out cannot manage where 1/cos is enormous.
+inline auto detach_cos_record(const double *state, double *ham, CosRecordView record, double *entry) -> void {
```
robertodr: I guess `record` is probably a small POD, but better to avoid the copy entirely?
```suggestion
inline auto detach_cos_record(const double *state, double *ham, const CosRecordView &record, double *entry) -> void {
```

cpp/monoprop/detail/evolution/CosineRecompute.h
```diff
@@ -181,6 +181,37 @@ template <size_t NumModes>
...
+    for (size_t k = 0; k < record.count; ++k) {
+        const size_t idx = record.indices[k];
+        entry[k] = state[idx];
+        ham[idx] = 0.0;
+    }
+}
+
+// Put them back and return the sum they owe. A layer scales the state of exactly its cosine set, so a state
+// the sweep left untouched marks an index it never scaled: already pre-layer, and owing the sum nothing.
+inline auto restore_cos_record(const double *entry, const double *state, double *ham, CosRecordView record)
```
robertodr:
```suggestion
inline auto restore_cos_record(const double *entry, const double *state, double *ham, const CosRecordView &record)
```

cpp/monoprop/detail/evolution/CosineRecomputeCallbacks.h
```diff
@@ -18,15 +18,41 @@
 // without dragging in the build-pipeline templates.
 
 #include <cstddef>
+#include <cstdint>
 #include <functional>
+#include <vector>
 
 namespace monoprop::detail {
 
+/// Below this the reverse pass records everything rather than divide by the layer's cosine; see Evolution.cpp.
```
robertodr:
```suggestion
/// Below  2^-20  the reverse pass records everything rather than divide by the layer's cosine; see Evolution.cpp.
```

robertodr: I had no idea what hexadecimal floating-point literals were 🤯

cpp/monoprop/detail/evolution/CosineRecomputeCallbacks.h
```diff
@@ -18,15 +18,41 @@
...
+
 // Per-layer cosine callbacks; `layer` selects the cosine set to replay. Scale is the forward path
-// (coeff *= cos); Accumulate is the reverse path (state *= cos, ham *= sec) and returns Σ state·ham.
+// (coeff *= cos); Accumulate is the reverse path (state *= cos, ham *= sec, then `record` restored by
+// index) and returns Σ state·ham, pairing each post-layer state with the pre-layer ham.
 using LayerCosScale = std::function<void(size_t layer, double *coeff, double cos_val)>;
-using LayerCosAccumulate =
-    std::function<double(size_t layer, double *state, double *ham, double cos_val, double sec_val)>;
+using LayerCosAccumulate = std::function<
+    double(size_t layer, double *state, double *ham, CosRecordView record, double cos_val, double sec_val)>;
```
robertodr:
```suggestion
    double(size_t layer, double *state, double *ham, const CosRecordView &record, double cos_val, double sec_val)>;
```

cpp/monoprop/detail/evolution/CosineRecompute.h
```diff
@@ -242,7 +274,8 @@ inline auto accumulate_cos_mask(double *state, double *ham, const CosMask &cos,
             state[i] *= cos_val;
         });
     }
-    return loc;
+    // sec is constant over the set, so one multiply here beats one inside the accumulator chain.
```
diagonal-hamiltonian: delete

cpp/monoprop/detail/evolution/CosineRecompute.h
robertodr: ok, it's a span of a structure-of-array, so the copy costs next to nothing

matteoacrossi:
@diagonal-hamiltonian can you ask your bot to dumb down the PR description to a human comprehension level? And remove the line breaks while at it.

cpp/monoprop/Evolution.cpp
```diff
@@ -420,51 +446,78 @@ auto state_operator_derivative_local(VecD &state,
...
+                sink->values->push_back(op[i]);
+            }
+        }
+        else if (sink->enabled) {
+            for (const uint32_t idx : *sink->pending) {
+                sink->indices->push_back(idx);
+                sink->values->push_back(op[idx]);
+            }
+        }
+        sink->pending->clear();
```
robertodr: Could we make an initialized `CosRecordSink` structurally required here, or validate its vector pointers before use? The exported overload accepts a default-constructed sink whose `indices`, `values`, and `pending` members are all null, but the implementation dereferences them whenever `sink != nullptr`. That turns an otherwise valid public API call into a potential segfault.

benches/conftest.py
```diff
@@ -222,10 +230,16 @@ def _meta() -> dict[str, Any]:
...
     """Return the resolved random-problem hyperparameters (defaults included)."""
-    return {
+    resolved = {
         name.replace("-", "_"): config.getoption(f"--{name}")
         for name, _default, _help in _RANDOM_OPTIONS
     }
+    # num_modes and cutoff are per picture unless overridden, so a flat value would be null here.
+    resolved["picture_sizing"] = {
+        picture: {key: resolved[key] or value for key, value in sizing.items()}
+        for picture, sizing in _PICTURE_SIZING.items()
```
robertodr: Could we check explicitly for `None` here instead of using `or`? An explicit `--cutoff 0` or `--num-modes 0` is currently treated as unset and replaced with the picture-specific default, changing the existing CLI semantics. Checking `value is None` would preserve all user-provided integers.

benches/conftest.py
```diff
@@ -473,15 +487,16 @@ def picture(request: pytest.FixtureRequest) -> str:
...
     opt = request.config.getoption
+    sizing = _PICTURE_SIZING[picture]
     problem = make_random_problem(
         gen_length=opt("--gen-length"),
         obs_terms=opt("--obs-terms"),
         num_generators=opt("--num-generators"),
-        num_modes=opt("--num-modes"),
-        cutoff=opt("--cutoff"),
+        num_modes=opt("--num-modes") or sizing["num_modes"],
+        cutoff=opt("--cutoff") or sizing["cutoff"],
```
robertodr: Could we check explicitly for `None` here instead of using `or`? An explicit `--cutoff 0` or `--num-modes 0` is currently treated as unset and replaced with the picture-specific default, changing the existing CLI semantics. Checking `value is None` would preserve all user-provided integers.

diagonal-hamiltonian:
@robertodr — thanks, all nine addressed. Posting as one comment rather than inline replies because there is a pending (unsubmitted) review on this PR, and GitHub refuses `POST .../comments/{id}/replies` with *"user_id can only have one pending review per pull request"* while one is open.

Two of your comments were live bugs and are fixed. The rest are resolved by the rewrite that just landed (`d3fc3288`), which changes where the record is taken and applied.

### Fixed

**[#discussion_r3860740362](https://github.com/Algorithmiq/monoprop/pull/288#discussion_r3860740362) / [#discussion_r3860741806](https://github.com/Algorithmiq/monoprop/pull/288#discussion_r3860741806) — `or` swallows an explicit `0`.** Correct, and it silently changed existing CLI semantics. Added

```python
def _or_default(value: int | None, default: int) -> int:
    """Fall back only on an unset option, so an explicit 0 still reaches the model."""
    return default if value is None else value
```

used for both `--num-modes` and `--cutoff`, and the `picture_sizing` comprehension now tests `is None` as well. Verified: `_or_default(0, 128) == 0`, `_or_default(None, 128) == 128`, and `--num-modes 0 --cutoff 0` now reaches the model instead of being replaced by the picture default.

**[#discussion_r3860736758](https://github.com/Algorithmiq/monoprop/pull/288#discussion_r3860736758) — default-constructed sink with null members is a segfault.** Good catch, and it is now impossible to express. `CosRecordSink` is gone entirely; the record is taken in `prepare_evolved_operator`'s own layer loop, before `evolve_step`, where `op` still holds the pre-layer values. `state_operator_derivative_local` takes `const detail::CosRecordView &record = {}`, and both consumers loop `for (k = 0; k < record.count; ++k)`. A default-constructed view has `count == 0`, so the null `indices`/`values` are never dereferenced — there is no longer a "pointer non-null but members null" state to validate.

### Resolved by the rewrite

**[#discussion_r3860323154](https://github.com/Algorithmiq/monoprop/pull/288#discussion_r3860323154) — `CosRecordView` on `LayerCosAccumulate`.** Dropped; the signature is byte-for-byte main's again. The record is applied *around* the kernel call rather than inside it: pre-divide the recorded indices by writing `value * cos`, let the unchanged kernel's own `*= sec` land back on the recorded value, then restore. So `accumulate_cos_mask` and `accumulate_cos_lazy` are unchanged, and `MonomialPropagator.inl`'s `cos_acc` lambda and `large_cosine_storage_tests.cpp` needed no edit at all.

**[#discussion_r3860269929](https://github.com/Algorithmiq/monoprop/pull/288#discussion_r3860269929) / [#discussion_r3860271140](https://github.com/Algorithmiq/monoprop/pull/288#discussion_r3860271140) / [#discussion_r3860441256](https://github.com/Algorithmiq/monoprop/pull/288#discussion_r3860441256) — pass `CosRecordView` by const reference.** Taken. `detach_cos_record` no longer exists; the pair is now

```cpp
inline auto predivide_cos_record(double *ham, const CosRecordView &record, double cos_val) -> void;
inline auto restore_cos_record(double *ham, const CosRecordView &record) -> void;
```

Neither needs `state` or `entry` any more — the `thread_local` state snapshot is gone with them.

Worth flagging that the old `restore_cos_record` was more than a style question: it inferred cosine-set membership from `state[idx] != entry[k]`. At `|cos| = 1 - 2^-60` that product rounds back bitwise, so the test reports "unchanged" and every recorded term at such a layer is silently dropped from `A` — and the spread bound makes exactly that layer reachable, since a small-angle layer is recorded on behalf of the layers *above* it. `test_gradient_where_no_layer_can_amplify` used a uniform small angle and did not cover it. Pre-dividing removes the question: an index outside the set is simply never visited, and the restore overwrites it. There is a new C++ case pinning this.

**[#discussion_r3860303729](https://github.com/Algorithmiq/monoprop/pull/288#discussion_r3860303729) — `kVanishingCos` doc wording.** Adopted your framing, but not "records everything": such a layer now records only its **cosine set**, since indices outside it are untouched by the layer. Recording all of `op` was ~0.46 GiB per layer at 38M terms, and a kicked-Ising circuit is `θ = π/4` on *every* layer. It now reads:

> Below 2^-20 a layer's cosine scaling is not invertible in double, so the reverse pass reads its whole cosine set back from the record instead of dividing; see Evolution.cpp.

**[#discussion_r3860305821](https://github.com/Algorithmiq/monoprop/pull/288#discussion_r3860305821) — hex float literals.** 🙂 `0x1p-20` is exactly 2^-20 with no decimal-to-binary rounding, which matters for a constant that is compared against `std::abs(cos_val)`.

### Also worth knowing

The PR description has been rewritten with the actual mathematics — why the division is *not* the instability (a purely scaled coefficient round-trips with bounded relative error), why the rotation is (inverting it is a subtraction that injects absolute error `≈ ε|h₂|`), and why `S(j) = Σ_{M>j} log₂|sec_M|` is the bound that turns one into the other.

The branch was 8 commits behind and `CONFLICTING`, so it had never received CI. It is now rebased on current `main` and mergeable, and Lint is green.

diagonal-hamiltonian:
## Benchmark numbers (the bench changes are not in this PR)

The knobs used to reach the recording regimes are **not** part of this PR — it carries library code, tests and docs only. They will follow as their own change. Recording what they measured here so the cost of this fix is on the record.

### Setup

One node of Deucalion (AMD EPYC 7742, `znver2`), **1 rank, 1 thread**, `MALLOC_ARENA_MAX=1`, `MONOPROP_MAX_NUM_MODES=1024`, built `-O3 -DNDEBUG -march=znver2`. Arms interleaved and flipped per rep within one allocation; ratios are the median of the **per-rep paired** ratios, so a node-state swing hits both arms and cancels. Arm identity is the `_core.abi3.so` md5, not the version stamp.

Problem in all cells: `benches/bench_random.py`, `gen_length=4`, `obs_terms=10000`, `num_generators=100`, `seed=0`. The angle draw is `param_scale * 0.1 * randn`, and each generator carries its own `standard_normal` coefficient, so the layer cosine is `cos(2·g·θ)`.

| cell | modes | cutoff | angles | terms | operator + graph | layers recording |
|---|---:|---:|---|---:|---:|---:|
| **H** | 128 | 16 | shipped draw | 38,223,121 | 3.33 GiB | **0** of 100 |
| **S** (Schrödinger) | 176 | 6 | shipped draw | 39,740,141 | 6.25 GiB | **0** of 100 |
| **M** | 128 | 16 | draw ×8 | 38,223,121 | 3.33 GiB | **46** of 100 |

"Layers recording" is `layout_cos_records`' own suffix sum over the seed-0 parameters. The shipped draw accumulates **2.7 bits** of amplification across the whole circuit against a 53-bit threshold, which is why H and S record nothing and take `main`'s path verbatim. At ×8 the total is 78.5 bits and the smallest `|cos|` is 4.3e-4 — still above `kVanishingCos = 2^-20`, so M exercises the rotation-endpoint record only, never the cosine-set one.

### Time and peak RSS

| cell | op | `main` time | branch time | `main` peak RSS | branch peak RSS |
|---|---|---:|---:|---:|---:|
| H | energy *(null control)* | 1.0484 s | 1.0482 s | 4.02 GiB | 4.02 GiB |
| H | gradient | 2.2973 s | 2.3175 s | 4.30 GiB | 4.30 GiB |
| S | energy *(null control)* | 1.5058 s | 1.5124 s | 6.29 GiB | 6.29 GiB |
| S | gradient | 3.9862 s | 3.9932 s | 6.58 GiB | 6.58 GiB |
| M | energy | 1.0461 s | 1.0474 s | 4.06 GiB | 4.03 GiB |
| M | gradient | 2.3081 s | 2.3275 s | 4.35 GiB | 4.32 GiB |

Medians. H and S are 10 reps; **M is a single rep** — quoted for its memory shape, not as a resolved timing.

### Paired ratios, H and S

| test | ratio | agree | sign-test p |
|---|---:|---:|---:|
| energy[heisenberg] *(null)* | 0.9994x | 6/10 | .75 |
| energy[schrodinger] *(null)* | 1.0027x | 10/10 | .0020 |
| gradient[heisenberg] | 1.0109x | 10/10 | .0020 |
| gradient[schrodinger] | 1.0021x | 6/10 | .75 |

Both resolved results survive Holm over the four tests. `energy` is the null control: with nothing recorded, `prepare_evolved_operator` calls `evolve_operator` — `main`'s code, unchanged.

Read each gradient against its **own cell's** null. Schrödinger shows nothing above its null. Heisenberg costs ~1.1% against a flat null, and since zero coefficients are recorded there, that residue is the layer-by-layer stepping the gradient path now does, not the record itself.

### What the record actually costs when it fires

In M, the gradient's memory window held **3.23 MiB more than the energy window on the same arm** (454,094,848 vs 450,707,456 B). On `main` the same difference is −8 KiB, i.e. nothing, and in the free cell the branch's own difference is 48 KiB, i.e. nothing. That 3.23 MiB across 46 recording layers is ~70 KiB per layer, against a 3.3 GiB operator — rotation endpoints are a tiny subset of the cosine set, so a live record is close to free, and the wall was indistinguishable from the free cell.

### Not yet measured

The two heavier regimes have **no completed measurement**:

- `cos(2θ) = 1e-3` pinned on every layer → 9.97 bits per layer, threshold crossed by layer 6, **94 of 100** layers recording endpoints, nothing vanishing.
- `θ = π/4` → `cos(2θ) = 6.1e-17`, below `kVanishingCos`, so all 100 layers record their **full cosine set**. This is the memory case, budgeted at roughly +23 GiB over a 4.3 GiB peak, and the case `main` gets numerically wrong by an unbounded factor.

A first attempt at the second group was mis-sized (200 generators at cutoff 16 reached 206 GiB RSS and was abandoned). Both are cheap to run once the bench knobs land.

robertodr (commented):
can you make clang-tidy happy on `cpp/monoprop/Evolution.cpp`?

diagonal-hamiltonian:
@robertodr its done big man

diagonal-hamiltonian:
Both outstanding review threads are addressed. Still posting as one comment rather than inline replies — there is a pending (unsubmitted) review on this PR, and GitHub refuses `POST .../comments/{id}/replies` with *"user_id can only have one pending review per pull request"* while one is open.

### @copilot-pull-request-reviewer — 32-bit record indices (4 threads)

Right, and fixed in `a3994b0b`. `CosRecordView::indices`, `LayerCosIndices`, `EvalScratch::cos_indices`, the two enumerators in `CosineRecompute.h` and the C++ test all use `TermIndex` now, so the width follows `monoprop_WIDE_TERM_INDEX` instead of being pinned at 32 bits. It is `uint32_t` on the default build, so nothing changes there.

No `checked_term_index()` on this path: these indices are positions in this rank's own coefficient array, and an operator that large cannot be built on a narrow build in the first place — `OperatorIndex` already throws with *"rebuild with -Dmonoprop_WIDE_TERM_INDEX"*. A per-index bounds check in the record loop would pay for a condition that is unreachable by construction.

Verified by compiling `Evolution.cpp`, `MPFunctions.cpp` and `combined_recompute_equivalence.cpp` with `-Dmonoprop_WIDE_TERM_INDEX` under the project's own `-Wconversion -Werror` flags: clean, so nothing on the path narrows a 64-bit index.

### @robertodr — clang-tidy on `Evolution.cpp`

Done in `32520dee`. Worth saying what they were, because none of them are lines this PR wrote: the CI line filter covers a changed file end to end, so every pre-existing warning in a file this PR touches lands on the PR as an annotation. All nine:

| file:line | check |
|---|---|
| `Evolution.cpp:187, 212, 273, 289` | `modernize-use-integer-sign-comparison` |
| `Evolution.cpp:195, 196, 226` | `readability-math-missing-parentheses` |
| `Evolution.cpp:465` | `misc-use-internal-linkage` |
| `MPFunctions.cpp:263` | `modernize-use-ranges` |

Behaviour is unchanged in each: `std::cmp_equal(rank, my_rank)` is the same comparison without the `static_cast<int>`, the parentheses are the precedence the expression already had, `evolve_step_traversal_impl` had no users outside its TU so it moves into the anonymous namespace, and `std::ranges::iota` fills the same range.

### Previously

The nine earlier comments were answered [here](https://github.com/Algorithmiq/monoprop/pull/288#issuecomment-5424810899). The `benches/conftest.py` `or`-vs-`None` bug you caught is real and is fixed, but it now travels with the bench knobs in their own PR rather than this one.

---

Both commits verified on one Deucalion node against the merged head: wide-index compile clean under `-Wconversion -Werror`, CTest 242/242 serial and 1/1 mpi, 592 Python tests passed / 8 skipped.

cpp/monoprop/MPFunctions.cpp
```diff
@@ -260,7 +260,7 @@ auto EvalState::indices_above(double threshold) const -> VecZ {
     // than silently paring against a different keep-set.
     if (threshold < 0.0) {
         VecZ all(length_);
-        std::iota(all.begin(), all.end(), size_t{0});
+        std::ranges::iota(all, size_t{0});
```
robertodr:
```suggestion
        std::iota(all.begin(), all.end(), 0uz);
```
appleclang on macos-15 doesn't know about `std::ranges::iota`

matteoacrossi:
Is any of the sonarqube issues relevant/easy to fix?

robertodr (approved):

PR #290, opened 2026-08-26
fix(build): 🐛 export detail/** free functions reachable across the shared-library boundary

ludmilaasb:

## Summary

`libmonoprop.so` is built with hidden symbol visibility by default. A function only shows up in the library's public symbol table if it's explicitly marked `monoprop_EXPORT`.

A handful of free functions under `cpp/monoprop/detail/**` were missing that annotation, even though they're called from the public `MonomialPropagator` header that every client includes. In practice, this meant: link your own program against the installed `monoprop` package, instantiate `MonomialPropagator`, and it crashes at load time with `undefined symbol: ...`.

We never caught this ourselves because our own C++ tests link against the internal object library directly, not the installed shared library — so they never go through the same path a real external user does.

## Changes

- Export the missing functions: 8 in `CpuTopology.h`, 12 in `MPGraphEncodingStorage.h`, and 3 more in `MPICompat.h` that turned up once we tried to actually build against the fixed library.
- Add a permanent test (`cpp/tests/link_export_probe/`) that links against the real shared library with strict undefined-symbol checking, so a future missing export fails CI right away instead of only showing up as a crash for someone downstream.
- Make our existing `find_package(monoprop)` CI check actually instantiate and run the propagator, instead of just building an empty `main()`.

## Checklist

## AI/LLM disclosure

- [x] I used the following tool to help write this PR description: Claude Code (claude-sonnet-5)
- [x] I used the following tool to generate or modify code: Claude Code (claude-sonnet-5)

> [!IMPORTANT]

> [!WARNING]


.github/workflows/test.yml
```diff
@@ -127,16 +127,79 @@ jobs:
 
           add_executable(smoke main.cpp)
           target_link_libraries(smoke PRIVATE monoprop::monoprop)
+          # monoprop::monoprop does not propagate a C++ standard requirement, so any consumer of its
+          # C++23 headers (std::popcount, std::format, ...) has to opt in itself, same as here.
+          target_compile_features(smoke PRIVATE cxx_std_23)
```
robertodr: ah yes, good point. we should make `cxx_std_23` `PUBLIC` on our exported target

cpp/tests/link_export_probe/CMakeLists.txt
robertodr, line 1: is this test really needed? It's something you exercise in `test.yml` directly on the exported target as found by `find_package`, no?

robertodr:
This is very good, thanks for catching it! There's a bit of a "logistic" problem, which we will need to solve separately: the sources for these functions are in `detail` and yet they're part of the exposed ABI. This is inconsistent, but the fix is a larger change.

cpp/tests/link_export_probe/CMakeLists.txt
ludmilaasb: :robot: _AI text below_ :robot:

Good question — they're not fully redundant. The probe does full explicit instantiation of `MonomialPropagator<N>` (`template class monoprop::MonomialPropagator<2>;`), so the linker checks *every* member function's symbol references against the shared target, not just the handful the `test.yml` smoke test's `main()` happens to call. The smoke test only exercises `build_graph`/`graph_memory_usage`/`expectation_value_and_gradient`/partition `size()` — a missing export anywhere else in the ~70-method public API would sail through it. The probe also runs as a normal `ctest` target in-tree, so it catches regressions locally without needing a wheel build. Keeping both.

.github/workflows/test.yml
ludmilaasb: Shall we include this change in this PR as well?

robertodr: yes

ludmilaasb: :robot: _AI text below_ :robot:

Done in `1d72513` — added `target_compile_features(monoprop INTERFACE cxx_std_23)` to the exported target in `cpp/monoprop/CMakeLists.txt`, and dropped the `target_compile_features(smoke PRIVATE cxx_std_23)` workaround from the `test.yml` smoke test since it's no longer needed. Verified locally with a fresh `find_package(monoprop CONFIG)` consumer that declares no standard of its own — it now picks up `-std=gnu++23` automatically.

robertodr: `-std=gnu++23` is a mistake. We need to adhere strictly to the standard, no compiler extensions allowed. We set `CMAKE_CXX_EXTENSIONS` to `FALSE`, so it's a bit suspicious that gets ignored :thinking:

ludmilaasb: :robot: _AI text below_ :robot:

Good catch. `CXX_EXTENSIONS` is a target property that CMake doesn't propagate through `target_link_libraries`/`target_compile_features`. Only the standard level (`cxx_std_23`) propagates. `CMAKE_CXX_EXTENSIONS FALSE` in our `CXXFlags.cmake` only governs `libmonoprop.so`'s own build. Fixed in `46346ed` by adding `set(CMAKE_CXX_EXTENSIONS OFF)` to the smoke test's own `CMakeLists.txt`. Verified via `compile_commands.json`: `-std=c++23` now, was `-std=gnu++23` before.

ludmilaasb:
I opening for review just to trig the CI! Don't review yet!

cpp/monoprop/detail/mpi/MPICompat.h
robertodr: fly-by fix for `clang-tidy` lint.
```suggestion
    static auto get() -> MPI_Datatype {
```

robertodr (approved):
thank you!

PR #276, opened 2026-08-25
perf: :package: build fat binary for multiple microarchitectures

robertodr:
## Summary

Enabling autovectorization has been done with the `-march=native` flag (`monoprop_ENABLE_ARCH_FLAGS=ON`) in local builds, but to make _portable_ wheels, we have been compiling to the baseline of the `manylinux` container. For the various uses of `popcount` in this codebase that means effectively **not** using any CPU primitives (neither [`POPCNT`](https://en.wikipedia.org/wiki/SSE4#POPCNT_and_LZCNT) nor [`VPOPCNTDQ`](https://en.wikipedia.org/wiki/AVX-512#VPOPCNTDQ_and_BITALG)).

The approach here creates a so-called [fat binary](https://en.wikipedia.org/wiki/Fat_binary) where a single wheel ships _multiple_ DSO, each compiled for a specific microarchitecture, and then there's a dispatcher wrapping and routing the calls to the best match for the machine once runs on.

Fixes #13.

### x86-64

The fat binary includes the following microarchitectures:[^1]

* `x86-64-v1` which is the baseline
* `x86-64-v2` which adds `POPCNT`
* `x86-64-v3
* `x86-64-v4-vpopcntdq-vw256` AVX512 + `VPOPCNTDQ` with 256 bit vector width
* `x86-64-v4-vpopcntdq-vw512` AVX512 + `VPOPCNTDQ` with 512 bit vector width

### aarch64

WIP

## Changes

- [ ] Retire entirely the use of `-march=native`
- [ ] Fat-binary build is default on Linux x86-64
- [ ] Fat-binary build is default on Linux aarch64

## Checklist

- [X] Tests added or updated to cover the changes
- [X] Documentation updated (docstrings, `docs/`, `CONTRIBUTING.md`) if needed
- [ ] `CHANGELOG` / release notes updated if applicable

## AI/LLM disclosure

- [ ] I did not use LLM tooling, or used it only privately for ideation
- [ ] I used the following tool to help write this PR description:
- [X] I used the following tool to generate or modify code: ClaudeCode: claude-opus-5

> [!IMPORTANT]
[...]

PR #311, opened 2026-08-29
feat(build)!: ✨ ship x86-64 wheels as a fat binary over five ISA tiers

### LINKED ISSUE #13: Document usage of architecture dependent compiler flags

robertodr:
## Summary

Enabling autovectorization has been done with the `-march=native` flag (`monoprop_ENABLE_ARCH_FLAGS=ON`) in local builds, but to make _portable_ wheels, we have been compiling to the baseline of the `manylinux` container. For the various uses of `popcount` in this codebase that means effectively **not** using any CPU primitives (neither [`POPCNT`](https://en.wikipedia.org/wiki/SSE4#POPCNT_and_LZCNT) nor [`VPOPCNTDQ`](https://en.wikipedia.org/wiki/AVX-512#VPOPCNTDQ_and_BITALG)).

The approach here creates a so-called [fat binary](https://en.wikipedia.org/wiki/Fat_binary) where a single wheel ships _multiple_ DSO, each compiled for a specific microarchitecture, and then there's a dispatcher wrapping and routing the calls to the best match for the machine once runs on.

Fixes #13.

<details><summary>:robot: Details :robot:</summary>
<p>

## Summary

Published `x86-64` wheels carry the propagation engine five times over, compiled for four ISA levels with the top level shipped twice at two vector widths, and `monoprop` picks one when it is imported. Off by default in source builds, where `-march=native` beats every tier.

Stacked on #304, which turns `_monoprop_query_machine_flags` into the general `ARCH_FLAGS <flags...>` form — exactly the generalization the tiers need, since a tier is a whole flag list and not a single `-march` token, so this branch adopts that vocabulary rather than carrying its own.

This is the fat-binary half of #276, extracted onto `main`. #276 is based on `refactor-drop-nttp`, so it carries the whole NTTP stack (#302–#309) underneath it; the four whole-module fat-binary commits do not depend on any of that and are what this PR is. The narrow-seam commits that follow them (`e646bbc`, `68ad0a5`, and two thirds of `34218ac`) are **not** extractable: their design is "tier exactly the one translation unit holding `build_layer`'s instantiation tree", and on `main` there is no such TU — `MonomialPropagator` is header-resident and instantiated per mode width in the generated binder TUs, so the module boundary is the narrowest seam that exists here. That narrowing is a payoff of #308 and should land on top of it.

1. **Four ISA levels, from a compile-time study of what each buys this codebase.** `x86-64`, `-v2`, `-v3`, and `-v4` + `avx512vpopcntdq`. Plain `v4` is not shipped: ablating the eight AVX-512 extensions one at a time, `-mavx512vpopcntdq` accounts for the entire `v4 → v4x` gain and the other seven for exactly zero, which is what a codebase of `std::popcount` word loops and no intrinsics looks like. `v1` ships despite being slowest because it is the floor.
2. **The top level ships twice, at `-mprefer-vector-width=256` and `512`.** Identical `-march`, identical `__builtin_cpu_supports` requirements — the pair exists because left unset GCC takes the AVX-512 width from the `-mtune` tables, so it was being decided by `monoprop_FAT_MTUNE`, a core chosen for its schedule. No feature bit answers the question, so the discriminator is a core-name table read through `__builtin_cpu_is`. `znver4` is on it because it is measured: 1.1% to the narrow tier on the 127-qubit kicked-Ising model, three-sample ranges disjoint, *against* GCC's own znver4 tuning.
3. **Two predicates per tier, and the distinction is load-bearing.** `runnable` is features only and gates a `monoprop_VARIANT` pin; `preferred` adds the core table and is what the automatic selection reads. They differ for exactly one tier. Conflating them makes the 512-bit tier unpinnable on precisely the machines worth comparing it on.
4. **`-ffp-contract=off` is a contract, not a tuning knob.** Without it `-march=x86-64-v3` and up contract `a*b+c` into an FMA and the energy moves 1–2 ULP — the same wheel answering differently per host CPU. It is set project-wide so a source build, a wheel and every tier stay bit-comparable.
5. **Why not `glibc-hwcaps`, which would need no code at all.** Its directory names are the four psABI levels, and the top tier is `v4` *plus* `avx512vpopcntdq`. Installing it as `x86-64-v4` hands it to Skylake-X and Cascade Lake, which are `v4` with no vector popcount, and they take SIGILL. The predicate has to be ours.
6. **Why not `target_clones`.** GCC will not inline across an `arch` mismatch, so a `target`-attributed wrapper around the engine is a `jmp`, `flatten` notwithstanding; and `#pragma GCC target` does not capture templates defined outside its region. `flatten` + `target_clones` is unaffordable regardless — measured on the runtime-width engine, four flattened clones took one TU from 16.7 s to killed at 21 minutes and ~100 GB of compiler memory.

## Deviations from #276

- `_isa` uses nanobind split mode rather than `NB_STATIC`: `wheel.py-api = "cp311"` sets `SKBUILD_SABI_VERSION`, and nanobind 3 refuses any linked module below `cp312`. It still links no engine object library, so it inherits no tier's arch flags.
- `just diff-baseline-variants` is **not** included: it calls `tools/capture-baseline.py`, which arrives with #302, and its undefined `baseline_dir` made the justfile unparseable for every recipe. The byte-wise cross-tier gate is therefore currently only `test_variants.py`'s single probe. Worth restoring once #302 lands — `-vw256` and `-vw512` are otherwise indistinguishable except by disassembly.
- The `flatten` / `target_clones` numbers are attributed to the runtime-width engine they were taken on, and the note about narrowing the seam says what it needs.

</p>
</details> 

### x86-64

The fat binary includes the following microarchitectures:[^1]

* `x86-64-v1` which is the baseline
* `x86-64-v2` which adds `POPCNT`
* `x86-64-v3
* `x86-64-v4-vpopcntdq-vw256` AVX512 + `VPOPCNTDQ` with 256 bit vector width
* `x86-64-v4-vpopcntdq-vw512` AVX512 + `VPOPCNTDQ` with 512 bit vector width

### aarch64

WIP

## Changes

- [ ] Retire entirely the use of `-march=native`
- [ ] Fat-binary build is default on Linux x86-64
- [ ] Fat-binary build is default on Linux aarch64

<details><summary>:robot: Details :robot:</summary>
<p>

## Changes

- Add `cmake/compiler_flags/FatBinary.cmake`: the only place a tier or a narrow-vector core is declared. It generates the loader's predicate table into `FatVariants.h`, so a tier cannot be built without being selectable or selectable without being built.
- Compile the engine once per tier behind `monoprop_engine_sources(...)` and configure every tier through one `_monoprop_configure_engine_objs`, so the tiers cannot drift apart in anything but arch flags. A new `.cpp` registered with `target_sources(monoprop-objs ...)` reaches only the baseline tier.
- Apply the baseline ISA as a global floor in `CMAKE_CXX_FLAGS`, not per target: nanobind's glue is compiled with whatever `-march` the toolchain defaults to, which is not the psABI baseline (Ubuntu's GCC is `--with-arch-64=x86-64-v3`), and without the floor `_isa` itself would fault on the machines it exists to detect.
- Add `src/monoprop/bindings/isa.cpp`, a ~100 KB baseline-ISA probe answering `supported_variants()` / `runnable_variants()` / `known_variants()` off the generated table, and `src/monoprop/_bootstrap.py`, which binds one variant as `monoprop._core` before anything else imports it.
- Configure `Variants.h` per tier rather than once, so `monoprop.__variant__` and `__compiler_flags__` report what actually loaded. They previously reported a `-march=native` query whenever `monoprop_ENABLE_ARCH_FLAGS` was ON, regardless of what was compiled.
- Add `monoprop_ARCH_MARCH` as the variant *id* a single-ISA build reports as `__variant__`, so what is compiled and what is reported cannot disagree. The flags themselves are `ARCH_FLAG`, fed to #304's `ARCH_FLAGS` query.
- Enable the fat binary for `*-manylinux_x86_64` wheels only, via a `cibuildwheel` override: it is x86-64 only, and the other legs are aarch64.
- Add `tests/test_variants.py` (15 cases) and `just build-fat` / `just test-variants`.
- Add `docs/content/docs/fat-binary.mdx`; update `AGENTS.md`, `README.md` and `docs/content/docs/building.mdx`.

</p>
</details> 

## Checklist

## AI/LLM disclosure

> [!IMPORTANT]

> [!WARNING]


cmake/compiler_flags/Clang.CXX.cmake
```diff
@@ -16,9 +16,17 @@ if(CMAKE_CXX_COMPILER_ID MATCHES Clang)
...
+  # coefficient accumulation. Measured across ISA levels, every evolved term stays bit-identical and
+  # only the energy moves, by 1-2 ULP from -march=x86-64-v3 up. That is small and it is also exactly
+  # the wrong shape: with a fat binary the same wheel would answer differently depending on which CPU
+  # it landed on, and `just diff-baseline` could no longer be a byte-wise gate. The project has no
+  # -ffast-math and treats accumulation order as a contract, so contraction is off everywhere -- not
+  # only in the tiers -- to keep a source build, a wheel and every tier bit-comparable.
   set(
     monoprop_CXX_FLAGS
-    "-Wall -Wno-padded -Wno-unknown-pragmas -Woverloaded-virtual -Wwrite-strings -fcolor-diagnostics -Wno-c++98-compat -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer"
+    "-Wall -Wno-padded -Wno-unknown-pragmas -Woverloaded-virtual -Wwrite-strings -fcolor-diagnostics -Wno-c++98-compat -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -ffp-contract=off"
```
robertodr, line 29: I thinks we're fine with the compiler generating FMAs for us.

PR #314, opened 2026-08-31
build(python): :art: dynamically add mpi4py to build-system.requires

robertodr:

## Summary

`mpi4py` sat in `build-system.requires`, so every build installed it — including the non-MPI builds that are the default everywhere, and the cibuildwheel jobs that explicitly pass `-Dmonoprop_ENABLE_MPI=OFF`. It is only needed for MPI builds, where `src/monoprop/bindings/CMakeLists.txt` reads its headers to hand a communicator across the nanobind boundary. PEP 508 markers cannot express "only when MPI is on", so the dependency has to become dynamic.

It is now injected per-build through PEP 517's `get_requires_for_build_wheel`, via scikit-build-core's `build.requires` setting. **`monoprop_ENABLE_MPI=ON` in the environment is the single switch:** a `[[tool.scikit-build.overrides]]` block matches it and both adds `mpi4py` and sets the CMake option, so the two cannot drift apart. Enabling MPI therefore moves from a config-settings flag to an environment variable in the justfile, CI, README and docs.

The config-settings route still works but cannot chain: an override's `if.` clause matches environment variables only, so `--config-settings=cmake.define.monoprop_ENABLE_MPI=ON` must carry `--config-settings=build.requires=mpi4py>=4.1.0` alongside it. Getting that wrong used to surface as an opaque `Could NOT find mpi4py (missing: PY_mpi4py)`; the bindings CMakeLists now fails with a message naming both routes.

## Changes

- `pyproject.toml`: drop `mpi4py` from `build-system.requires`; add a `[[tool.scikit-build.overrides]]` block keyed on `if.env.monoprop_ENABLE_MPI` setting `build.requires` and `cmake.define.monoprop_ENABLE_MPI`. It must stay last in the `[tool.scikit-build]` section — an array-of-tables swallows any bare key that follows it.
- `src/monoprop/bindings/CMakeLists.txt`: `find_python_module(mpi4py QUIET)` plus an explicit `FATAL_ERROR` spelling out both ways to enable MPI.
- `justfile` (`test-mpi`, `bench-build-mpi`) and `.github/workflows/qa-analysis.yml` (`sonarqube-analysis`, `clang-tidy`): switch to the environment switch. MPI moves out of `SKBUILD_CMAKE_DEFINE`, which keeps its other defines.
- `docs/content/docs/building.mdx`: documents the environment route, the uv build-cache caveat, and a "Driving it with config-settings instead" subsection covering the `--config-settings-package` form and its repeated-pin cost.
- `docs/content/docs/testing.mdx`, `README.md`, `cpp/tests/README.md`, `AGENTS.md`: updated to the environment switch.

## Checklist

- [X] Documentation updated (docstrings, `docs/`, `CONTRIBUTING.md`) if needed

## AI/LLM disclosure

- [x] I used the following tool to help write this PR description: ClaudeCode: claude-opus-5
- [x] I used the following tool to generate or modify code: ClaudeCode: claude-opus-5

> [!IMPORTANT]

> [!WARNING]

### DIFF
--- a/.github/workflows/qa-analysis.yml
+++ b/.github/workflows/qa-analysis.yml
@@ -257,7 +257,8 @@ jobs:
 
       - name: Install package
         env:
-          SKBUILD_CMAKE_DEFINE: "monoprop_ENABLE_CXX_UNIT_TESTS=OFF;monoprop_ENABLE_MPI=ON"
+          SKBUILD_CMAKE_DEFINE: "monoprop_ENABLE_CXX_UNIT_TESTS=OFF"
+          monoprop_ENABLE_MPI: "ON"
         run: |
           uv sync --no-progress --all-extras -v
 
@@ -305,7 +306,8 @@ jobs:
 
       - name: Build (generates compile_commands.json)
         env:
-          SKBUILD_CMAKE_DEFINE: "CMAKE_CXX_COMPILER=clang++;monoprop_ENABLE_CXX_UNIT_TESTS=OFF;monoprop_ENABLE_MPI=ON"
+          SKBUILD_CMAKE_DEFINE: "CMAKE_CXX_COMPILER=clang++;monoprop_ENABLE_CXX_UNIT_TESTS=OFF"
+          monoprop_ENABLE_MPI: "ON"
         run: |
           uv sync --no-progress --all-extras -v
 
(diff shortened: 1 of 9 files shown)


Panadestein (approved):
Thank you!

PR #316, opened 2026-08-31
chore(benches): four operations, three measures

diagonal-hamiltonian:
Reduces `benches/` to what a performance review reads: **execution time and peak memory**, over four uniform operations. First of two; #311 adds the calibrated rung table on top.

## What goes

| removed | why |
| --- | --- |
| `test_random_pare`, `test_random_inplace` | neither is time-or-memory of a core operation |
| `test_model` | duplicated `test_model_propagate` under a name that collapsed both models to one label in the report |
| the `resting-memory` measure | a process-start footprint, not an operation's cost — dropped from `bmf.py`, the report, and the workflow threshold |

Leaves `build_graph`, `propagate`, `energy`, `gradient` uniform across both bench modules, and exactly three tracked measures: `latency`, `peak-memory`, `terms`.

**Deletions, never renames.** Benchmark names are Bencher's history key. Verified against `main`:

```
main: 22 collected   this branch: 16
removed: exactly the 6 parametrised names of the 3 deleted tests
added:   (empty)
```

So no surviving series is orphaned or silently re-keyed.

## What stays, deliberately

Every `opmem*` / `opbytes` recording in `conftest.py`. They render nowhere, but the out-of-tree A/B harness reads them — `benches/results/README.md` now says so, since "unrendered" read as "dead" to me on first pass.

## Also

- `bench.yml` uploads `bmf.json` and `benches/results/*.json` as artifacts, so a CI benchmark run is inspectable after the fact.
- `_display_op` fix: a non-picture parameter now names the model, so `[hubbard]` and `[pauli]` stop sharing a row label.

## Verification

- `pytest`: **622 passed, 8 skipped** — identical to `main`, since `benches/` is not in the default testpaths.
- `monoprop-bench-bmf` emits exactly `latency`, `peak-memory`, `terms`.
- `prek run --from-ref origin/main --to-ref HEAD` clean.

### DIFF
--- a/docs/content/docs/benchmarks.mdx
+++ b/docs/content/docs/benchmarks.mdx
@@ -324,8 +324,6 @@ For a fixed seed and problem size, `terms` is deterministic; any change therefor
 algorithmic change. It is tracked for that reason alone: it is the only check that a timing win is
 not an accuracy change. Timing alerts from shared runners are recorded but do not fail the build.
 
-`latency` is the mean because Bencher's t-test threshold consumes a mean and a spread.
-
 #### Testbeds
 
 Bencher identifies each series by `(branch, testbed, measure)`. A testbed therefore denotes one
--- a/docs/content/docs/benchmarks.mdx
+++ b/docs/content/docs/benchmarks.mdx
@@ -269,16 +269,20 @@ monoprop_NUM_THREADS=10 just bench serial-t10  # cap the partition count
 uv run --group bench monoprop-bench-report benches/results  # rebuild report, no re-run
 ```
 
-The suite contains two benchmark groups:
+Both groups measure the same four operations -- `build_graph`, `propagate`, `energy` and
+`gradient` -- so a number means the same thing whichever problem produced it:
 
-- `bench_random.py` measures `build_graph`, `pare`, `energy`, `gradient`, and `inplace` on a random
-  problem. Configurable with `--gen-length` (4), `--obs-terms` (10000), `--num-generators`
+- `bench_random.py` runs them on a random problem, in both the Heisenberg and the Schrödinger
+  picture. Configurable with `--gen-length` (4), `--obs-terms` (10000), `--num-generators`
   (100), `--num-modes` (128), `--cutoff` (6), `--seed` (0) and `--bench-rounds` (1); defaults in
   parentheses.
-- `bench_models.py` measures two fixed models marked `slow`: a 120-qubit Fermi-Hubbard 29-step Trotter
-  run (`test_model[hubbard]`) and a 127-qubit kicked-Ising run over 20 layers
-  (`test_model[pauli]`). Override any config field with `--<model>-<field>`, e.g.
-  `--pauli-num-layers 30`; `just bench --help` lists them all.
+- `bench_models.py` runs them on two fixed models marked `slow`, Heisenberg only: a 120-qubit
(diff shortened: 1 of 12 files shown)


docs/content/docs/benchmarks.mdx
```diff
@@ -307,18 +311,20 @@ just bench-bmf ci-linux         # the same results as Bencher Metric Format JSON
...
-| `resting-memory` | bytes | settled operator + graph footprint | +10% |
+| `peak-memory` | bytes | per-operation peak RSS (`VmHWM`), summed over ranks | +10% |
 | `terms` | count | terms in the evolved operator | any change |
 
 For a fixed seed and problem size, `terms` is deterministic; any change therefore indicates an
-algorithmic change. Timing alerts from shared runners are recorded but do not fail the build.
+algorithmic change. It is tracked for that reason alone: it is the only check that a timing win is
+not an accuracy change. Timing alerts from shared runners are recorded but do not fail the build.
+
+`latency` is the mean because Bencher's t-test threshold consumes a mean and a spread.
```
```suggestion
```

robertodr: BTW, latency of what?

packages/monoprop-bench-tools/src/monoprop_bench_tools/bmf.py
```diff
@@ -24,17 +24,20 @@
 uploaded together under ``--adapter json``; the stock ``python_pytest`` adapter
 would keep the timings and drop everything else.
 
-Four measures are emitted:
+Three measures are emitted:
 
 ``latency`` (nanoseconds)
     Per-operation mean, with the interval one standard deviation either side.
+    The mean is Bencher's contract: its t-test threshold consumes a mean and a
+    spread.
```
robertodr:
```suggestion
```

robertodr: It would be good to know latency of what

packages/monoprop-bench-tools/src/monoprop_bench_tools/bmf.py
```diff
@@ -24,17 +24,20 @@
...
 ``peak-memory`` (bytes)
-    Per-operation peak resident footprint, summed across ranks under MPI.
+    Per-operation peak resident footprint, summed across ranks under MPI. The
+    sum bounds the job; ``memhwm_max`` bounds a node, but CI is single-rank,
+    where the two coincide.
```
robertodr:
```suggestion
    Per-operation peak resident footprint, summed across ranks under MPI. The
    sum bounds the job; ``memhwm_max`` bounds a node. The two coincide for single-rank jobs.
```

packages/monoprop-bench-tools/tests/test_report.py
```diff
@@ -262,3 +254,16 @@ def test_build_report_omits_empty_schrodinger_section(tmp_path: Path) -> None:
     md = _collapse(report.build_report(tmp_path))
     assert "## Heisenberg" in md
     assert "## Schrödinger" not in md
+
+
+def test_two_models_do_not_share_a_row_label() -> None:
+    # Every fixed-model test is named test_model_*, so the model lives only in the
+    # parameter. Dropping it labels the hubbard and the pauli row identically and the
+    # report shows two rows nobody can tell apart.
```
robertodr:
```suggestion
```

robertodr (changes requested):
Thanks. Please look at my suggestions.

robertodr (approved):

PR #317, opened 2026-08-31
docs(benches): :building_construction: benchmarking ladder for sensitive PRs

diagonal-hamiltonian:

`benches/LADDER.md` — *the benchmarking ladder for sensitive PRs* — is five groups of benchmark configurations at the sizes the library is actually used at. Each row gives the exact pytest flags and the `-k` selector that produce it, so a group is a handful of ordinary `just bench` invocations. Nothing runs these, and no benchmark gates a PR.

```bash
just bench L1-hubbard-branch --hubbard-cutoff=10 --hubbard-lower-atol=4.2e-05 \
    -k "test_model_propagate and hubbard"
```

| group | shape | what it is for |
| --- | --- | --- |
| **L1** — one thread | 1 rank, 1 partition | hubbard and pauli `propagate` at ~10M terms; random `gradient` at ~20M, paired with and without `--pare-threshold` |
| **L2a** — one node, one rank, ~1B terms | 1 rank, all cores | all four operations over three models and both pictures |
| **L2b** — one node of ranks, same problems | `N`=1, `R × P` = cores | MPI with no network in it |
| **L3** — several nodes, same problems | L2b's `R` and `P`, your `N` | the same shape at `N` > 1 isolates the network |
| **L4** — strong and weak scaling | your `R`, `P`; `N` = 1…64 | hubbard `propagate` on one size sequence; both ladders built from it |

## What the review asked for

**Gradient rows at 20M terms, with and without `pare_threshold=1e-10`.** `--pare-threshold` did not exist; it is now a pytest option threaded into `expectation_value_functional` and `expectation_value_and_gradient_functional` for both the random and the fixed-model benchmarks. Unset reproduces today's behaviour exactly, so the Bencher series are unbroken. Measured at 1 rank / 1 partition / 1 thread (`_core.so` md5 `b201ec4`), `--obs-terms=295000` landing 19,902,244 terms:

| row | ~s (2 reps) | ~GiB |
| --- | ---: | ---: |
| random `gradient` | 11.05, 11.02 | 2.5 |
| random `gradient`, `--pare-threshold=1e-10` | 0.75, 0.75 | 2.8 |

Hubbard is not a candidate — 29 successive `build_graph` calls retain 29 layer-sets and `skip_if_graph_will_not_fit` skips it — so the gradient rows use the random problem, whose size is directly dialable. Proved the knob reaches the engine rather than being swallowed: `1e-10` leaves the energy bit-identical (the point of a retention cutoff), `1e10` moves it 0.8049 → 0.8307.

**L3's ratios removed.** The "against L2" column is gone; L3 reads like L2 and L1 — flags, `-k`, terms, `~s`, `~GiB/node` — with the shape those cells were measured at stated underneath (`N`=4, `R`=8, `P`=16, medians of two reps) rather than as a prescription.

**L3 and L4 generalised, and the shape made mandatory.** `N`, `R` and `P` are the caller's throughout, and `benches/conftest.py` now **refuses to start a session of more than one rank with `monoprop_PARTITIONS` unset**. The engine's own default is `ranks == 1 ? cores : 1`, so an unset knob measured one partition per rank at a plausible wall time — a trap the document used to merely warn about. Every rank raises after the collective `_nodes()`, so nothing is left in a collective. One-rank runs — `just bench`, `bench-smoke`, `bench-ci`, every CI workflow — are untouched.

**The shape recorded where the benchmarks record.** `meta` already carried `ranks`, `nodes`, `ranks_per_node` and `monoprop_threads`, with report columns for each — but `partitions_env` was written *only when the env var was set*, so an undeclared partition count and an unrecorded one both rendered `—`. It is now recorded unconditionally, `"unset"` when absent.

## Model sizes are inputs now

| was | now |
| --- | --- |
| `--pauli-num-qubits` accepted any value while `HEAVY_HEX_TOPOLOGY` stayed the fixed 127-qubit IBM Eagle map — silently a different model, or an index past the operator | raises, naming the topology as the reason and pointing at `--pauli-lower-atol` for sizing |
| `--hubbard-observable-site` above `--hubbard-num-sites` — a documented "trap" that produced a wrong observable | raises |
| a mode count above the extension's compile-time `MAX_NUM_MODES` failed deep inside the extension | raises, naming the limit and the cmake define |
| `--pauli-observable-qubit` outside the register | raises |

`LADDER.md` gains a *Model size knobs* table covering these plus `--obs-terms`, which is an upper bound rather than an exact count: monomials are drawn independently and duplicates collapse, by about `obs_terms / 2·C(2·num_modes, gen_length)`. Checked against a direct count at 142 modes — 0.03/0.19/0.37% measured at 200k/1M/2M draws against 0.04/0.19/0.38% predicted — giving 0.06% at L1's 295k and 2.8% at L2's 14.75M. Deterministic for a fixed `--seed`, so the calibration reproduces.

## Also

- `just bench-mpi` never forwarded `monoprop_PARTITIONS` or `monoprop_NUM_THREADS` through `mpiexec -x`, though its own doc comment showed `monoprop_NUM_THREADS=2 just bench-mpi …`. Both are forwarded now, guarded so an unset one is not an mpiexec error.

## Traps, each from a failure it caught during calibration

- **`srun --cpu-bind=cores` with no `--cpus-per-task` on the `srun` confines each task to one core.** Measured in one allocation: no flags → 128, `--cpus-per-task=128` → 128, `--cpu-bind=cores` alone → 1. ~100x, with nothing in the timing to say so.
- **The report runs outside `srun`.** Under it, one process per rank races to write the same file. The two JSON artifacts are safe: only rank 0 writes them.
- **A two-operation row's peak is the MAX over its operations, never the sum.** `HighWaterMark` resets `VmHWM` per benchmark, so both windows contain the same resident operator.
- `--cpu-bind=none` cost 1.45x; `nproc` lies inside a job.
- The `monoprop_PARTITIONS` and observable-site traps are no longer traps — both now raise.

## The L2a / L2b split

`L2` → `L3` moved two things at once — multiple processes *and* the network — so anything that
appeared at L3 was attributable to neither. L2 splits: **L2a** is the old L2 unchanged (one rank,
`partitions = threads = cores`), and **L2b** is L3's shape at `N`=1. Each edge of the ladder now
moves one thing: partitions, processes, the network, then node count.

L2b is measured at `N`=1, `R`=8, `P`=16 on a 128-core node, two reps with the cell order flipped
(job 1862701, `_core.so` md5 `b201ec44`). Term counts reproduce L2a's and L3's **exactly** —
1,001,661,534 / 985,970,588 / 948,937,993 / 597,445,055 — which is the geometry-independence check
the rung relies on.

`8 × 16` is stated as a choice, not derived as an optimum: one rank per NUMA domain on this
machine, and L3's shape, so `L2b` → `L3` differs only in the node count. The section says so rather
than implying a sweep.

Two findings recorded next to the table:

- **The per-rank cost that is not a share of the operator is 3.6–8.1 GiB on the Heisenberg rows and
  0.2–0.6 GiB on the Schrödinger ones**, on one shape, one node and one binary. The flag sets
  differ in the observable — 14.75M terms against 200k. The raw observable is well under a GiB, so
  something indexed by it scales with it; that is not attributed further here.
- **L2b, not L2a, is the memory-worst rung for the Heisenberg rows**, because the per-rank cost is
  paid `R` times with no extra nodes to spread it over. An earlier draft of this change claimed a
  row fitting L2a fits L2b; the measurement refutes it and the claim is gone.

`~GiB/node` is a sum over ranks, as L3's column is, so reading it against L2a's single-process peak
would be a sum-against-max comparison — the error that produced a phantom 2.3× in this project
before. The table says so. Wall time is *not* compared across the two rungs: the same 1×128 work
measured 140 s in L2a's job and 176 s in another allocation, which is exactly why that comparison
needs both arms in one job.

Also in this push: a new trap — `P` above the cores a rank can see, which is what carrying L2a's
`monoprop_PARTITIONS` into an L2b launch does — and the multi-rank skeleton now names the
interpreter directly, because `uv run` re-resolves the environment in every rank.

**Stacked on top:** #323 rebuilds the continuous benchmarking around these rungs.

## Verification

- `uv run pytest`: **622 passed, 8 skipped** — unchanged from #316.
- `prek run --all-files` over every changed file: clean.
- The multi-rank gate, on a 2-rank job: unset → `ERROR: monoprop_PARTITIONS is unset on a run of 2 ranks…` from both ranks, exit 4; set to 8 → runs, and `{"ranks": 2, "nodes": 1, "ranks_per_node": 2, "partitions_env": "8", "monoprop_threads": "8"}` lands in `results/<label>.json` and in the report's Configuration table. One rank with the knob unset still runs.
- Each rejection exercised through the CLI: `--pauli-num-qubits=100` and `--hubbard-observable-site=70` each fail with the message naming the constraint.
- Every pytest flag in `LADDER.md` checked against `pytest benches --help`; all present. Every `-k` run through `--collect-only`; each selects exactly the intended node ids.


benches/LADDER.md
```diff
@@ -0,0 +1,240 @@
...
+transfers to your machine and a disagreement means the *problem* changed. The `~s` and `~GiB/node`
+columns are one measurement on 128-core, 242 GiB nodes, rounded to a figure you can size a job
+with. They are not a baseline; your baseline is the merge base measured next to your branch.
+
+## L1 — one thread, ~10M terms
+
+One node, one rank, one partition: `export monoprop_PARTITIONS=1 monoprop_NUM_THREADS=1`.
+Removes every threading and MPI effect, so what moves is the kernel.
+
+| row | flags | `-k` | terms | ~s | ~GiB |
```
diagonal-hamiltonian: could you also benchmark the regular gradient, and the gradient with pare_threshold = 1e-10. bump terms in both to 20M

benches/LADDER.md
```diff
@@ -0,0 +1,240 @@
...
+GiB per Mterm (`propagate`, `energy`, `build_graph`, `gradient`) — within 15% of each other, so no
+operation is a multiple of another.
+
+## L3 — four nodes, the same problems
+
+Same flags, same term counts, same cores per node as L2; 8 ranks per node × 16 partitions instead
+of 1 × 128, so `R` = 32 and `P` = 512. Directly comparable to L2 row for row, which is the point of
+having both.
+
+| row | ~s | ~GiB/node | against L2 |
```
diagonal-hamiltonian: remove the faster than and less memory, give ballparks like in the others and not comparing against L2

benches/LADDER.md
```diff
@@ -0,0 +1,240 @@
...
+observable terms all give ~19M. Nor can the other axes compensate — `--cutoff 8` and
+`--num-modes 200` each exceeded a 242 GiB node. ~19M is the picture's ceiling, measured.
+
+At ~949M terms on one rank over 128 cores the four operations cost 0.100, 0.107, 0.115 and 0.115
+GiB per Mterm (`propagate`, `energy`, `build_graph`, `gradient`) — within 15% of each other, so no
+operation is a multiple of another.
+
+## L3 — four nodes, the same problems
+
+Same flags, same term counts, same cores per node as L2; 8 ranks per node × 16 partitions instead
```
diagonal-hamiltonian: make this general to whatever the user puts in and make it manditory to supply ranks/partition config

benches/LADDER.md
```diff
@@ -0,0 +1,240 @@
...
+resident floor**, paid once per rank, so eight ranks on a node pay it eight times: on random
+`build_graph` that is 75.6 of the measured 104.2 GiB/node before the build allocates anything.
+More ranks per node buys speed and costs a floor, and it hurts most where the operation's own cost
+is smallest — which is why the 19M-term Schrödinger rows cost more per node on four nodes than on
+one. The benchmarks page has the per-rank breakdown.
+
+## L4 — strong and weak scaling
+
+Hubbard `propagate` only: a communication cost is a function of `P`, so varying the model as well
+would confound it. 8 ranks per node × 16 partitions throughout, so `R` = 8N and `P` = 128N.
```
diagonal-hamiltonian: make general and force inputing config. Make sure its recorded where the benchmarks record

benches/LADDER.md
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Added, measured at one rank / one partition / one thread on Deucalion (`7965c98`, `_core.so` md5 `b201ec4`). `--obs-terms=295000` lands 19,902,244 terms; both arms are the same problem, term count identical to the digit.

| row | ~s | ~GiB |
| --- | ---: | ---: |
| random `gradient` | 11.05, 11.02 (2 reps) | 2.5 |
| random `gradient`, `--pare-threshold=1e-10` | 0.75, 0.75 (2 reps) | 2.8 |

`--pare-threshold` did not exist, so it is now a pytest option wired into `expectation_value_functional` and `expectation_value_and_gradient_functional` for both the random and the fixed-model benchmarks; unset keeps the exact graph, which is what every existing run and the CI series already measure. Verified it reaches the engine rather than being swallowed: at `1e-10` the energy is bit-identical to the unpared value (the point of a retention cutoff), while an absurd `1e10` moves it 0.8049 → 0.8307.

Hubbard was not a candidate here — 29 successive `build_graph` calls retain 29 layer-sets and `skip_if_graph_will_not_fit` skips it — so the gradient rows use the random problem, whose size is directly dialable.

benches/LADDER.md
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Done. The "against L2" column is gone; L3 now reads like L2 and L1 — flags, `-k`, terms, `~s`, `~GiB/node` — with no ratio anywhere in the table. The one sentence under it says which shape those cells were measured at (`N`=4, `R`=8, `P`=16, medians of two reps) so they can be read as a job-sizing ballpark rather than a target, plus the per-process memory floor, which is the reason memory per node does not fall the way time does.

benches/LADDER.md
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Generalised, and mandatory in code rather than in prose.

L3 is now "several nodes, the same problems": `N`, `R` and `P` are yours. Choosing `R × P` equal to L2's core count is what makes the L2/L3 pair isolate per-process cost; anything else is still a valid run, just not that comparison. The `srun` template takes `R` and `P` as substitutions instead of hard-coding 8 × 16.

Forcing the config: `benches/conftest.py` now refuses to start a session of more than one rank with `monoprop_PARTITIONS` unset. The engine's own default is `ranks == 1 ? cores : 1`, so an unset knob measured one partition per rank at a plausible wall time — the exact failure the old "traps" bullet described. Every rank raises after the collective `_nodes()`, so nothing is left in a collective. Verified on a 2-rank job: unset → `ERROR: monoprop_PARTITIONS is unset on a run of 2 ranks...` from both ranks, exit 4; set → runs clean. One-rank runs (`just bench`, `bench-smoke`, `bench-ci`, every CI workflow) are untouched, and `uv run pytest` is still 622 passed / 8 skipped.

benches/LADDER.md
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Same generalisation: L4 holds `R` and `P` fixed across a ladder and varies only `N`, with `P_total` = `N × R × P`; the atol table and the weak/strong rules are unchanged because they were only ever functions of `N` and the size row. The stated `R`=8, `P`=16 is now labelled as the shape the timings came from, not as the prescribed shape. Same hard gate as L3 — a multi-rank session without `monoprop_PARTITIONS` refuses to start.

On recording: it was already partly there and partly broken. `meta` carries `ranks`, `nodes`, `ranks_per_node` and `monoprop_threads`, and `REPORT.md` has had columns for them — but `partitions_env` was only written when the env var was set, so an undeclared partition count and an unrecorded one both rendered as `—`. It is now recorded unconditionally, as `"unset"` when absent. Verified end to end on the 2-rank job: `{"ranks": 2, "nodes": 1, "ranks_per_node": 2, "partitions_env": "8", "monoprop_threads": "8"}` in `results/<label>.json`, and the same values in the report's Configuration table. `LADDER.md` now points at that table and says to check it before comparing two rows.

Related: `just bench-mpi` never forwarded `monoprop_PARTITIONS` or `monoprop_NUM_THREADS` through `mpiexec -x`, though its own doc comment showed `monoprop_NUM_THREADS=2 just bench-mpi ...`. Both are forwarded now, guarded so an unset one is not an mpiexec error.

AGENTS.md
```diff
@@ -1,199 +1,63 @@
...
+  `lint.yml` runs the same hooks over the PR's commit range, so a skipped lint is a red PR.
 
+## Git and PRs
 
-## Architecture Overview
+- Commits and PR titles: `<type>(<optional scope>): <gitmoji> <description>`. Types: `feat`
+  (minor), `fix` (patch), `docs`, `style`, `refactor`, `test`, `chore`; `!` for breaking changes.
+- Agent commits carry `Assisted-by: <harness>:<model>` and no `Co-authored-by`.
+- Follow `.github/PULL_REQUEST_TEMPLATE.md`; prefix agent-authored descriptions and comments with
+  `:robot: _AI text below_ :robot:`.
```
robertodr, line 30:
```suggestion
  `:robot: _AI text below_ :robot:`. **Do not wrap the text** in PR descriptions.
```

robertodr (dismissed):
split the changes to `AGENTS.md` to a separate PR.

Panadestein (approved):

PR #323, opened 2026-09-01
chore(ci): 👷 benchmark the ladder rungs, one testbed per shape and silicon

diagonal-hamiltonian:

Stacked on #317, which defines the rungs. This is the continuous-benchmarking side only.

## What was wrong

The benchmark lanes declared no shape and no problem. `monoprop_PARTITIONS` was unset everywhere,
so `resolve_partition_count_`'s `ranks == 1 ? cores : 1` decided it — every point in the
`ubuntu-26.04` and `aws-c7i-4xlarge` series was one partition per core of whatever runner it landed
on, unrecorded. A runner resize would have moved every series without moving a testbed name.

The first revision of this PR fixed the shape and left the problem: the rungs were named after
`benches/LADDER.md` but ran the suite's dataclass defaults, so the sizes had nothing to do with the
rungs they were named after, and nobody had measured what they cost.

## What it does now

Each rung declares **a shape, a round count and a problem**, and the tables below are measured, not
derived. `rungs` is one rung per line, `<name> <ranks> <partitions> <rounds> | <pytest args>`.

| rung | `R` × `P` | launch | `monoprop_PARTITIONS` | rounds | testbed |
| --- | --- | --- | ---: | ---: | --- |
| L1 | 1 × 1 | `pytest benches` | 1 | 3 | `<cpu>-8c-L1` |
| L2a | 1 × 8 | `pytest benches` | 8 | 1 | `<cpu>-8c-L2a` |
| L2b | 4 × 2 | `mpiexec -n 4 --map-by slot:PE=2 --bind-to core` | 2 | 1 | `<cpu>-8c-L2b` |

`R × P` = 8 at both L2 rungs, so L2a → L2b moves the process count and nothing else. Rounds is 1
above L1 because `pedantic` builds round *k+1*'s propagator before releasing round *k*'s — at these
sizes that is a memory setting, not a statistics setting.

### L1 — LADDER.md's own rows, one thread

| row | flags | `-k` | terms | ~s | ~GiB |
| --- | --- | --- | ---: | ---: | ---: |
| hubbard `propagate` | `--hubbard-cutoff=10 --hubbard-lower-atol=4.2e-05` | `test_model_propagate and hubbard` | 9,953,109 | 25 | 0.87 |
| pauli `propagate` | `--pauli-cutoff=12 --pauli-lower-atol=1.22e-04` | `test_model_propagate and pauli` | 10,069,308 | 21 | 1.06 |
| random `gradient` | `--num-generators=1000 --num-modes=142 --cutoff=6 --obs-terms=295000` | `test_random_gradient and heisenberg` | 19,902,244 | 11 | 2.60 |

One process, one selector — the three flag sets use disjoint options. Whole-rung peak 2.60 GiB in
3:53. The term counts reproduce LADDER.md's L1 table exactly. LADDER.md's fourth L1 row, the same
gradient pared at `1e-10`, is dropped: `--pare-threshold` is session-wide and cannot share a
process with the unpared row.

### L2a and L2b — one operator, three operations

`build_graph` publishes the graph `energy` and `gradient` evaluate, so the three share one operator
and the peak is the max over them, not the sum. `propagate` is excluded: it holds its own. Both
rungs take the same flags.

`-k "(test_random_build_graph or test_random_energy or test_random_gradient) and heisenberg"`
`--num-generators=1000 --num-modes=142 --cutoff=6 --obs-terms=2500000`

| row | terms | ~s L2a | ~GiB L2a | ~s L2b | ~GiB/node L2b | ~GiB worst rank |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| random `build_graph` | 167,515,463 | 83 | 19.11 | 68 | 23.71 | 6.14 |
| random `energy` | 167,515,463 | 13 | 17.54 | 6 | 20.36 | 5.17 |
| random `gradient` | 167,515,463 | 39 | 18.85 | 16 | 21.68 | 5.50 |

Peak 19.11 GiB at L2a and 23.71 GiB/node at L2b, in 135 s and 90 s. `~GiB/node` is the sum over the
node's ranks; it errs high, because a page shared between ranks is charged to each.

`--obs-terms=2500000` is **the largest measured point that fits**, not an interpolation: the next
step tried, 3.5M, reaches 35.49 GiB/node against the instance's 32 GiB. A two-point fit over 1.5M
and 2.5M predicted 31.1 GiB there and under-called it by 3.9, so this axis is not interpolated.

The term count is identical at both shapes — the geometry-independence check, which the `terms`
measure holds to 0%. L2b's node sum is 1.24x L2a's peak here against 1.29x at 104M terms, so the
per-rank multiplier falls as the operator grows.

## Review findings fixed

- **`--bench-rounds 3` was a 2× memory error**, not a statistics choice: `pedantic` holds round
  *k+1*'s propagator alongside round *k*'s. Rounds is now per rung, 1 on the sized ones.
- **Placement was printed, never asserted**, and from rank 0's fields rather than the `*_min`
  reductions that exist to catch a partial collapse. `check_shape.py` now fails a rung when
  `partitions > 1` and `single_cpu_threads_min == 0`. It does not fire on an all-zero summary,
  which means `/proc` was unreadable rather than nothing pinned.
- **Pinning was the caller's option.** `--map-by slot:PE=$monoprop_PARTITIONS --bind-to core` moved
  into `bench-ci-mpi`, which cannot now run unpinned and refuses without `monoprop_PARTITIONS`.
  `*ARGS` is freed for the pytest args the rungs need.
- **The testbed name reads cpuinfo that is already recorded.** `cpu_slug.py` takes
  `machine_info.cpu.brand_raw` out of the `time-<label>.json` pytest-benchmark already writes, so
  the name derives from an artifact the run uploads rather than from a second collector. No
  dependency is added: `py-cpuinfo` is present as pytest-benchmark's own.
- Comments cut to one line each; `benchmarks.mdx`'s prose replaced by the tables above.

### Second review round

- **`ranks`/`partitions` were unvalidated**, so a typo coerced to 0 in the bash arithmetic and
  failed inside `mpiexec -n <word>` instead of naming the field. All three fields now require a
  positive decimal, `per-rank` excepted — which also rejects `0` (division by zero at `per-rank`,
  `mpiexec -n 0` otherwise) and leading zeros (`08` is an invalid octal literal there).
- **`BENCH_CORES` ignored the affinity mask.** `psutil.cpu_count(logical=False)` counts the whole
  machine, so under a cpuset or taskset every rung was sized from cores the run never had. It is
  now the distinct `(package, core)` pairs among `sched_getaffinity(0)`, read from `/sys` — still
  physical, so an SMT pair on one core resolves to 1. Unreadable topology falls back to `psutil`
  only when nothing is masked, and otherwise refuses.
- **`cpu_slug.py` raised on a malformed artifact.** A truncated file or a missing
  `machine_info.cpu.brand_raw` is now one `::error::` naming the file and the field.
- `bench-ci-resolve-cores` prints `BENCH_CORES=<n>` when `GITHUB_ENV` is unset, so the ladder runs
  outside Actions; `bench-ci-rungs` names the recipe that supplies it.

## What is not measured

**On this silicon, not the runner's.** Every cell was measured on an 8-core mask of a Deucalion
x86 node (`znver2`, SMT off). Term counts are deterministic at `--seed=0` and transfer exactly;
memory transfers well; the `~s` column is indicative for Sapphire Rapids.

**SMT is off on the calibration machine.** `PE=2` resolved against cores there, so each rank got a
2-CPU mask. On the runner (SMT on) `--bind-to core` should give each rank 2 cores / 4 hwthreads and
the engine 2 physical cores; the assertion holds either way, but the mask figure will differ.

**L2b's 4 ranks is a judgement, not an optimum.** The instance is one NUMA domain, so LADDER.md's
one-rank-per-domain rule picks no number. 4 × 2 makes L2b the memory-worst rung, which is the right
place to hang the ceiling, and keeps more than one partition per rank where 8 × 1 would not.

**Nothing benchmarks until a maintainer sets `vars.BENCH_BARE_METAL`.** `bench_main.yml` is deleted
and the README badge with it, rather than pointing at a workflow that never runs: a GitHub-hosted
runner has ~2 physical cores, making L2a 1 × 2 and L2b 2 × 1 — restatements of L1 through a noise
floor no threshold survives.

## Verification

| check | result |
| --- | --- |
| L1 / L2a / L2b at the flags and shapes above, on an 8-core mask | 11/11 runs rc=0, job COMPLETED in 23:01 |
| L1 term counts against LADDER.md's L1 table | exact: 9,953,109 / 10,069,308 / 19,902,244 |
| terms identical at 1x8 and 4x2 | 167,515,463 both |
| `--obs-terms` bracketing | 2.5M = 23.71 GiB/node fits; 3.5M = 35.49 does not |
| placement at 1x8 (`P` == physical cores) | 8 of 11 threads on a CPU each, 8-CPU mask — the boundary does not collapse |
| placement at 4x2 under the CI `mpiexec` line verbatim | 2 of 5 threads pinned, 2-CPU mask per rank — `PE=2` resolved against cores, not hwthreads |
| `mpiexec -n 4 --map-by slot:PE=2 --bind-to core` vs `srun --cpu-bind=cores` | identical shape and pinning |
| rung parser, 18 cases | real block, rung-after-MPI (fd 3), unbalanced quote, quotes-only, missing rounds, non-numeric rounds, oversubscribed, `per-rank`→0, no args, comments/blanks, plus `0` and `08` in each of the three fields — all as intended; harness generated from `bench.yml` itself |
| `-k` expression through `just` | arrives as one argv element; `bench-ci-mpi` refuses without `monoprop_PARTITIONS` (rc=1) |
| `cpu_slug.py` | `AMD EPYC 7742 64-Core Processor` → `amd-epyc-7742-64-core` end-to-end; truncated JSON, missing `machine_info`, `"cpu": null` and no artifact each give one `::error::` naming the file |
| `prek run --from-ref <base> --to-ref HEAD` | all 24 hooks pass, rc=0 — the gate `lint` runs |
| `check_shape.py` matrix | 7/7: the three real results pass; wrong partitions, wrong ranks and a collapsed placement each fail; an all-zero pinning summary does **not** false-alarm |
| `visible_physical_cores` against a synthetic 16-hwthread / 8-core / 1-package topology | 16 hwthreads → 8; an 8-hwthread mask → 8; an SMT pair on one core → 1; unreadable `/sys` → `None` |
| rank 0 is the only writer of `time-<label>.json` | `benches/conftest.py:283-290` nulls the writer off rank 0, and `--benchmark-json` is `type=Path`, so no rank opens the file at parse time |
| `pytest packages/monoprop-bench-tools/tests` | 38 passed |
| all three workflow YAML files | parse |

### Third review round

- **The CI-only recipes are hoisted into `bench.yml`.** All six `bench-ci*` recipes are gone from
  the `justfile`, which keeps only `bench`, `bench-mpi`, `bench-build-mpi` and `bench-smoke`. Three
  `(bench.yml, <step name>)` entries added to `ALLOWED` in `tools/check-workflow-commands.py`.
  `bench-ci-resolve-cores`'s body became `.github/workflows/scripts/resolve_cores.py`. The rung
  loop moved verbatim; the 16 parser cases behave identically against the extracted `run:` block.
- **The badge is back**, pointing at `bench_bare_metal.yml` — the workflow that actually runs.
- **`pytest-benchmark` 5.3.0**, in its own commit, which swaps `py-cpuinfo` for `py-cpuinfo2`.
  `machine_info.cpu.brand_raw` survives the fork: `Apple M5` → `apple-m5` end to end.

## Summary by CodeRabbit

- **New Features**
  - Added structured serial and MPI benchmark ladders with configurable execution steps.
  - Added validation for benchmark shape, CPU configuration, and core availability.
  - Added separate result artifacts and CPU-specific Bencher tracking for each benchmark rung.

- **Documentation**
  - Expanded benchmarking guidance for local execution, resource requirements, artifacts, and ladder configuration.
  - Updated benchmark badge and tool usage examples.

- **Chores**
  - Removed the dedicated main-branch benchmark workflow and consolidated benchmarking in the reusable workflow.


.github/workflows/bench.yml
```diff
@@ -1,8 +1,8 @@
 name: Benchmark
 
-# Shared body of the Bencher continuous-benchmarking runs. Callers choose the
-# machine and the Bencher testbed; everything downstream is identical, so the
-# testbeds stay comparable in shape even though their absolute numbers are not.
+# Shared body of the Bencher runs; the caller names the machine and the benches/LADDER.md rungs.
+# Every rung runs on the same runner from the same MPI-enabled build and uploads to its own
+# testbed, because Bencher keys history on (branch, testbed, benchmark, measure).
```
robertodr:
```suggestion
```

.github/workflows/bench.yml
```diff
@@ -52,64 +46,47 @@ jobs:
         with:
           cache-suffix: benchmark
 
+      # uv's `cache-keys` cover neither `SKBUILD_*` nor `monoprop_ENABLE_MPI`, so without the
+      # reinstall the warm cache above can hand back an MPI-off wheel and every rung still pass.
```
robertodr:
```suggestion
```

.github/workflows/bench_bare_metal.yml
```diff
@@ -1,12 +1,9 @@
...
-# provisioned per run by RunsOn. The point is the noise floor: nothing else is
-# scheduled on the machine, so this testbed can eventually carry thresholds
-# tight enough to gate on, which the shared GitHub runner cannot.
-#
-# The node exists only for the duration of the run, so cost tracks the number
-# of commits to main rather than wall-clock time.
+# monoprop's continuous benchmarking, on a dedicated EC2 node provisioned per run by RunsOn.
+# The point is the noise floor: nothing else is scheduled on the machine, so these testbeds can
+# carry thresholds tight enough to gate on. The node exists only for the run, so cost tracks the
+# number of commits to main rather than wall-clock time.
```
robertodr:
```suggestion
```

.github/workflows/bench_bare_metal.yml
```diff
@@ -20,22 +17,21 @@ concurrency:
...
-      # Names the hardware, not the provider: a different instance type is a
-      # different testbed, because its history cannot be compared to this one.
-      testbed: aws-c7i-4xlarge
       label: ci-bare-metal
-      # A dedicated node can afford the slow fixed models
-      bench-markers: ""
-      bench-rounds: "3"
+      # Three rungs of benches/LADDER.md on one node from one build. L1 is that document's own L1
+      # rows; L2a and L2b share one problem sized to ~25 GiB of the instance's 32 GiB and differ
+      # only in shape. See docs/content/docs/benchmarks.mdx for the tables and the 4x2 choice.
```
robertodr:
```suggestion
```

docs/content/docs/benchmarks.mdx
```diff
@@ -335,27 +340,86 @@ measures are recorded:
...
+| random `gradient` | 167,515,463 | 39 | 18.85 | 16 | 21.68 | 5.50 |
+
+Peak 19.11 GiB at L2a and 23.71 GiB/node at L2b, in 135 s and 90 s. `~GiB/node` is the sum over
+the node's ranks — it errs high, because a page shared between ranks is charged to each, so it is
+the number to provision against and the wrong number to divide.
+
+`--obs-terms=2500000` is **the largest measured point that fits**, not an interpolation: the next
+step tried, 3.5M, reaches 35.49 GiB/node, past the 32 GiB the instance has. A two-point fit over
+1.5M and 2.5M predicted 31.1 GiB there and under-called it by 3.9, so this axis is read off
+measured points rather than interpolated.
```
```suggestion
`--obs-terms=2500000` is **the largest measured point that fits**.
```

docs/content/docs/benchmarks.mdx
```diff
@@ -335,27 +340,86 @@ measures are recorded:
...
+
+**No benchmark workflow gates a pull request.** It tracks `main`, and a threshold is a report, not
+a verdict — measuring a change is the author's job, per "Benchmarking ladder" above. Every run
+uploads each rung's `bmf-<rung>.json` and raw result JSON as an artifact, so a failed Bencher
+upload does not take the measurement with it.
+
+The workflow is inert until a maintainer sets the `BENCH_BARE_METAL` repository variable to
+`true`; without a RunsOn stack behind the label the job would queue until it timed out, once per
+commit. There is no shared-runner series: a GitHub-hosted runner has about two physical cores, so
+L2a would be 1 × 2 and L2b 2 × 1 — restatements of L1 through a noise floor no threshold survives.
```
```suggestion
```

pyproject.toml
```diff
@@ -236,6 +236,8 @@ exclude = [
   "packages/monoprop-bench-tools/src/monoprop_bench_tools/_version.py",
   "docs/conf.py",
   "performance_scripts/*.py",
+  # CI glue, not a package: prints to the workflow log and lives outside any module tree.
```
```suggestion
```

README.md
```diff
@@ -4,7 +4,6 @@
 [![Documentation](https://github.com/Algorithmiq/monoprop/actions/workflows/docpages.yml/badge.svg)](https://docs.monoprop.algorithmiq.tech/)
 [![Test monoprop](https://github.com/Algorithmiq/monoprop/actions/workflows/test.yml/badge.svg)](https://github.com/Algorithmiq/monoprop/actions/workflows/test.yml)
 [![codecov](https://codecov.io/gh/Algorithmiq/monoprop/graph/badge.svg?token=1GgmPnNUxk)](https://codecov.io/gh/Algorithmiq/monoprop)
-[![Track benchmarks](https://github.com/Algorithmiq/monoprop/actions/workflows/bench_main.yml/badge.svg)](https://github.com/Algorithmiq/monoprop/actions/workflows/bench_main.yml)
```
robertodr, line 7: Why did this badge disappear?

robertodr (changes requested):
The new recipes in `justfile` are used (and useful) exclusively in the CI job. They can be stated inline in the YAML file for the github workflow. The `bench-ci` recipe predates this change, but also that one should go away from the `justfile`.

robertodr:
Also, I still maintain that using [`py-cpuinfo2`](https://github.com/akx/py-cpuinfo2) would give more complete information on how the architecture you're running on

diagonal-hamiltonian (commented):

I tried this and had to back it out: `prek` rejects it.

`tools/check-workflow-commands.py` fails any workflow `run:` block containing `uv run`, `pytest` or `mpiexec`. Inlining the recipes produces:

```
check workflows call recipes.............................................Failed
  .github/workflows/bench.yml: step 'Resolve the core count' runs 'uv run' — call the recipe that owns the command
  .github/workflows/bench.yml: step 'Run the ladder rungs' runs 'uv run' — call the recipe that owns the command
  .github/workflows/bench.yml: step 'Run the ladder rungs' runs 'pytest' — use `just test-py` / `just test-py-mpi`
  .github/workflows/bench.yml: step 'Run the ladder rungs' runs 'mpiexec' — use the MPI recipe for the leg being run
  .github/workflows/bench.yml: step 'Track benchmarks with Bencher' runs 'uv run' — call the recipe that owns the command

  Commands belong in the justfile; workflows choose matrices, environment and artifacts.
```

That hook is yours, from #326 (`refactor(ci): ♻️ make the justfile the single source of build commands`), and its docstring is the argument against this change:

> Workflows choose matrices, environment and artifacts; the commands themselves live in recipes, so that a failing lane can be reproduced locally by running what it ran. This guards that split, which nothing else can: a copy pasted back into a `run:` block works perfectly well until it drifts from the recipe it duplicates.

So I've left the recipes where they are. Which way do you want it?

1. **Keep the recipes** (what's pushed). `bench.yml` stays four `just` calls, and the ladder is reproducible by hand.
2. **Inline anyway**, and I add three `(bench.yml, <step name>)` entries to `ALLOWED`. I have this version working — the rung loop, the Bencher loop and all ten parser cases behave identically inlined — so it's a `git apply` away if you want it.
3. **Drop `uv run` / `pytest` / `mpiexec` from `FORBIDDEN`** entirely, which un-guards every other workflow too.

I'd take (1), and not only because of the hook: the recipes are the only reason a maintainer can reproduce a rung. On the reproducibility point, two things Copilot flagged are now fixed either way — `bench-ci-resolve-cores` prints `BENCH_CORES=<n>` instead of requiring `GITHUB_ENV`, and `bench-ci-rungs` names the recipe that supplies it — so the whole ladder now runs outside Actions:

```bash
export BENCH_CORES=8 RUNGS="L1 1 1 3 | -k 'test_model_propagate and hubbard'"
just bench-ci-rungs local
```

But if you'd rather the benchmark workflow be self-contained and take the exceptions, say so and I'll push (2).

README.md
diagonal-hamiltonian: :robot: _AI text below_ :robot:

It pointed at `bench_main.yml`, which this PR deletes — the badge URL would have 404'd.

`bench_main.yml` was the GitHub-hosted series. It goes because a GitHub-hosted runner has ~2 physical cores, which makes L2a 1 × 2 and L2b 2 × 1 — restatements of L1 through a noise floor no threshold survives. So there is no shared-runner rung left to badge.

I didn't repoint it at `bench_bare_metal.yml` because that workflow is inert until a maintainer sets `vars.BENCH_BARE_METAL`, so the badge would read `no status` in the README indefinitely. Two options, tell me which:

- **Leave it out** until the variable is set, then add `[Track benchmarks](.../bench_bare_metal.yml)` in the same commit that flips it.
- **Add it now** and accept `no status` in the meantime.

A Bencher perf badge (`https://bencher.dev/perf/monoprop/badge`) is a third option and arguably the more useful one, since it tracks the measurement rather than the workflow's exit code — but it needs the project to have data first.

.github/workflows/scripts/cpu_slug.py
```diff
@@ -0,0 +1,47 @@
...
+    return SEPARATORS.sub("-", NOISE.sub(" ", brand).lower()).strip("-") or "unknown-cpu"
+
+
+def main() -> str | None:
+    paths = sorted(Path("benches/results").glob("time-*.json"))
+    if not paths:
+        return "::error::no timing artifact to read the CPU from"
+    brand = json.loads(paths[0].read_text())["machine_info"]["cpu"]["brand_raw"]
+    print(slugify(str(brand)))
+    return None
```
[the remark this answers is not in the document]
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Fixed in 06265d77. Both failure modes now name the file and the field:

```
::error::benches/results/time-a.json is not valid JSON: Expecting value: line 1 column 17 (char 16)
::error::benches/results/time-a.json has no machine_info.cpu.brand_raw
```

`TypeError` is caught alongside `KeyError` so a `"cpu": null` — a schema change rather than a missing key — lands on the same line rather than a traceback.

justfile
```diff
@@ -404,13 +399,143 @@ bench-ci LABEL *ARGS:
...
+    # count would cross the `partitions > visible cores` threshold on any SMT machine.
+    cores = psutil.cpu_count(logical=False)
+    if not cores:
+        sys.exit("cannot determine the physical core count on this runner")
+
+    print("variant ", monoprop.__variant__)
+    print("cores   ", cores, "physical,", psutil.cpu_count(logical=True), "logical")
+    print("affinity", len(os.sched_getaffinity(0)))
+    with open(os.environ["GITHUB_ENV"], "a") as env:
+        print(f"BENCH_CORES={cores}", file=env)
```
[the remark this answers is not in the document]
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Fixed in 06265d77. `bench-ci-resolve-cores` now writes to `GITHUB_ENV` when it exists and prints the assignment when it doesn't:

```python
if path := os.environ.get("GITHUB_ENV"):
    with open(path, "a") as env:
        print(f"BENCH_CORES={cores}", file=env)
else:
    print(f"BENCH_CORES={cores}")
```

The second site you flagged is `bench-ci-rungs`, which read `$BENCH_CORES` under `set -u` and would have aborted with `unbound variable`. It now says which recipe supplies it:

```
: "${BENCH_CORES:?bench-ci-rungs needs BENCH_CORES; run just bench-ci-resolve-cores first}"
```

So the whole ladder runs outside Actions:

```bash
export BENCH_CORES=8 RUNGS="L1 1 1 3 | -k 'test_model_propagate and hubbard'"
just bench-ci-rungs local
```

.github/workflows/scripts/check_shape.py
```diff
@@ -0,0 +1,60 @@
...
+        return (
+            f"::error::{path.stem} ran at ranks={meta['ranks']} "
+            f"partitions_env={meta['partitions_env']}, asked for {ranks}x{partitions}"
+        )
+    # An all-zero summary means /proc was unreadable, not that nothing was pinned.
+    if int(partitions) > 1 and threads and not placed:
+        return (
+            f"::error::{path.stem} placed no thread on a CPU of its own. "
+            "partitions above the rank's visible cores returns an empty placement order."
+        )
```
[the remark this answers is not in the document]
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Fixed in 06265d77.

diagonal-hamiltonian:

@robertodr on `py-cpuinfo2` — agreed on the substance, and I want to separate two things that got tangled in the earlier revision of this branch.

**What I object to is the wiring, not the package.** `py-cpuinfo2` was in via a project-wide `[tool.uv] override-dependencies`, and that override applies to the whole resolution, not just the bench group — so it replaced `py-cpuinfo` for every consumer, `pytest-benchmark` included. That silently changes what produces the `machine_info` block this PR now keys the testbed name off, in every lane, not just the benchmark one. That is the part that should not come back.

**Where the tree stands now** (after the rebases, both are gone):

- no `py-cpuinfo2` and no `override-dependencies` anywhere in `pyproject.toml` or `uv.lock` — `git log -S` finds neither in this branch's history
- `py-cpuinfo 9.0.0` is present as `pytest-benchmark`'s own dependency (`uv.lock:1768`), which is what writes `machine_info.cpu` into `time-<label>.json`
- `cpu_slug.py` reads exactly one field from it, `machine_info.cpu.brand_raw`, and that JSON is uploaded as an artifact, so the record survives the run

I can't reproduce the field-by-field comparison from the earlier revision — those commits were squashed away — so I'll not restate it as measured.

**Concretely, what I'd propose:** add `py-cpuinfo2` to the `bench` dependency group only, with no override, and have the rung loop dump its full `get_cpu_info()` next to `bmf-<rung>.json`. That gets you the richer architecture record (cache sizes, flags, `hz_advertised`, `vendor_id_raw`, stepping) as its own artifact, without touching what `pytest-benchmark` resolves or what the testbed name derives from.

Want that in this PR, or as a follow-up? I'd slightly prefer the follow-up, so a resolution change lands on its own commit rather than inside the one that establishes the testbeds — but it's a small patch either way, so your call.

justfile
```diff
@@ -424,13 +419,160 @@ bench-ci LABEL *ARGS:
...
+      IFS=$' \t' read -r rung ranks partitions rounds <<<"$head"
+      case "${rung:-}" in '' | '#'*) continue ;; esac
+      # A missing field would otherwise fall through to a default and measure something else,
+      # and a non-numeric one reaches the arithmetic below as an unset name, so worth 0.
+      for field in ranks partitions rounds; do
+        case "${!field:-}" in
+          '') echo "::error::rung $rung has no $field"; exit 1 ;;
+          *[!0-9]*)
+            [[ "$field" == partitions && "${!field}" == per-rank ]] && continue
+            echo "::error::rung $rung has a non-numeric $field: '${!field}'"; exit 1 ;;
```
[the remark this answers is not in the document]
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Confirmed, all four cases. Fixed in 50344251.

```
$ bash -c 'set -euo pipefail; ranks=0; partitions=$((8 / ranks))'
bash: 8 / ranks: division by 0 (error token is "s")

$ bash -c 'ranks=0; partitions=1; ((partitions < 1 || ranks * partitions > 8)); ((ranks == 1)) || echo "-> mpiexec -n $ranks"'
-> mpiexec -n 0

$ bash -c 'ranks=08; ((ranks * 1 > 8))'
bash: ((: 08: value too great for base (error token is "08")
```

All three fields now require a positive decimal, `per-rank` excepted:

```bash
for field in ranks partitions rounds; do
  value="${!field:-}"
  [[ "$field" == partitions && "$value" == per-rank ]] && continue
  case "$value" in
    '') echo "::error::rung $rung has no $field"; exit 1 ;;
    *[!0-9]* | 0*)
      echo "::error::rung $rung needs a positive decimal $field, not '$value'"; exit 1 ;;
  esac
done
```

I did not add explicit `10#` prefixes. Rejecting leading zeros removes the only way an octal literal could reach the arithmetic — the one computed value, `partitions=$((BENCH_CORES / ranks))`, is already a bash integer — so `10#` would be dead weight over a guarantee the validation now makes. Say the word if you'd still rather have it belt-and-braces.

Verified over 18 parser cases:

```
ranks=0 fixed P            rc=1  ::error::rung L1 needs a positive decimal ranks, not '0'
ranks=0 per-rank           rc=1  ::error::rung L1 needs a positive decimal ranks, not '0'
partitions=0               rc=1  ::error::rung L1 needs a positive decimal partitions, not '0'
rounds=0                   rc=1  ::error::rung L1 needs a positive decimal rounds, not '0'
ranks=08                   rc=1  ::error::rung L1 needs a positive decimal ranks, not '08'
partitions=08              rc=1  ::error::rung L1 needs a positive decimal partitions, not '08'
rounds=08                  rc=1  ::error::rung L1 needs a positive decimal rounds, not '08'
per-rank still accepted    rc=0
oversubscribed             rc=1  ::error::rung L2b wants 4x4 on 8 physical cores
real 3-rung block          rc=0
```

justfile
```diff
@@ -424,13 +419,152 @@ bench-ci LABEL *ARGS:
...
+    import psutil
+
+    import monoprop
+
+    if not monoprop.has_mpi:
+        sys.exit("monoprop was built without MPI; every rung shares this binary")
+
+    # PHYSICAL cores: that is what the engine enumerates, and sizing from the logical
+    # count would cross the `partitions > visible cores` threshold on any SMT machine.
+    cores = psutil.cpu_count(logical=False)
```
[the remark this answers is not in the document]
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Correct, and it matters more than the "restricted runner" framing suggests: this ladder was calibrated on an 8-core *mask* of a Deucalion node, so running it under a mask is a normal workflow, not an edge case. Fixed in 50344251.

`BENCH_CORES` is now the distinct `(physical_package_id, core_id)` pairs among `os.sched_getaffinity(0)`, read from `/sys/devices/system/cpu/cpu<N>/topology`. That is affinity-aware *and* still physical, which matters on an SMT machine: a 2-hwthread mask covering one core must resolve to 1, not 2.

Taking your second option where the first is impossible — if the topology is unreadable it falls back to `psutil.cpu_count(logical=False)`, but only when nothing is masked, and otherwise refuses:

```
cannot read CPU topology and only 8 of 16 CPUs are visible;
export BENCH_CORES with the physical cores this mask covers
```

Unit-tested against a synthetic 16-hwthread / 8-core / 1-package topology:

```
all 16 hwthreads     -> 8   (expect 8)
mask of 8 hwthreads  -> 8   (expect 8: cpu0-7 are 8 distinct cores)
mask 0,8 (SMT pair)  -> 1   (expect 1: same core)
unreadable /sys      -> None
```

The log line now records both numbers, so a masked run is visible in the artifact rather than inferred:

```
cores    8 visible physical of 64
affinity 16 of 128 logical
```

justfile
```diff
@@ -424,13 +419,152 @@ bench-ci LABEL *ARGS:
...
+    @mkdir -p "{{ bench_results }}"
+    : "${monoprop_PARTITIONS:?bench-ci-mpi needs monoprop_PARTITIONS; it sizes --map-by slot:PE}"
+    label="$1"; ranks="$2"; shift 2; \
+    monoprop_BENCH_LABEL="$label" monoprop_BENCH_RESULTS="{{ bench_results }}" \
+        uv run --no-sync mpiexec -n "$ranks" \
+        --map-by "slot:PE=$monoprop_PARTITIONS" --bind-to core \
+        -x monoprop_BENCH_LABEL -x monoprop_BENCH_RESULTS \
+        -x monoprop_PARTITIONS -x monoprop_NUM_THREADS \
+        python -m pytest benches -o filterwarnings=default \
+        --benchmark-json="{{ bench_results }}/time-$label.json" \
```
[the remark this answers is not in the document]
diagonal-hamiltonian: :robot: _AI text below_ :robot:

This one does not hold — the guard is in `benches/conftest.py`, not the justfile, so it isn't visible from the recipe alone.

Only rank 0 ever writes `time-<label>.json`. `pytest_configure` unregisters the reporter and nulls the benchmark session's writer on every non-zero rank ([`benches/conftest.py:283-290`](https://github.com/Algorithmiq/monoprop/blob/main/benches/conftest.py#L283-L290)):

```python
# Nulling the session handle skips the write; closing alone leaves ``with self.json``
# raising, so do both.
bench_session = getattr(config, "_benchmarksession", None)
if bench_session is not None:
    bench_session.json = None
bench_json = getattr(config.option, "benchmark_json", None)
if bench_json is not None:
    if hasattr(bench_json, "close"):
        bench_json.close()
    config.option.benchmark_json = None
```

`pytest_sessionfinish` returns early off rank 0 for `results/<label>.json` too, and the module docstring states the invariant: *"Every rank runs the whole session, so rank 0 alone prints, writes and records."*

On the concurrent-open half of the concern: pytest-benchmark declares `--benchmark-json` as `type=Path`, not `argparse.FileType`, so no rank opens or truncates the path at parse time either — I checked the installed plugin source rather than assuming. There is exactly one writer and one open.

Leaving this as-is. If you spotted a path where a non-zero rank still reaches the writer, point at it and I'll look again.

README.md
robertodr: Regardless of `vars.BENCH_BARE_METAL` (which is already set to true for this repo) we want to have the badge, if it says "no status" it's very visible and we need to fix it.

robertodr:
### Reply to [this comment](https://github.com/Algorithmiq/monoprop/pull/323#pullrequestreview-5131162518)

I would like that `justfile` only contains recipes that are **generally** useful in CI and locally. If a recipe is only ever used/meaningful in CI (_e.g._ the ones that mention bencher) then it _must_ not be in the `justfile` and has to be hoisted in a workflow. If the `prek` hook fails, add a tight exception to the allow-list in the local hook.

### Reply to [this comment](https://github.com/Algorithmiq/monoprop/pull/323#issuecomment-5569619655)

`pytest-benchmark` itself depends on `py-cpuinfo2` as of [version 5.3.0](https://github.com/ionelmc/pytest-benchmark/blob/v5.3.0/pyproject.toml#L39), [because `py-cpuinfo` has not seen a release in 4 years.](https://github.com/ionelmc/pytest-benchmark/pull/308). I suggest you bump the dependency on `pytest-benchmark` (by running `uv lock --upgrade`) as a stacked commit on top of this. I do not think upgrading to `py-cpuinfo2` will affect the recorded series: the author of the fork states that it's basically the same as `py-cpuinfo`. In any  case: the goal of using it is to record more complete info about the machine as _additional_ metadata of a run.

README.md
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Added back in 595b0924, pointed at `bench_bare_metal.yml`.

diagonal-hamiltonian:

Both done.

**Hoist** (fa83f444). All six `bench-ci*` recipes are gone from the `justfile`; `bench.yml` now runs them inline in the three steps that owned them. The `justfile` keeps only `bench`, `bench-mpi`, `bench-build-mpi` and `bench-smoke`. Three tight `(bench.yml, <step name>)` entries added to `ALLOWED` — keyed by step name, so a rename still forces them to be reconsidered.

Two notes:

- `bench-ci-resolve-cores`'s Python body became `.github/workflows/scripts/resolve_cores.py` rather than a heredoc in YAML, next to `check_shape.py` and `cpu_slug.py`. The `has_mpi` guard `bench-ci-mpi` ran per rung now runs once there.
- `benchmarks.mdx` advertised `just bench-ci` / `bench-bmf` / `bench-ci-mpi` as the local reproduction path, so that section now points at `just bench` / `just bench-mpi` with the rung's shape and flags.

The rung loop moved verbatim. Re-ran the parser cases against the extracted `run:` block under bash 5.2: 16/16 identical — real block, rung-after-MPI (fd 3), unbalanced quote, quotes-only, missing/non-numeric rounds, oversubscribed, `per-rank`→0, no args, comments/blanks, and `0`/`08` in each of the three fields. `prek run --from-ref main --to-ref HEAD` passes, `check-workflow-commands` included.

**`py-cpuinfo2`** (cfa07624), its own commit: `uv lock --upgrade-package pytest-benchmark`, 5.2.3 → 5.3.0, which swaps `py-cpuinfo` 9.0.0 for `py-cpuinfo2` 10.1.1. `uv.lock` only — the existing `pytest-benchmark>=5` already admitted it. The fork keeps the `cpuinfo` module and the `machine_info.cpu.brand_raw` key, verified end to end rather than assumed: `just bench-smoke` → `Apple M5` → `cpu_slug.py` → `apple-m5`. `pytest packages/monoprop-bench-tools/tests`: 35 passed, 5 skipped (non-Linux).

robertodr (approved):

PR #325, opened 2026-09-01
fix!: 🐛 fix TermIndex at 32 bits and check the append against the ceiling

diagonal-hamiltonian:

## Summary

`monoprop_WIDE_TERM_INDEX` widened `TermIndex` from `uint32_t` to `uint64_t` so that a
single partition could index more than 2^32 terms. That partition is not reachable in
practice: 2^32 terms is a few hundred GiB of resident operator **inside one partition**,
and the answer to "too many terms per partition" is more partitions or more MPI ranks,
both of which divide the terms per partition. No wheel ever shipped the switch, and the
default build — the one everything is released and measured on — was already 32-bit, so
this removes a configuration, not a behaviour.

What the switch did cost is real: a second type width through the operator index, the
graph-encoding records and the cross-rank storage; three `#if`-guarded test bodies that
only one configuration ever compiled; a dedicated `just` recipe; two CI lanes; and five
documentation sites. Two error messages also told the user to "rebuild with
`-Dmonoprop_WIDE_TERM_INDEX`", pointing at a configuration nobody builds instead of at the
partition count that actually fixes the problem.

This PR fixes `TermIndex` at 32 bits, removes the switch everywhere, and turns the ceiling
into a documented hard limit that is checked *before* an append rather than after — see the
second commit, which closes a real hole.

## Changes

- **`cpp/monoprop/TypeAliases.h`** — `TermIndex` is unconditionally `std::uint32_t`, with
  the reason stated where the alias lives.
- **The ceiling is still checked at every narrowing.** `OperatorIndex::check_index_fits`
  and `detail::checked_term_index` are unchanged in behaviour; their messages now say
  "raise the partition or rank count" instead of naming a build flag. Nothing starts
  wrapping silently.
- **`grow_rows_geometric()` now refuses an append that would pass the ceiling, before it
  grows** (second commit, `fix:`). That function is the single point every append into a
  partition's term store passes through — `push_back` via `MPOperator::append_term`, and the
  three bulk sites in `Resolve.h`, `Engine.h` and `MPOperator.h` — and it had no ceiling
  check of its own. The only guard was `check_index_fits()` at hash-insert time, i.e. *after*
  the append, so `push_back` could take a partition past 2^32 terms unnoticed, and a bulk
  append that did trip the guard had already grown `rows_` and moved `size_` to a count whose
  last index is unrepresentable, with no way to unwind it. The check names both counts, and is
  written `n > kIndexCeiling - base` rather than `base + n > kIndexCeiling` because the sum is
  what would wrap. An empty append stays legal at every size, including at the ceiling itself:
  the last valid index is `kIndexCeiling - 1`, so a store of exactly `kIndexCeiling` terms is
  full, not over.
- **Tests keep the coverage that survives.** The `#if`-guarded narrow arms are now
  unconditional: the ceiling round-trip and its `BOOST_CHECK_THROW`
  (`graph_encoding_tests`), and a `static_assert(sizeof(TermIndex) == 4)` that every lane
  compiles (`operator_index_tests`) — that assert is what replaces the deleted CI lane. The
  occupied-slot record's rule is now written out as the fixed 12 bytes it has always been
  in the shipping build (`large_cosine_storage_tests`). Only the u64-above-2^32 round-trip
  is deleted, because it asserted behaviour that no longer exists.
- **Build**: the CMake `option()` and its status line, both `target_compile_definitions`
  entries, and the `[[tool.scikit-build.overrides]]` block keyed on the environment switch.
  The MPI override is untouched.
- **CI**: `test.yml` loses the `wide` matrix axis, its two `wide: "on"` lanes, the job env,
  the cache suffix and the test label (19 lanes → 17). The `mpi` axis is unchanged.
- **Docs**: `just test-wide` and the switch are gone from `README.md`, `building.mdx`,
  `testing.mdx` and the `justfile`; `AGENTS.md` records that the width is fixed and why.
  `features/parallelism.mdx` gains the user-facing statement of the limit — one partition
  caps at ~4.3e9 terms, crossing it raises a `RuntimeError`, and the fix is more partitions
  or ranks, or a larger `lower_atol`.

## Verification

Built with MPI on (GCC 14.3.0 / OpenMPI 5.0.8, `monoprop_MAX_NUM_MODES=250`) and run
serially:

```
$ ctest --test-dir build/.../cpp/tests -L serial
100% tests passed, 0 tests failed out of 271     # 269 + the two new ceiling cases

$ monoprop_PARTITIONS=4 python -m pytest -m "not mpi" -q
622 passed, 8 deselected in 44.32s
```

The two new cases declare the boundary rather than materialising it — 2^32 rows is hundreds of
GiB — so they assert on the refusal path, which returns before a single row is allocated, and on
`size()` being unchanged afterwards. One of them also pins `SIZE_MAX` being refused rather than
wrapping.

Note for reviewers: like the pre-existing `TermIndexCeilingReached`, this throw is rank-local, and
monoprop `MPI_Abort`s on a rank-local error — so under MPI it ends the job rather than unwinding
cleanly. That is unchanged in kind by this PR; the gain is that the operator is left intact and the
message says what to do.

`clang-format` is clean on all nine changed C++ files.

`grep -rn WIDE_TERM_INDEX` over the tree returns nothing, and the switch is absent from the
generated `compile_commands.json`.

## Checklist

- [ ] `CHANGELOG` / release notes updated if applicable — no `CHANGELOG` in the repo; the
      `BREAKING CHANGE:` footer carries the release note.

## AI/LLM disclosure


cpp/include/monoprop/MPFunctions.h
```diff
@@ -68,8 +68,7 @@ class monoprop_EXPORT EvalState {
 private:
     size_t length_ = 0;
     bool is_dense_ = false;
-    /// Widened from TermIndex, whose width is a build knob (monoprop_WIDE_TERM_INDEX) that has no
-    /// business in an exported signature.
+    /// Widened from TermIndex: a 32-bit storage width has no business in an exported signature.
```
```suggestion
```
don't even know what this means...

cpp/include/monoprop/MPFunctions.h
robertodr, line 1: can you silence the clang-tidy warnings on this file?

cpp/monoprop/detail/graph_encoding/MPGraphEncoding.cpp
robertodr: also here, appease clang-tidy please 🙏

AGENTS.md
```diff
@@ -37,6 +37,10 @@ monoprop is a high-performance C++/Python hybrid library implementing Majorana a
 - **Core C++ Engine**: Public headers in `cpp/include/monoprop/` and implementation in `cpp/monoprop/`
 - **Python Interface**: User-facing API in `src/monoprop/` with C++ bindings in `src/monoprop/bindings/`
 - **Template-Based Design**: Heavily templated C++ code with compile-time mode limits (`monoprop_MAX_NUM_MODES`)
+- **`TermIndex` is fixed at 32 bits** (`cpp/monoprop/TypeAliases.h`) and is not a build knob: one
+  partition caps at 2^32 terms, enforced at every narrowing by `OperatorIndex::check_index_fits` and
+  `detail::checked_term_index`. Hundreds of GiB of operator would fit in one partition first, so the
+  answer to the ceiling is more partitions or ranks, not a 64-bit variant.
```
robertodr:
```suggestion
```

robertodr (dismissed):
very good. can you appease clang-tidy on the files you touched? then we're good to merge

cpp/include/monoprop/MPFunctions.h
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Deleted. It only made sense while the width was a build knob — the point used to be "an exported
signature must not depend on `monoprop_WIDE_TERM_INDEX`", and with `TermIndex` fixed there is nothing
left for it to say. The note on `values_` keeps the `rows_`/`values_` pairing documented.

cpp/include/monoprop/MPFunctions.h
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Done — `rows_` and `values_` no longer carry a redundant `= {}` (`readability-redundant-member-init`,
lines 72 and 74). Those two were the only warnings the clang-tidy job reported for this file.

The clang-tidy job on 62dd61e reports 0 annotations, down from 10 across 4 sites.

cpp/monoprop/detail/graph_encoding/MPGraphEncoding.cpp
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Done — `checked_world_slot` is now `static` (`misc-use-internal-linkage`). It is the one function in
this TU with no declaration in `MPGraphEncodingStorage.h`, so nothing outside could have been using
it; `checked_mpi_int` and `checked_term_index` keep external linkage because that header exports them.

While here I also fixed the fourth site the job flagged in a file this PR touches:
`MPGraphEncodingTypes.h:121` now parenthesises `2 * sizeof(TermIndex)`
(`readability-math-missing-parentheses`).

The clang-tidy job on 62dd61e reports 0 annotations, down from 10 across 4 sites.

AGENTS.md
diagonal-hamiltonian: :robot: _AI text below_ :robot:

Dropped. Worth knowing where the fact still lives: `docs/content/docs/features/parallelism.mdx` states
the per-partition ceiling for users, so it is written down somewhere — say the word if you would rather
it were not there either.

diagonal-hamiltonian:

All four threads addressed in 62dd61e.

**clang-tidy: 0 annotations on 62dd61e**, down from 10 across 4 sites. All four were pre-existing —
the QA job's line filter spans every line of a changed file, so touching these files is what surfaced
them:

| site | check | fix |
| --- | --- | --- |
| `MPFunctions.h:72,74` | `readability-redundant-member-init` | dropped the redundant `= {}` on `rows_` and `values_` |
| `MPGraphEncoding.cpp:38` | `misc-use-internal-linkage` | `checked_world_slot` is now `static` — the one function in that TU with no declaration in `MPGraphEncodingStorage.h`, so nothing outside could have used it |
| `MPGraphEncodingTypes.h:121` | `readability-math-missing-parentheses` | parenthesised `2 * sizeof(TermIndex)` |

Also dropped the `EvalState::rows_` comment and the `AGENTS.md` bullet, per the two suggestions. The
per-partition ceiling is still stated for users in `docs/content/docs/features/parallelism.mdx`.

Re-verified on the target machine (GCC 14.3.0 / OpenMPI 5.0.8, MPI on): **271/271 serial CTest cases
passed**, and `clang-format` is clean on all nine changed C++ files.

robertodr (approved):

PR #327, opened 2026-09-01
fix(bench-tools): 🐛 fold the peak-RSS mark into enclosing windows

diagonal-hamiltonian:

## Summary

`mm->hiwater_rss` is a single per-process field. `HighWaterMark.__enter__` resets it (via
`/proc/self/clear_refs`), so an inner window silently erased the peak of every window enclosing it.

The benchmark suite nests exactly that way. `OpMemory.open()` is called inside pedantic `setup`,
*after* the propagator has been constructed, while the autouse `record_memory` window is already
open — and that fixture documents itself as spanning `setup`, "so it predicts an OOM kill". It did
not: every recorded `memhwm` began at `op_memory`'s reset and omitted the construction transient
entirely.

That is not a cosmetic gap. On the Schrödinger random rung at 128 partitions this harness reported
**3.6 GiB** for a run whose true peak was **79.8 GiB** — a 22× understatement, and the reason the
blow-up fixed in the follow-up PR went unnoticed. `reset_peak_rss`'s own docstring already warned
that a window reset costs you `ru_maxrss` as a ceiling; the same argument applies to a Python
window one frame up, which is what this fixes.

## Changes

- `reset_peak_rss()` folds the current mark into every open window (`max`) before discarding it,
  through a weak registry of windows between `__enter__` and `__exit__`.
- `HighWaterMark.__exit__` takes `max(self.peak_bytes, observed)` rather than overwriting, so a fold
  from an inner window survives.
- Docstrings: the nesting contract on `HighWaterMark` and `reset_peak_rss`, and `record_memory`'s
  claim to span `setup` now says what it depends on.
- Regression test: an 80 MiB transient that ends *before* an inner window opens must still appear in
  the outer window's delta. It fails on `main` (reports 8 MiB) and passes here.

## Verification

- `packages/monoprop-bench-tools/tests/test_memory.py` — 5 passed; the new case verified to fail
  against `main`'s `cpu.py` and pass against this one.
- Full Python suite: **624 passed, 8 skipped** (skips are the `--with-mpi` cases).
- `uvx prek run --from-ref main --to-ref HEAD` — all hooks green.

## Checklist

## AI/LLM disclosure

### DIFF
--- a/benches/conftest.py
+++ b/benches/conftest.py
@@ -445,9 +445,12 @@ def record_memory(request: pytest.FixtureRequest, bench_comm: Any) -> Iterator[N
     """Record ``memhwm`` (summed peak RSS) and ``memhwm_max`` (the worst rank's peak RSS).
 
     It spans ``setup``, so it predicts an OOM kill but is the wrong number for comparing
-    operations -- ``opmemdelta`` is that. The sum alone has inverted per-rank readings here
-    before, so ``memhwm_max`` is recorded alongside it, never in place of it. Both reduces
-    are collective; only rank 0 records.
+    operations -- ``opmemdelta`` is that. Spanning ``setup`` relies on windows nesting: the
+    ``op_memory`` window opens inside ``setup``, after construction, and resets the same
+    per-process kernel field, so without the fold in ``reset_peak_rss`` this figure would
+    silently start at that reset and omit the construction transient entirely. The sum alone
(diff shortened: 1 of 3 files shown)


Panadestein (dismissed):
Thanks @diagonal-hamiltonian, this is an important patch.

Panadestein (approved):

PR #328, opened 2026-09-01
fix(cpp): 🐛 build only world-slot-local Schrodinger paired basis

diagonal-hamiltonian:

## Summary

In the Schrödinger picture, `MonomialPropagator`'s constructor built the **entire global** paired
monomial basis into a dense `MonomialList` and only then hash-filtered it down to the slot's `1/P`
share:

```cpp
auto op = schrodinger_ ? generate_paired_op<NumModes>(sc / 2 + sc % 2, logical_num_modes_)
                       : local_heisenberg_terms;
...
if (my_rank == find_rank<NumModes>(mono, num_ranks)) { /* keep 1/P of it */ }
```

`PartitionGroup` constructs its S children on their own master threads and waits for all of them, and
the child's `num_ranks` is the flat `P = R·S` world — so **all P full copies of the global basis were
live at the same instant**. Peak was `P × Σ_{k≤K} C(n,k) × sizeof(Monomial)` for
`K = ⌈schrodinger_cutoff/2⌉`.

The Heisenberg arm never had this problem, and the fix is to make Schrödinger behave the same way:
`local_heisenberg_terms` is filtered *as it is built*, so it is O(local). Now both are.

`for_each_paired_monomial` hands the walk to a callback, so each slot keeps only what it owns and
nothing ever materializes the global set. The emission order is moved verbatim, so each slot's kept
subsequence — and therefore every `i++` row index, and therefore every result — is unchanged.

## Changes

- `MajoranaAlgebra.h`: `for_each_paired_monomial` (callback walk) and `paired_op_size` (closed form,
  saturating rather than wrapping). `generate_paired_op` becomes a thin wrapper over the walk, for
  callers that want random access over a basis small enough to hold.
- `MonomialPropagator.inl`: the Schrödinger arm enumerates and inserts; the Heisenberg arm iterates
  its already-local list. One shared insert lambda, so there is a single place a row index is assigned.
- **Second bug, same lines:** the store `reserve` divided by `num_ranks` on both arms. That is right
  for a global count and a **P-fold under-reserve** for Heisenberg, whose list is already local — at
  R·S = 128 it reserved 1/128 of the rows and then reallocated and rehashed through the entire
  initial insert. Each arm now reserves from its own count.
- **Third:** an over-wide `schrodinger_cutoff` admits `2^num_modes` terms. That used to fail as an
  allocation; with the list gone it would be a walk that never finishes, so the basis size is checked
  up front and the error names `schrodinger_cutoff` rather than the ceiling it happens to trip. The
  check has two arms, and both are needed: the count not fitting `size_t` at all (rank-independent,
  because every slot walks the whole global basis however the rows are divided) and the slot's share
  of rows exceeding what a term index can address. Only the first survives a `wide:on` build with
  several partitions, which is why it is there.
- Tests: `partition_equivalence_tests.cpp` had **no** Schrödinger case at all — every case passed
  `std::nullopt` for `schrodinger_cutoff`. It has two now, one of which reads the constructor's output
  with no graph in between. Plus a literal order pin on the walk, `paired_op_size` ≡ emitted count over
  every `n × k` including the clamped over-ask, and a disjoint/covering/subsequence check on the
  hash filter in `mpi_utils_tests.cpp`.

## Results

One 128-core node, 142 modes unless stated, `monoprop_PARTITIONS` explicit, arms interleaved with the
order flipped per rep, 3 reps. Baseline arm = this branch's parent (`_core.so`
`c27cb88371213b820c52adb2c1b6ca99`); fixed arm = this branch (`42f8cdc5add3501bc109f8b9b1004a4c`).
Every number below was reproduced on the final binary after the guard was tightened.

Peak RSS vs partition count, at a fixed problem — the mechanism, directly:

| S | before (GiB) | after (GiB) |
| --- | --- | --- |
| 1 | 2.93 – 2.94 | 2.99 – 3.00 |
| 16 | 11.13 – 11.24 | 3.19 |
| 64 | 41.56 – 41.62 | 3.45 – 3.46 |
| 128 | 79.32 – 80.26 | 3.39 – 3.40 |

Before: **0.61 GiB per partition** (predicted 668.5 MB = 16,711,839 monomials × 40 B — the basis size
`paired_op_size` computes, and the ledger's own `d_state_coeffs_nonzero`). After: flat across a 128×
range, **23.5×** lower at S=128, 12/12 cells in the same direction. All 24 cells agree on `terms` and
on `opmembreak.total_bytes`, so this was pure waste, not a trade.

Two configurations that were `SIGKILL` on a 223 GiB node now simply run:

| case | before | after |
| --- | --- | --- |
| `--cutoff=7` (basis 464,784,177) | `rc=137` | 70.0 GiB, 466.9M terms |
| `--num-modes=180` (basis 43,268,956) | `rc=137` | 8.6 GiB, 45.6M terms |
| `--num-modes=160` | 127.5 GiB | 5.6 GiB |
| `--num-generators=4000` | 80.2 GiB | 14.8 GiB |

These were previously believed to be properties of the problem. They were this bug: one copy of the
`--cutoff=7` basis is 18.6 GB, which fits a node easily; 128 copies is 2.4 TB.

`schrodinger_cutoff` remains superexponential in its own right — 8 → 16.7M initial terms, 10 → 465M at
142 modes — which is now documented rather than discovered.

## Verification

- CTest: **278/278** and **276/276** passed, 0 failed (labels `cxx`/`mpi`/`mpi-2`/`serial`/`unit`),
  including the hybrid `R>1, S>1` partition-equivalence case and the Schrödinger MPI fresh-insert
  equivalence case. All 6 new cases confirmed present and passing by name.
- Python suite on compute nodes across four layouts — `R×S` = 2×1, 2×16, 4×8 and **16×16 (world
  256)**: **601 passed** in each. The 16×16 layout matters here because the over-wide-cutoff check
  reads `num_ranks` as the flat `R·S` off the child communicator, which only a `HybridComm` provides.
- `uvx prek run --from-ref main --to-ref HEAD` — all hooks green.

## Checklist

## AI/LLM disclosure

### DIFF
--- a/cpp/monoprop/algebra/MajoranaAlgebra.h
+++ b/cpp/monoprop/algebra/MajoranaAlgebra.h
@@ -19,6 +19,7 @@
 #include <bit>
 #include <complex>
 #include <cstdint>
+#include <limits>
 #include <optional>
 #include <stdexcept>
 #include <vector>
@@ -137,10 +138,39 @@ auto monomial_from_selector(const std::vector<bool> &selector, size_t inactive_m
     return current;
 }
 
-// All fully paired Majorana monomials with up to max_ones pairs, over the active logical modes only.
-template <size_t NumModes>
-auto generate_paired_op(size_t max_ones, size_t logical_num_modes) -> MonomialList<NumModes> {
-    MonomialList<NumModes> combinations;
+// How many monomials for_each_paired_monomial emits: sum over k in [0, min(max_ones, n)] of C(n, k).
+// Saturates at SIZE_MAX instead of wrapping — a caller sizing an allocation or rejecting an over-wide
+// cutoff must never be handed a modular residue that reads as small.
+inline auto paired_op_size(size_t max_ones, size_t logical_num_modes) -> size_t {
+    constexpr auto saturated = std::numeric_limits<size_t>::max();
+    max_ones = std::min(max_ones, logical_num_modes);
+    size_t total = 1;
+    size_t binomial = 1; // C(n, k) carried across iterations
+    for (size_t k = 1; k <= max_ones; ++k) {
+        const size_t factor = logical_num_modes - k + 1;
+        if (binomial > saturated / factor) {
+            return saturated;
+        }
+        // Multiply before dividing: C(n,k-1)*(n-k+1) is divisible by k, so this stays exact.
+        binomial = binomial * factor / k;
+        if (total > saturated - binomial) {
+            return saturated;
+        }
+        total += binomial;
+    }
+    return total;
+}
+
+// All fully paired Majorana monomials with up to max_ones pairs, over the active logical modes only,
+// handed to f one at a time. Never materialized: the set is up to 2^logical_num_modes wide, and a
+// distributed build runs one of these walks per world slot concurrently.
+//
+// The emission order is a contract. Ascending pair count, then, within a pair count, ascending
+// lexicographic order of the selected modes' index tuple. A distributed propagator turns this sequence
+// into row indices by keeping the entries it owns and numbering them 0, 1, 2, ..., so reordering the walk
+// renumbers every stored row and stops two runs at one (rank, partition) count from being bit-identical.
+template <size_t NumModes, typename F>
+auto for_each_paired_monomial(size_t max_ones, size_t logical_num_modes, F &&f) -> void {
     // Clamp in pairs, not bits: max_ones counts pairs and bounds the fill over `selector`, one slot per mode.
     max_ones = std::min(max_ones, logical_num_modes);
     auto selector = std::vector(logical_num_modes, false);
@@ -150,13 +180,24 @@ auto generate_paired_op(size_t max_ones, size_t logical_num_modes) -> MonomialLi
         std::fill(selector.begin(), selector.begin() + num_ones, true);
 
         do {
-            combinations.push_back(monomial_from_selector<NumModes>(selector, inactive_mode_prefix));
+            f(monomial_from_selector<NumModes>(selector, inactive_mode_prefix));
         }
         while (std::ranges::prev_permutation(selector).found);
 
         std::ranges::fill(selector, false);
     }
+}
 
+// The materialized form of for_each_paired_monomial, for callers needing random access over a basis small
+// enough to hold. Not for a distributed construction path: this is the whole global basis, so every world
+// slot would hold a copy of all of it.
+template <size_t NumModes>
+auto generate_paired_op(size_t max_ones, size_t logical_num_modes) -> MonomialList<NumModes> {
+    MonomialList<NumModes> combinations;
+    combinations.reserve(paired_op_size(max_ones, logical_num_modes));
+    for_each_paired_monomial<NumModes>(max_ones, logical_num_modes, [&](const Monomial<NumModes> &mono) {
+        combinations.push_back(mono);
+    });
     return combinations;
 }
 
--- a/cpp/monoprop/algebra/MajoranaAlgebra.h
+++ b/cpp/monoprop/algebra/MajoranaAlgebra.h
@@ -139,8 +139,7 @@ auto monomial_from_selector(const std::vector<bool> &selector, size_t inactive_m
 }
 
 // How many monomials for_each_paired_monomial emits: sum over k in [0, min(max_ones, n)] of C(n, k).
-// Saturates at SIZE_MAX instead of wrapping — a caller sizing an allocation or rejecting an over-wide
-// cutoff must never be handed a modular residue that reads as small.
+// Saturates at SIZE_MAX instead of wrapping.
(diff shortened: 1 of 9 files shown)


cpp/monoprop/detail/monomial_propagator/MonomialPropagator.inl
```diff
@@ -24,8 +24,11 @@
 #include <cstdlib>
 #include <cstring>
 #include <functional>
+#include <format>
```
robertodr:
```suggestion
#include <format>
#include <functional>
```

cpp/monoprop/algebra/MajoranaAlgebra.h
```diff
@@ -137,10 +138,39 @@ auto monomial_from_selector(const std::vector<bool> &selector, size_t inactive_m
     return current;
 }
 
-// All fully paired Majorana monomials with up to max_ones pairs, over the active logical modes only.
-template <size_t NumModes>
-auto generate_paired_op(size_t max_ones, size_t logical_num_modes) -> MonomialList<NumModes> {
-    MonomialList<NumModes> combinations;
+// How many monomials for_each_paired_monomial emits: sum over k in [0, min(max_ones, n)] of C(n, k).
+// Saturates at SIZE_MAX instead of wrapping — a caller sizing an allocation or rejecting an over-wide
+// cutoff must never be handed a modular residue that reads as small.
```
robertodr:
```suggestion
// Saturates at SIZE_MAX instead of wrapping.
```

cpp/monoprop/detail/monomial_propagator/MonomialPropagator.inl
```diff
@@ -161,11 +164,40 @@ MonomialPropagator<NumModes>::MonomialPropagator(const OperatorDict &initial_ope
...
-    auto sc = schrodinger_cutoff.value_or(cutoff + 2);
-    sc = std::min(sc, static_cast<unsigned int>(2 * logical_num_modes_));
-    auto op = schrodinger_ ? generate_paired_op<NumModes>(sc / 2 + sc % 2, logical_num_modes_) : local_heisenberg_terms;
+    // Schrodinger's initial rows are the global paired basis, hash-partitioned over the P = R*S world
+    // slots; Heisenberg's local_heisenberg_terms was already filtered to this slot's share above. Hence the
+    // two reserves: a global count needs its 1/P share, a local one already is the share.
+    size_t max_pairs = 0;
+    size_t expected_local_terms = std::max<size_t>(1, local_heisenberg_terms.size());
+    if (schrodinger_) {
+        // schrodinger_ is schrodinger_cutoff.has_value(), set in the member-init list, so the deref holds.
```
robertodr:
```suggestion
```

cpp/monoprop/detail/monomial_propagator/MonomialPropagator.inl
```diff
@@ -161,11 +164,40 @@ MonomialPropagator<NumModes>::MonomialPropagator(const OperatorDict &initial_ope
...
+        max_pairs = sc / 2 + sc % 2;
+        const size_t global_terms = paired_op_size(max_pairs, logical_num_modes_);
+        // paired_op_size saturates rather than wrapping, so this is "too large to count", not a size.
+        const bool uncountable = global_terms == std::numeric_limits<size_t>::max();
+        const size_t share = global_terms / std::max<size_t>(1, num_ranks);
+        // A cutoff near the mode count makes the basis 2^logical_num_modes. Two independent ways that
+        // is hopeless, and both are needed: the count does not fit size_t at all, which is
+        // rank-independent because every slot walks the whole global basis however the rows are
+        // divided; or the slot's share of rows cannot be addressed by a term index. Rejected here by
+        // name rather than as a walk that does not finish.
```
robertodr:
```suggestion
```

cpp/monoprop/detail/monomial_propagator/MonomialPropagator.inl
```diff
@@ -174,13 +206,23 @@ MonomialPropagator<NumModes>::MonomialPropagator(const OperatorDict &initial_ope
...
+    // index is a position in the kept subsequence, so the enumeration order below is load-bearing.
+    const auto keep_if_owned = [&](const Monomial<NumModes> &mono) {
         if (my_rank == find_rank<NumModes>(mono, num_ranks)) {
             mp_op_.append_term(mono);
             mp_op_.store->emplace(mono, i++);
         }
+    };
+    if (schrodinger_) {
+        // Enumerated, never listed: PartitionGroup constructs the S partitions concurrently on their own
+        // masters, so a list here would hold P full copies of the global basis at one instant.
```
robertodr:
```suggestion
        // PartitionGroup constructs the S partitions concurrently on their own
        // masters, so a list here would hold P full copies of the global basis at one instant.
```

robertodr (approved):

PR #331, opened 2026-09-02
fix: :bug: cstddef include before boost/test

MGDelcey:
It seems that using modern compilers with cleaned up transitive includes creates issues when including boost/test due to missing std::size_t definition. This PR fixes this by including cstddef before the boost/test include. This would hopefully be unnecessary when boost fix their includes.

### DIFF
--- a/cpp/tests/build_graph_tests.cpp
+++ b/cpp/tests/build_graph_tests.cpp
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <cstddef>
 #include <boost/test/data/monomorphic.hpp>
 #include <boost/test/data/test_case.hpp>
 #include <boost/test/unit_test.hpp>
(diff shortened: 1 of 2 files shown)


cpp/tests/build_graph_tests.cpp
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <cstddef>
```
robertodr, line 15:
```suggestion
#include <cstddef>

```

cpp/tests/mpfunctions.cpp
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <cstddef>
```
robertodr, line 15:
```suggestion
#include <cstddef>

```

robertodr (approved):
Thanks

PR #337, opened 2026-09-03
feat(propagator): ✨ read individual evolved coefficients by term

JoshCudbyAlgo:

## Summary

`evolved_operator` decodes the whole evolved operator: it enumerates the index, materialises a key per surviving term, and builds a term dictionary. `evolved_operator_coefficients` probes the index with the terms the caller names instead, so the decode costs one entry per term *requested* rather than one per term the operator carries. Reading a few amplitudes out of an evolved state (Schrödinger picture) is the motivating case. The graph contraction is unchanged and still dominant — this removes the per-surviving-term tail, not evolution cost.

Contracts, all pinned by tests: an absent term reads back as `0`; there is deliberately no `atol`, since the caller named its terms and magnitude filtering would silently zero some of them (threshold the result instead); the result is positional, and a repeated term is answered once per occurrence; the empty term agrees with `evolved_operator` in both pictures — `core_term()` in Heisenberg, an ordinary index row in Schrödinger; and it is rank-local, exactly as `evolved_operator` is.

Terms must be canonical, as the constructor's `initial_operator` keys must be. The engine keys terms by an order-insensitive bitset, so a raw `(1, 0)` would resolve to the row of `(0, 1)` and read back its coefficient without the anticommutation sign — which a lookup has no coefficient of its own to carry. Raw sequences are therefore validated through `Majorana` rather than normalized; use `Majorana.from_unsorted` and apply the sign it returns.

## Changes

- `MonomialPropagator::evolved_operator_coefficients(parameters, terms)`, built on the operator index's group-prefetch `find_batch` rather than a `for_each` enumeration. Under partitions each partition probes on its own master thread into its own vector and the merge runs on the facade thread, as every other partition fan-out in the file does.
- Bound as `evolved_operator_coefficients(terms, parameters=None)`, returning a complex NumPy array.
- A `_term_slots` front-end hook: Majorana indices for `MajoranaPropagator` (`Majorana` terms, index tuples or index arrays, with non-canonical ones rejected), symplectic slots for `PauliPropagator` (`Pauli` terms). Cross-front-end terms raise `TypeError`. It is a default rather than an abstract method, so a front-end that does not encode terms still constructs.
- 11 C++ and 17 Python cases oracled against `evolved_operator(atol=0.0)`, covering both pictures crossed with 1/2/4 partitions. Partitioned Schrödinger is the only configuration where the identity is a hash-partitioned index row, so it is what exercises the empty key against the concurrent merge.
- `test_evolved_operator_coefficients_is_rank_local`: the ranks' answers sum to the serial result, plus a check that the split is real — counted against the terms each rank *owns* rather than its nonzero coefficients, which would drift if an owned term evolved to exactly zero.
- A "Reading individual terms" section in `docs/content/docs/features/evaluation.mdx`.

## Verification

274/274 C++ `ctest` and 611 Python tests pass locally, and `prek run --all-files` is clean over this branch's files. **The MPI leg was never executed**: the dev environment is a serial build, so `test_evolved_operator_coefficients_is_rank_local` is reviewed but unrun and CI's `[mpi:on]` lanes are its first real run. Branched from `7101b33`, which `main` has since moved past, so a rebase before merge would be sensible.

## Checklist

- [ ] `CHANGELOG` / release notes updated if applicable — N/A, the repository keeps no CHANGELOG

## AI/LLM disclosure

- [x] I used the following tool to help write this PR description: Claude Code (Claude Opus 5)
- [x] I used the following tool to generate or modify code: Claude Code (Claude Opus 5)

## Summary by CodeRabbit

- **New Features**
  - Added `evolved_operator_coefficients` to retrieve selected evolved-operator coefficients without enumerating the full operator.
  - Supports Majorana, Pauli, and raw index terms, preserving query order and repeated terms.
  - Returns zero for absent terms and supports identity terms, Schrödinger and Heisenberg modes, and rank-local execution.

- **Documentation**
  - Added guidance covering supported term formats and query behavior.


cpp/monoprop/detail/monomial_propagator/MonomialPropagator.inl
```diff
@@ -1169,4 +1169,62 @@ auto MonomialPropagator<NumModes>::evolved_operator_terms(const VecD &parameters
...
+        for (const auto &partition_found : map_partitions_(probe)) {
+            for (size_t q = 0; q < out.size(); ++q) {
+                out[q] += partition_found[q];
+            }
+        }
+    }
+
+    // Match evolved_operator_terms(): round off the anti-hermitian numerical noise.
+    for (auto &coeff : out) {
+        coeff = {std::round(coeff.real() * 1e12) / 1e12, std::round(coeff.imag() * 1e12) / 1e12};
```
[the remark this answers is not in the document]
JoshCudbyAlgo: Reject: this behaviour is required to stay identical to evolved_operator_terms.

docs/content/docs/features/evaluation.mdx
```diff
@@ -102,3 +102,27 @@ To read the fully evolved operator as a [MajoranaOperator][] (or [PauliOperator]
...
+```python notest
+# The coefficients of these three terms alone, in this order.
+terms = [Majorana(0, 1), Majorana(2, 3), Majorana(0, 3)]
+coeffs = sim.evolved_operator_coefficients(terms, parameters)
+```
+
+The terms are given in the front-end's own vocabulary: [Majorana][] terms (or
+raw index tuples) for [MajoranaPropagator][]; [Pauli][] terms for
+[PauliPropagator][]. The returned NumPy array carries one coefficient per
+requested term, in the order requested.
```
matteoacrossi:
```suggestion
The terms are given as [Majorana][] terms (or
raw index tuples) for [MajoranaPropagator][] or [Pauli][] terms for
[PauliPropagator][]. The returned NumPy array contains the corresponding coefficients, preserving the order.
```

src/monoprop/monomial_propagator.py
```diff
@@ -533,6 +539,65 @@ def evolved_operator(
...
+        terms: Iterable[OperatorTerm],
+        parameters: ParameterValues = None,
+    ) -> np.ndarray:
+        """Return the coefficients of ``terms`` alone in the evolved operator, in the order given.
+
+        A cheaper
+        [evolved_operator][monoprop.monomial_propagator.MonomialPropagator.evolved_operator] when
+        only a few terms are wanted: the contraction is identical and still dominant, but the index
+        is *probed* with these terms rather than enumerated, so the decode costs one entry per term
+        *requested* rather than one per term the evolved operator carries.
```
matteoacrossi:
```suggestion
        A cheaper alternative to
        [evolved_operator][monoprop.monomial_propagator.MonomialPropagator.evolved_operator] when
        only a few terms are needed. Only those terms are decoded from the evolved operator, avoiding enumerating all its terms.
```

src/monoprop/monomial_propagator.py
```diff
@@ -533,6 +539,65 @@ def evolved_operator(
...
+
+        A term the operator does not carry reads back as ``0``, and there is deliberately no
+        ``atol`` -- the caller named its terms, so magnitude filtering would silently zero some of
+        them. Terms must be canonical; for a Majorana product that is not, use
+        [Majorana.from_unsorted][monoprop.majorana.Majorana.from_unsorted] and apply the sign it
+        returns. A repeated term is answered once per occurrence.
```
matteoacrossi: Remove some of this slop

JoshCudbyAlgo: Deslopped here and elsewhere

PR #340, opened 2026-09-08
fix(ci): 🐛 install MPI for the bare-metal benchmark

diagonal-hamiltonian:

## Summary

The bare-metal benchmark has been red since #323 merged. That PR turned the job into the
three-rung ladder, which builds with `monoprop_ENABLE_MPI=ON` and launches L2b under
`mpiexec -n 4` — but `bench.yml` sets up its environment with only a `cache-suffix`, and
`.github/actions/setup` installs `tools/packages/apt-mpi.txt` only when it is handed
`mpi: on`. So the runner had no `MPI::MPI_CXX` while the build was told to require it, and
the run stopped in *Install package*:

```
--   Package 'mpi-cxx' not found
CMake Error at .../FindPackageHandleStandardArgs.cmake:290 (message):
*** CMake configuration failed
```

Every bare-metal run before #323 was green because the old workflow built without MPI. The
fix is the one input, plus a cache suffix that moves with the configuration: `benchmark` was
filled by those non-MPI builds, and a `monoprop` restored from it is the binary every rung
shares — `resolve_cores.py` would then stop the run one step later with *"monoprop was built
without MPI"*.

Nothing downstream of the build had ever run in CI, so this was validated end to end rather
than by making the configure step pass: `bench_bare_metal.yml` dispatched on this branch,
all three rungs measured, shape-checked and uploaded. Run linked in a comment below.

## Changes

- `.github/workflows/bench.yml`: pass `mpi: "on"` to the setup action, and move the
  `setup-uv` cache suffix to `benchmark-mpi`.

## Checklist

## AI/LLM disclosure

## Summary by CodeRabbit

* **Chores**
  * Updated benchmark build setup to ensure MPI support is enabled.
  * Improved benchmark environment caching for MPI-enabled builds.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -42,7 +42,11 @@ jobs:
       - name: Set up the build environment
         uses: ./.github/actions/setup
         with:
-          cache-suffix: benchmark
+          # A cache filled by a non-MPI build would restore a monoprop without MPI, which every
+          # rung below then shares.
+          cache-suffix: benchmark-mpi
+          # The build below requires MPI::MPI_CXX, and the multi-rank rungs need mpiexec.
+          mpi: "on"
 
       - name: Install package
         env:


diagonal-hamiltonian:

End-to-end validation: [run 34228895756](https://github.com/Algorithmiq/monoprop/actions/runs/34228895756), dispatched on this branch, green in 9m21s.

- `Found MPI_CXX: /usr/lib/x86_64-linux-gnu/openmpi/lib/libmpi.so (found version "3.1")` — the build configures, which is what was failing.
- `variant default`, `cores 8 visible physical of 8`, `BENCH_CORES=8`.
- All three rungs measured and shape-checked: `ranks=1 partitions_env=1`, `ranks=1 partitions_env=8 pinned>=8`, `ranks=4 partitions_env=2 pinned>=2`. No `::error::` emitted.
- `cpu slug: intel-xeon-platinum-8488c`, so the testbeds are `intel-xeon-platinum-8488c-8c-L1` / `-8c-L2a` / `-8c-L2b`, exactly as `docs/content/docs/benchmarks.mdx` documents them.
- Three reports uploaded, each carrying all three measures (`latency`, `peak-memory`, `terms`).
- `terms` matches the documented figures: 19,902,244 at L1, and 167,515,463 at **both** L2a and L2b — the geometry-independence check the `terms` threshold exists to hold.

Peaks landed just under the documented sizing figures (L1 2.45 vs ~2.60 GiB, L2a 18.36 vs ~19.11 GiB, L2b 20.73 vs ~23.71 GiB/node), so the `--obs-terms=2500000` ceiling still has headroom on this node.

This run also auto-created a Bencher branch named `fix/bench-mpi-toolchain` (`BENCHER_BRANCH` is `github.ref_name`); worth archiving after merge. The three testbeds are the ones `main` will reuse, so they should stay.

Panadestein (approved):

PR #346, opened 2026-09-09
fix(propagator)!: 🐛 number an incremental Heisenberg axis like the equivalent circuit

antonnykanen:

## Summary

`build_graph` grew a graph in two opposite directions when extending. Layers were
prepended, which is correct — in the Heisenberg picture, conjugating an already-conjugated
operator by one more gate makes that gate act on the reference state *first*, and
`Circuit.__add__` already acknowledged this ("incremental multi-call building, whose
ordering is picture-dependent"). Parameter indices were appended unconditionally
(`mapping = [self._n_params + m for m in circuit.resolved_mapping]`).

So `build_graph(a); build_graph(b)` built the circuit `b + a` but numbered it as if it had
built `a + b`: parameter 0 drove the gate that comes last. Nothing raised. The expectation
value stayed plausible and simply belonged to a different state. Schrödinger appends both
layers and indices, so it was already consistent.

The invariant that should hold, where `==` means the same parameter vector yields the same
expectation value:

```
build_graph(a); build_graph(b)  ==  build_graph(b + a)   # Heisenberg
build_graph(a); build_graph(b)  ==  build_graph(a + b)   # Schrödinger
```

The graph satisfied this in both pictures; the axis only in Schrödinger. It now holds for
both, for the axis, the value **and** the gradient.

Two details worth flagging in review:

- The renumbering is a **cyclic rotation** of the index values, not a shift. A shift would
  leave index 0 unused mid-way and `validate_parameter_mapping` rejects a non-contiguous
  mapping — which is also why no caller-side workaround can place the index directly, and
  the fix has to live here.
- `seed_parameters` is now read on the post-call axis and un-rotated before it reaches the
  engine, which still reads it against the axis the graph has at build time. This is the
  half that needs coordination — see *Breaking change* below.

## Changes

- `monomial_propagator.py`: rotate the per-layer mapping by `(m + num_new) % n_params` after
  a Heisenberg extension, and un-rotate `seed_parameters` before the engine build. Assigned
  via the simulator's per-layer mapping (length `graph_layers`), so the reading is
  unambiguous and `_n_params` stays correct.
- `majorana_propagator.py` / `pauli_propagator.py`: both override `build_graph` with their
  own docstring — the pages users of the public classes actually read — and both still said
  indices "are shifted up onto the accumulated parameter axis automatically". Updated with
  the picture rule and the seed's axis.
- `parameter_mapping`: documented the two API hazards found along the way. The setter's
  docstring never reaches the docs site (griffe renders a property from its getter only), so
  the warning lives on the getter, where it renders as a callout: the per-layer and per-gate
  forms are told apart by length alone but indexed differently (equivalent-circuit order vs
  gate-arrival order), they collide for any all-single-monomial graph, and `list(range(n))`
  is not a neutral reset. Making the two forms structurally unambiguous would mean new API on
  both the Python and C++ surfaces, so it is left out of this PR.
- `MonomialPropagator.h`: note that C++ callers own the numbering — the header takes
  `parameter_mapping` as given, so there is no implicit assignment to fix on that side.
- `evaluation.mdx`: new *Extending a graph across several calls* section, cross-linked from
  `simulation_modes.mdx`.
- `tests/test_circuit.py`: five tests, below.

## Verification

Every new test fails on `main` and passes here — reverted the fix and re-ran to confirm: 6
failures, all Heisenberg legs, Schrödinger legs passing either way.

- `test_incremental_build_wires_the_picture_s_equivalent_circuit` — the invariant per picture
  across all three fixtures, asserted on the axis, the value and the gradient at angles
  *different from the build values*. The existing tests compared energies at the build
  parameters, where the misnumbering is invisible; that is why this shipped. The LiH fixture
  supplies mixed-arity gates, so `n_gates != graph_layers`.
- `test_incremental_heisenberg_moves_each_block_to_the_front_of_the_axis` — blocks of width
  2, 1, 3, so `n_new > 1`. Two guards assert the forward composition and a block reversed
  inside itself both give *different* answers, so the equality cannot pass vacuously.
- `test_incremental_reindex_keeps_a_multi_monomial_gate_on_one_index` — pins `[0, 0, 1, 1, 2]`
  across an extension.
- `test_extend_seed_parameters_are_read_on_the_post_call_axis` — at `lower_atol=0.1` a
  correctly seeded extension reproduces the one-call build's `graph_size`, `size` **and**
  value exactly, while a pre-call-axis seed differs. Holds at every atol probed from 0.05 to
  0.3. My first attempt at this test asserted that the appended block's seed entries are
  unreachable, which is false — they seed the new layers too — and it passed without the fix;
  worth knowing if this area is touched again.
- `test_extend_without_seed_builds_structurally` moved with the contract (`c1 + c2` →
  `c2 + c1`). Note it was passing vacuously: its observable expectation is exactly 0 for
  every gate order.

Beyond the suite: 589 Python tests, 267 C++ tests, `just build-docs` clean, `prek` clean on
the changed files. Checked by hand that an empty extension, an identity gate carrying an
explicit index, and extending after an in-place `contract_partially` all behave, and that the
invariant holds for `PauliPropagator` too (shared engine).

The reporter validated the same approach as a `build_graph` wrapper against Algorithmiq's
`aurora_state` ADAPT loop, which extends a Heisenberg graph one gate per iteration: full
solver suite with its own workaround disabled at 123 passed / 95 skipped (6 failing before
any fix), and machine precision against exact statevector energies on three systems.

## Breaking change

Parameter indices change on every Heisenberg `build_graph` extension, so anything caching an
index or a gradient vector across such a call changes meaning.

The axis half is safe to land alone: `aurora_state`'s workaround (renumber the axis in layer
order) is idempotent, so it becomes a verified no-op and nothing breaks if it stays. **The
seed half is not.** `aurora_state` currently rotates `seed_parameters` itself to compensate,
and with the upstream rotation added it would double-rotate silently — it only affects
`lower_atol`/`upper_atol` runs, so it does not raise and no test catches it; it just feeds
coefficient-informed truncation the wrong coefficients. That rotation has to come out in the
same coordinated bump (`aurora_state` pins `monoprop==0.9.1a0`).

Fixing only the axis would avoid the coordination but leave an axis in natural order with a
seed in append order — a worse contract. This PR takes the coordinated option, as recommended
in the report.

If your release process owns `VERSION` bumps (history shows them as standalone `chore:`
commits), the `VERSION` line is the part to drop from this PR.

## Checklist

- [x] `CHANGELOG` / release notes updated if applicable 

## AI/LLM disclosure

## Summary by CodeRabbit

- **New Features**
  - Incremental graph construction now remaps existing parameters and seeded values when extending graphs in the Heisenberg picture.
  - Parameter ordering and circuit composition are handled consistently across Heisenberg and Schrödinger simulation modes.

- **Documentation**
  - Expanded guidance on extending graphs across multiple calls, parameter-axis ordering, gate numbering, seeded truncation, gradients, and accumulated circuit behavior.
  - Added cross-references to related expectation-value and gradient documentation.


VERSION
```diff
@@ -1 +1 @@
-0.9.1a0
+0.9.1a1
```
robertodr: hyperactive bot... this bump is wrong

docs/content/docs/features/evaluation.mdx
```diff
@@ -13,7 +13,42 @@ parameter-index order (`values[i]` is the angle for gates carrying `param == i`,
...
-`MajoranaPropagator.build_graph` call.
+`MajoranaPropagator.build_graph` call — that keeps the gate order explicit rather
+than leaving it to the picture.
+
+## Extending a graph across several calls
+
+[build_graph][monoprop.monomial_propagator.MonomialPropagator.build_graph] can also grow
+one graph a block of gates at a time, as an ADAPT-style loop does. Which circuit that adds
+up to is decided by the [simulation mode](/concepts/simulation_modes), because the
+Heisenberg picture consumes each call's gates back-to-front:
```
adamglos92:
```suggestion
[build_graph][monoprop.monomial_propagator.MonomialPropagator.build_graph] can also grow
one graph a block of gates at a time. Which circuit that adds
up to is decided by the [simulation mode](/concepts/simulation_modes), because the
Heisenberg picture consumes each call's gates back-to-front:
```

tests/test_circuit.py
adamglos92, line 1: Would be great if you could also add similar tests for PauliPropagator

tests/test_circuit.py
```diff
@@ -765,6 +765,214 @@ def test_build_graph_in_two_calls_schrodinger(fixture: str) -> None:
...
+    assert signature(seeded) == signature(single)
+
+    stale = _incremental_propagator(lower_atol=0.1)
+    stale.build_graph(existing, seed_parameters=existing.parameters)
+    stale.build_graph(
+        appended,
+        seed_parameters=list(existing.parameters) + list(appended.parameters),
+    )
+    assert signature(stale) != signature(single)
+
```
adamglos92: suggestion: if we are testing intended behaviour and we know the test is not flaky, no point of testing it is not flaky

tests/test_circuit.py
```diff
@@ -765,6 +765,214 @@ def test_build_graph_in_two_calls_schrodinger(fixture: str) -> None:
...
+    within_block.build_graph(
+        _compose(
+            [
+                _block(gens[5], gens[4], gens[3]),
+                _block(gens[2]),
+                _block(gens[1], gens[0]),
+            ]
+        )
+    )
+    assert abs(within_block.expectation_value(theta) - reference) > 1e-3
```
adamglos92: suggestion: no need for explicit checks tests are not flaky (@robertodr or @ludmilaasb can you confirm?)

adamglos92 (commented):
It seems that it would be cleaner if the fix would be done on the C++ side.

tests/test_circuit.py
antonnykanen: removed the negative controls

tests/test_circuit.py
antonnykanen: removed the negative controls

tests/test_circuit.py
antonnykanen: done

adamglos92 (approved):

antonnykanen:
Can I merge this or wait?

adamglos92:
I approved, but I would still prepare if one of  @Algorithmiq/software-engineers would review it

docs/content/docs/features/evaluation.mdx
```diff
@@ -28,11 +28,17 @@ sim.build_graph(a)
...
+the incremental build and the one-call build of the composition agree on
+[parameter_mapping][monoprop.monomial_propagator.MonomialPropagator.parameter_mapping] and on
+the expectation value and gradient at any parameter vector. In Heisenberg each extension
+therefore takes the low indices and lifts the ones already in the graph by
+`circuit.n_parameters`.
+
+Gate *arrival* order is the exception: it is not renumbered, so the two builds number their
+gates differently even though their parameter axes match. That only matters if you re-wire the
+graph by assigning a per-gate mapping — see the warning on
+[parameter_mapping][monoprop.monomial_propagator.MonomialPropagator.parameter_mapping].
```
ludmilaasb, line 41: This AI phrasing is very annoying. "It is this, not that". 

I suggest changing for more natural writing style.

src/monoprop/monomial_propagator.py
```diff
@@ -268,9 +268,12 @@ def build_graph(
...
         if rotate_axis:
-            # Cyclic in the index *values*, not a shift: the mapping must stay a bijection of
-            # 0..n-1 at every step, and a shift would leave 0 unused. Assigning the per-layer
-            # form (length graph_layers) keeps `_n_params` correct.
+            # A cyclic rotation, because both blocks move: the existing indices go up by
+            # `num_new`, and the block just appended wraps from the top of the axis down to
+            # 0..num_new-1. Shifting only the existing ones would collide with the new block and
+            # leave its numbering untouched, which is the bug. Assigning the per-layer form
+            # (length graph_layers) picks that reading unambiguously; the rotated mapping is
+            # still contiguous, so this only skips the property setter's redundant validation.
```
ludmilaasb, line 276: Please remove this and the other narrative comments from the bot.

antonnykanen:
Ok, tried to tweak docs and comments

ludmilaasb (approved):
LGTM

adamglos92 (approved):

